PR #7, opened 2015-09-02
Updated code to reflect changes in syntax

JaUg3:


src/utilities.jl
```diff
@@ -81,3 +81,21 @@ function format_names(names::Vector{UTF8String}, default::String, unknown::Strin
...
+  end
+  array = Array{Array{T,1},2}(ny,nx)
+
+  for (i,v) in enumerate(args)
+    row = floor(Int,(i-1)./nx)+1
+    col = mod(i-1,nx)+1
+    array[row,col] = v
+  end
+  array
+end
```
jcrist: Why is this needed? Why can't the arrays of vectors be created using:

```julia
julia> Vector[[1, 2, 3], [4, 5]]
2-element Array{Array{T,1},1}:
 [1,2,3]
 [4,5]  
```

This is only a small change from the previous version, but removes the need for a custom function.

jcrist: Ah, this doesn't work for 2d arrays of vectors. Perhaps a change in interface is needed then. Suggestions welcome.

jcrist:
Apologies on the delayed review - this somehow slipped through my email. Overall this looks good. I'm reluctant to add the `array` function, as it feels a little clunky - the interface for creating MIMO transfer functions should probably be changed somehow to avoid the need for such a function.

JaUg3:
I agree that the `array` function is clunky, but as of now it is only used in the tests to generate the 2d arrays, it is still up to the user to decide how to create arrays. 
I would argue that the interface in the toolbox is intuitive and that is the lack of a simple syntax to initiate 2d arrays in julia that is the problem. By keeping the current syntax we can hope that this will be reintroduced, but I have seen no indication that one is on the way.

As it is now, the user could for example use the `reshape` function
```
reshape(Vector[[1,2], [3,4], [5,6], [7,8]],2,2)
2x2 Array{Array{T,1},2}:
 [1,2]  [5,6]
 [3,4]  [7,8]
```
However, notice the order of the elements, which is transposed as compared to the original syntax, which is why I created the array function.
```
Vector[[1,2] [3,4]; [5,6] [7,8]]
2x2 Array{Array{T,1},2}:
 [1,2]  [3,4]
 [5,6]  [7,8]
```

jcrist:
As long as it's only for testing (not public facing), then I'm fine with merging it. Note that this also works though:

```julia
julia> [Vector[[1, 2, 3], [4, 5]] Vector[[2, 4, 6], [8, 10]]]
2x2 Array{Array{T,1},2}:
 [1,2,3]  [2,4,6]
 [4,5]    [8,10]
```

Not sure if that's more or less readable than the `array` function, but it solves the same problem and is idiomatic. I'll leave it up to you which to use.

If you do go with `array`, I'd prefer to not export it, and to change the name to something more informative (`vecarray'?).

JaUg3:
There seems to be a lot of discussion on how this syntax should be in julia and it will probably change again in 0.5. I therefore suggest that we keep the `array` function internal (named as `vecarray`) since it should survive most of the possible changes in the syntax.
It is then possible for the use its favorite syntax for whetever version of julia it's running.

I will push a new version with these changes shortly, and if you have no objections I will use the following code instead
```
function vecarray(T::Type, ny::Int,nx::Int, args::AbstractArray...)
    if ny*nx != length(args)
        error("Number of vectors must fit dimensions")
    end
    array = reshape(collect(a),nx,ny)
    permutedims(array,[2,1])
end
```

JaUg3:
I updated the code according to my previous comment and it should be ready to pull.

jcrist:
Looks good to me, thanks for your contribution!

PR #10, opened 2016-03-03
Major implementations and bug fixes

mfalt:
We have implemented several new function and fixed a lot of bugs and problems.
**I will merge this in a few days if there are no objections.**

We have removed most of the deprecated syntax, fixed the travis build, and started a transition to the Plots.jl package for plotting, and possibly Polinomials.jl for polynomials.
Among other things, this pull request implements:
`zpk` form for transfer functions,
`minreal`, `ss2tf`, `place` using acker, `margin`,
`gangoffour`, `gangofseven`,
`gangoffourplot`, `pzmap`, `nicholsplot`,
`pid`, `pidplots`, `rlocus`, `leadlink`, `laglink`, `leadlinkat`, `leadlinkcurve`.

Other notable changes is the removal of the ambiguous `gain` in favor of `dcgain`.


mfalt:
We should probably write tests for the new functionality before merging this.

REQUIRE
```diff
@@ -1,2 +1,3 @@
 PyPlot
-Requires
+Plots
+Polynomials
```
jcrist: Missing a newline at the end of the file.

src/analysis.jl
```diff
@@ -222,3 +241,209 @@ function fastrank(A::Matrix{Float64}, meps::Float64)
...
+function gangofseven(P::TransferFunction,C::TransferFunction,F::TransferFunction)
+    if P.nu + P.ny + C.nu + C.ny + F.nu + F.ny > 6
+        error("gof only supports SISO systems")
+    end
+    S, D, N, T = gangoffour(P,C)
+    RY = T*F
+    RU = N*F
+    RE = S*F
+    return S, D, N, T, RY, RU, RE
+end
```
jcrist: Missing a newline here as well. I suspect this is a problem with your text editor.

src/discrete.jl
```diff
@@ -23,20 +26,252 @@ function c2d(sys::StateSpace, Ts::Real, method::Symbol=:zoh)
...
+e.g integral part [1, -1]^k
+AS     : Pre-specified factor of S,
+e.g notch filter [1, 0, w^2]
+
+Outputs: R,S,T  : Polynomials in controller
+
+See function DAB how the solution to the Diophantine-
+Aryabhatta-Bezout identity is chosen.
+
+Michael Lundh     LastEditDate : Wed Mar 21 14:35:36 1990
```
jcrist: What is this "LastEditDate" from? Was this just copied out of some Matlab/octave code?

src/freqresp.jl
```diff
@@ -14,6 +14,9 @@ function freqresp{S<:Real}(sys::LTISystem, w::AbstractVector{S})
         s = im*w
     end
     sys = _preprocess_for_freqresp(sys)
+    if isa(sys,TransferFunction) && !isa(sys.matrix, Array{SisoRational})
```
jcrist: Using `isa` is an antipattern in Julia, especially if the code is type stable. Is there a way to perform this check at compile time by defining a small method overloaded for the different type configurations? Something like this (untested, and needs a better name):

```
_normalize_system(sys) = sys
_normalize_system(sys::TransferFunction) = tf(sys)
_normalize_system(sys::TransferFunction{SisoRational}) = sys
```

src/plotting.jl
```diff
@@ -148,97 +172,215 @@ function bodeplot(systems::Vector{LTISystem}, w::AbstractVector)
...
+    for k=Gains
+        PVals=Ni_Ph(k,0:0.1:360)
+        GVals=Ni_Ga(k,0:0.1:360)
+        for l in PCyc
+            Plots.plot!(fig,PVals+l*360,GVals,c=CalcRgb(k),linewidth=LWidth, grid=false)
+            if true
+                if sign(k) > 0
+                    mla = 210
+                else
+                    mla = 210
```
jcrist: Both cases result in the same `mla`

jcrist:
This is a huge PR, and I only gave it a brief review. Looks like a lot of nice new features, but I didn't verify their correctness (I trust you on that :)). A few things I noticed, with various levels of importance:

- All the new functionality definitely needs tests. I no longer have time to work on this, so I can't *mandate* that this practice is upheld, as I'm not the one doing the work :), but I strongly suggest you get good test coverage.
- There are quite a number of lines that end in semicolons. While semicolons are allowed, they're completely unnecessary, and bit of a stylistic issue.
- There are many huge chunks of commented out code. Again, this is mostly a code cleanliness issue. I personally prefer not to commit commented out functions, and instead keep those implementations in version control in a branch, but this is just an opinion.

One thing I did notice was that there were a few functions that appeared to be ported directly from Matlab/octave implementations. Depending on the license of those implementations this *may* be ok, but it also may not be. In general, implementing algorithms from papers is good, implementing algorithms from compatibly licensed implementations is also ok (although both should have a good citation).

mfalt:
Thank you for your comments, I agree with you on all points. As were are a few people who have been working on this I have not verified all the parts myself, but I intend to, especially regarding the origin of some if the code. 

We have been working alongside a few of our courses, trying to include the functionality needed to solve the exercises into this package. I have recently noticed that people started working of my development branch instead of this one, so I felt that it was quite urgent to show that we intended to take those changes and improvements here, making it possible for people to comment and discuss them.

I am also aware that some of the parts are far from perfect, and your comments here are very good and should be fixed before we pull this. However, I also feel that there is an interest in getting several of these changes pulled so that people can start using them. While I have started writing tests and intend to include them in this PR (as I commented above), I am not convinced that it is best to get full coverage before we are able to pull this.

The current tests are after all ensuring that we are not breaking existing functionality. The only exception being the "gain" function for which it would be great to have your input; what was the intended behavior of this function?

jcrist:
> I am not convinced that it is best to get full coverage before we are able to pull this.

Totally up to you on this, you're the one doing the work here :).

> The only exception being the "gain" function for which it would be great to have your input; what was the intended behavior of this function?

If I remember correctly, it's the gain portion of the zero-pole-gain form. I think the more complicated state space implementation was to deal with a non-minimal realization of the system, but I'm not sure. It's been too long since I've done any controls stuff :).

src/discrete.jl
baggepinnen: Your observation is correct. The copyright holder mentioned on the line below is our department. The last edited date should be updated if kept, though.

jcrist: Is the cited software available online somewhere? Is it compatibly licensed? If not, is there a relevant paper with the algorithm/code in it you could cite? In general, if you can't point at a source that says "this code is ok for me to copy", then it's not ok for you to copy.

baggepinnen: Since we are the copyright holders, we should be able to put any licence we desire on it. The algorithm is written on our department, for use in our courses and is made available from our website. I can add a citation to the book it is implemented from, also written at our department.

jcrist: Citing the book, and removing the "last modified" and copyright sounds like a good solution.

mfalt:
`sigmaplot` and `marginplot` needs to be converted to Plots.jl before we pull (because of changes in `plotScale`), we otherwise only seem to be lacking tests by now.

mfalt:
I think we have addressed the most important issues now, so I plan to pull this request tomorrow. Once that is done I will register it in the official Julia repository. We should still write test for the new functionality (we have tested all the code in one way or another, but we need structured tests). We should also look into making sure we have type stability and a few other things, I will create issues for this and move the issues from my own repository to this.

PR #9, opened 2015-12-12
AbstractFloat and AbstractString depwarn fixes

kersulis:
Just taking care of a few deprecation warnings. I also removed a few leading spaces.

PR #24, opened 2016-03-13
Fixed complaints from Lint and removed some type instability

mfalt:
This commit takes care of most of the complaints from Lint, some of the changes fixes minor bugs, others merely reduce confusion (such as avoiding variable names that are already defined as functions).

It also takes care of some type instability (and I am still not sure why!). I updated the example to reflect this. The simplest example is: 
Pre-commit:
```
julia> typeof([tf([1],[1]) for i = 1:2])
Array{Any,1}
```
Post-commit:
```
julia> typeof([tf([1],[1]) for i = 1:2])
Array{Control.TransferFunction,1}
```

This commit also incorporates the changes in PR #9 with the exception of the changes to polys.jl since we are planning on transitioning to Polynomials.jl.

Could you take a look at the changes in this commit @baggepinnen? We should also consider if we want to keep the `__precompile()__` that I accidentally added.

We should still write tests for all bugs that this commit fixes.


src/types/transferfunction.jl
```diff
@@ -187,25 +187,25 @@ function zpk(tf::TransferFunction)
...
-    tf.matrix = matrix
-    return tf
+    s.matrix = matrix
+    return s
 end
 
-tf(num::Vector, den::Vector, args...; kwargs...) =
-    tf(reshape(Vector[num], 1, 1), reshape(Vector[den], 1, 1), args...; kwargs...)
+tf(num::Vector, den::Vector, Ts::Real=0; kwargs...) =
+    tf(reshape(Vector[num], 1, 1), reshape(Vector[den], 1, 1), Ts; kwargs...)
```
mfalt: Could anyone explain why this removes type instability of a call like `tf([1],[1])` (or why we had type instability in the first place)?

PR #26, opened 2016-03-16
Update README.md

jleny:
The type inference for the array of transfer functions with a comprehension does not seem to be reliable. There is a note about this in the Julia documention http://docs.julialang.org/en/release-0.4/manual/arrays/
For me, without controlling the type explicitly, CLs is inferred as Array{Any,1}, which stepplot does not accept.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Note that this package requires Julia 0.4.
 
 ## Documentation
 
-All exported functions have docstrings, which can be viewed from the REPL. A
+Some of the exported functions have docstrings, which can be viewed from the REPL. A
 documentation website is on the list of things to do...
 
 ## Usage
@@ -59,7 +59,7 @@ TransferFunction:
 Continuous-time transfer function model
 
 # Create an array of closed loop systems for different values of Kp
-julia> CLs = [kp*P/(1 + kp*P) for kp = [1, 5, 15]];
+julia> CLs = TransferFunction[kp*P/(1 + kp*P) for kp = [1, 5, 15]];
 
 # Plot the step response of the controllers
 julia> stepplot(CLs);


mfalt:
This was a problem with the constructor
`tf(num::Vector, den::Vector, args...; kwargs...)`
for which the compiler couldn't infer the type and therefore set the array to `Any`, I fixed this in the same PR that I updated the README (#24) but I assume that you installed via Pkg.add("ControlSystems") and therefore got the last tagged version (which doesn't have this fix). Could you tests Pkg.clone("ControlSystems") to see if the problem is fixed? If so, we could add a new tag and push it to METADATA.

mfalt:
I will pull this for now to avoid confusion until we have a new tag,

jleny:
I'm a bit confused: I was and I'm still currently on master on my machine (after Pkg.checkout("ControlSystems"), now latest commit 8774794) and if I write CLs = [kp*P/(1 + kp*P) for kp = [1, 5, 15]] with a comprehension as the example previously said, the type is still inferred as Array{Any,1}. Are you saying I should get Array{ControlSystems.TransferFunction,1}? To get that, I still need an explicit TransferFunction in front of the array, or to declare P as const.

mfalt:
You are right, I didn't test the example properly. Since P is a global variable, the compiler doesn't know that it will remain a TransferFunction and thus has to assume it to be Any. The commit I referred to was only fixing that
```julia
julia> typeof([tf([1],[1,kp]) for kp = [1,5,15]])
Array{ControlSystems.TransferFunction,1}
```
instead of
```julia
julia> typeof([tf([1],[1,kp]) for kp = [1,5,15]])
Array{Any,1}
```
this works since there is no global variables included, but was previously a problem even within functions. The problem we have in the example only exists in the REPL. The following code works without a problem for example
```julia
julia> f() = begin
       s = tf("s")
       P = 1/(s+1)
       Cls = [P*kp/(1+P*kp) for kp = [1,5,15]]
       end;

julia> typeof(f())
Array{ControlSystems.TransferFunction,1}
```
since the type of P can be known. I don't think we have any other choice than leaving the example as it is until the inference in the REPL is improved by the julia team.

mfalt:
It seems like we could also do
`CLs = map( (kp) -> kp*P/(1 + kp*P), [1, 5, 15])`
which works (and I think the reason is because the anonymous function is compiled for this specific P).

jleny:
I think the explicit LTISystem as you had before or TransferFunction as it is now in front of the array is fine for now. I feel that introducing map in this introductory example would make it more confusing for people with little programming experience. And indeed const is not prettier.

PR #33, opened 2016-03-23
Implement feedback functions

baggepinnen:


mfalt:
We should write tests for this and other PRs preferably, this would for example give an incorrect answer for `Transferfunction{SisoZpk}`. We should probably deprecate and switch `tfden` to something like `denpoly` which returns a polynomial and is implemented per Siso Type, and then using polynomial multiplication.

baggepinnen:
I did not consider any other type than regular TFs, should specify the types of the input arguments.

An issue with denpoly could be that no one would find that function, they would look for something like `den` and `num`

mfalt:
Another error in the code is that `B` is not defined in the first function.
We could print an error referring the user to `denpoly` when `den` and `num` are called on a transferfunction.
I see no reason why this function should be restricted to a specific Siso type since its equaly easy to implement for any type.

PR #34, opened 2016-03-23
Generalized transferfunction and vector creation

mfalt:
This PR adds the following:
* A new SisoType: SisoGeneralized which allows for a much wider range of transferfunctions but with limited capabilities. It is for example possible to have systems with delays and systems like "exp(-sqrt(s))/(s+1)". It is also possible to convert to and from this type (only from when it consists of simple operations like +,-,/,*). The systems can be simulated in frequency and time domain, although the time domain should be more robust, but they do not allow for calculation of poles and zeros for example.
* Fix for vector creation: There has been a lot of problems with creating vectors like [tf1, tf2] when they are of different types, this is fixed now.
* SisoType as part of the transferfunction, julia is now able to dispatch based on which siso type it consists of. The type is now `TransferFunction{T<:SisoTf}` instead of just `TransferFunction`.
* Proper `convert` functions between all types that had a `promote_rule`.

I will add some more tests showing the conversion to and from the SisoGeneralized type before we should pull this.

This is a rebase of the messy abstracttf branch.


src/freqresp.jl
```diff
@@ -66,6 +72,16 @@ function evalfr(sys::TransferFunction, s::Number)
     return res
 end
 
+#This is called after _preprocess_for_freqresp
+function evalfr(mat::Matrix, s::Number)
+    S = promote_type(typeof(s), Float64)
+    res = Array(S, size(mat)...)
+    for i in eachindex(mat)
+        res[i] = evalfr(mat[i], s)
```
baggepinnen: Why not `map(evalfr,mat)` ?

src/timeresp.jl
```diff
@@ -60,10 +63,15 @@ function impulse(sys::StateSpace, t::AbstractVector)
     end
     return y, t, x
 end
-impulse(sys::StateSpace, Tf::Real) = impulse(sys, _default_time_vector(sys, Tf))
-impulse(sys::StateSpace) = impulse(sys, _default_time_vector(sys))
-impulse(sys::LTISystem, args...) = impulse(ss(sys), args...)
+function impulse(sys::TransferFunction{SisoGeneralized}, t::AbstractVector)
+    u = zeros(length(t), sys.nu)
+    u[1,:] = 1/(t[2]-t[1])
+    lsim(sys::TransferFunction{SisoGeneralized}, u, t)
```
baggepinnen: lsim(sys, u, t)

src/types/sisogeneralized.jl
```diff
@@ -0,0 +1,108 @@
...
+            expr = expr.args[3]
+        end
+        new(expr)
+    end
+end
+
+SisoGeneralized(str::AbstractString) = SisoGeneralized(parse(str))
+
+function minreal(sys::SisoGeneralized, eps::Real=sqrt(eps()))
+    error("minreal is not implemented for abstract transferfunctions")
```
baggepinnen: Replace error with `warn` in case minreal is called inside function

src/types/transferfunction.jl
```diff
@@ -180,29 +201,44 @@ function zpk{T<:Vector,S<:Vector}(z::VecOrMat{T}, p::VecOrMat{S}, k::VecOrMat, T
...
 function zpk(tf::TransferFunction)
-    tf = copy(tf)
+    oldmat = tf.matrix
     matrix = Array(SisoZpk, tf.ny, tf.nu)
-    for o=1:tf.ny
-        for i=1:tf.nu
-            matrix[o, i] = convert(SisoZpk, tf.matrix[o, i])
-        end
+    for i in eachindex(oldmat)
+        matrix[i] = convert(SisoZpk, oldmat[i])
```
baggepinnen: `matrix[i] = convert(SisoZpk, oldmat[i])`
`matrix = map(m->convert(SisoZpk, m), oldmat)`

PR #36, opened 2016-03-24
Compute the Hinfinity norm for continuous-time systems

jleny:
Implemented the same algorithm as in Matlab/SLICOT directly in Julia. Only for continuous-time systems for now.

Note that with this change norm(sys,2) returns a number, but norm(sys,Inf) returns a pair (ninf,fpeak). This might not be desirable.

There are also some minor changes, to isstable, and evalfr (map_to_unit_circle was not used).


mfalt:
This is a great PR, thanks! However, I am not convinced that we want to define the stability to include marginally stable systems (maybe this could be an option). In that case we need to consider that the system can have multiple poles on the same location on the imaginary axis, in which case it is unstable.

jleny:
To clarify, in the **current** version systems we have
```
if any(real(pole(sys)).>0)
    return false
```
what I'm proposing is what you say: marginally stable systems are **not** stable. This would be compatible with the way the discrete-time case is handled just below. Still in the current version
```
if any(abs(pole(sys)).>=1)
            return false
```

mfalt:
I'm sorry, I misinterpreted the code, I agree that your way is better. I will merge this tomorrow,

PR #42, opened 2016-03-25
Fixed isapprox kwargs

mfalt:
Will this fix the problem? The only difference as I can see is that I do not default atol to 0 for SisoTf (which seems weird).

### DIFF
--- a/src/types/polys.jl
+++ b/src/types/polys.jl
@@ -178,11 +178,11 @@ function ==(p1::Poly, p2::Poly)
     end
 end
 
-function isapprox(p1::Poly, p2::Poly)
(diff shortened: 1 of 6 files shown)


baggepinnen:
If the tests pass I'm happy. I have problems viewing this on my phone, but it seems the only thing different that I did not try was modifying isapprox for Poly. Was that perhaps the reason the more general isapprox was called instead of isapprox(poly), resulting in the stackoverflow?

mfalt:
I was never able to get stack overflow, but the isapprox for poly failed since it didnt accept kwargs. I am pushing an update for the inames/inputnames too.

PR #52, opened 2016-04-25
Critical direct term added to ss2tf and better default tol in sisotf

mfalt:
Defaulting atol to same as in Base, fixed ss2tf, added tests for the bug.


mfalt:
@baggepinnen Could you please check that you agree with the atol change and then merge.

baggepinnen:
Looks good. You can go ahead and delete the branch if it's no longer needed.

Issue #17, opened 2016-03-11
bode for SISO returns all data in last dim


mfalt:
_From @baggepinnen on February 22, 2016 19:25_

This behavior causes lots of squeeze statements
```julia
julia> Control.bode(S)
(
1x1x200 Array{Float64,3}:
[:, :, 1] =
 1.00504

[:, :, 2] =
 1.04091

[:, :, 3] =
 1.07809

...
```

_Copied from original issue: mfalt/Control.jl#22_

mfalt:
_From @baggepinnen on February 22, 2016 19:26_

The same goes for `freqresp` and probably more funs

mfalt:
This should be a priority (and will be breaking some existing code), but should be included in one of the next versions.

baggepinnen:
What if bode returned an nu x ny matrix with vectors? This would align well
with a TF being an nu x ny matrix of siso systems.
In the siso case, the vector could be automatically popped out of the 1 by
1 matrix

mfalt:
But the problem then is that we have to have a special case for siso systems, I think the solution above is more elegant and user friendly.

mfalt:
Fixed in a2c7bbf

PR #53, opened 2016-04-25
Frequency no longer in last dim

### LINKED ISSUE #39: propagate args and kwargs to Plots.plot commands

mfalt:
Fixing the problem with frequencies begin added in the last dim #17 instead of first in frequency responses. Also added several good tests for time and frequency domain. We should not pull this until we have used it for a while since it contains many changes, but feel free to look through my changes @baggepinnen.

Should also fix #39.


mfalt:
Included in #55

PR #54, opened 2016-04-26
Lqg

baggepinnen:
Work in progress, not to be pulled until more tests are implemented. Please comment on the stuff available so far


src/plotting.jl
```diff
@@ -36,6 +37,8 @@ function setPlotScale(str::AbstractString)
     _PlotScale, _PlotScaleFunc, _PlotScaleStr = plotSettings
 end
 
+setPlotScale("log10")
```
mfalt: We should consider whether or not we should keep this.

src/types/lqg.jl
```diff
@@ -0,0 +1,216 @@
...
+Initializes an LQG object. This object can be used to contruct an LQG controller.
+
+# Constructors
+
+`LQG(P,Q1,Q2,R1,R2, qQ, qR, sysc, L, K, integrator)` This is mainly used inside the toolbox
+
+`LQG(P,Q1,Q2,R1,R2; qQ=0, qR=0, integrator=false)` Supply all weighting and covariance matrices as well as optional loop-transfer recovery gains. The boolean `integrator` indicates whether or not the system will be augmented with constant input disturbances to introduce integral action.
+
+`LQG(P, qQ, qR; integrator=false)`  if called like this, where `qQ,qR` are scalars, the weighting matrices will be
+`Q1 = qQ^2 C'C,  Q2 = I,  R1 = qR^2 B B',  R2 = I`
```
mfalt: This could look nicer

baggepinnen: What part, the import statement or the documentation?

mfalt: I added the comment on the last line of what is shown above. I meant the equations.

test/test_simplification.jl
```diff
@@ -11,4 +11,19 @@ G = ss([-5 0 0 0; 0 -1 -2.5 0; 0 4 0 0; 0 0 0 -6], [2 0; 0 1; 0 0; 0 2],
...
+P = [1/(s+1) 2/(s+3); 1/(s+1) 1/(s+1)]
+sys = ss(P)
+sysmin = minreal(sys)
+
+@test size(sysmin.A,1) == 3 # Test that the reduction of sys worked
+
+t = 0:0.1:10
+y1,x1 = step(sys,t)[[1,3]]
+y2,x2 = step(sysmin,t)[[1,3]]
+@test_approx_zero sum(y1.-y2) # Test that the output from the two systems are the same
```
mfalt: use norm instead?

baggepinnen: Done, good point

test/test_simplification.jl
```diff
@@ -11,4 +11,19 @@ G = ss([-5 0 0 0; 0 -1 -2.5 0; 0 4 0 0; 0 0 0 -6], [2 0; 0 1; 0 0; 0 2],
...
 
+
+## MINREAL ##
+
+s = tf("s")
+P = [1/(s+1) 2/(s+3); 1/(s+1) 1/(s+1)]
+sys = ss(P)
+sysmin = minreal(sys)
+
+@test size(sysmin.A,1) == 3 # Test that the reduction of sys worked
```
mfalt: It would be nice with an example for which we know what the answer will be and compare to that instead. Testing the steps as you do work, but is not perfect?
A test where the system can almost be reduced would be nice to verify that we do not accidentally remove any dynamics would be nice.

mfalt:
It seems like minreal can't handle systems that are not diagonalizeable, for example
```
sys1 = ss(1/(s+1)^2)
minreal(sys1)
```
will crash, so even if the system contains dynamics that can be removed, it will crash if any part is not diagonalizable.
It will also crash if the system contains any unstable dynamics (or integrators)
`minreal(ss(1/s))`

baggepinnen:
Yeah I noticed those issues as well. The issue for non-diagonalizable systems is trivial, just remove the diagonalization. The `balreal`function relies on `gram` which requires stable as. systems.

src/types/statespace.jl
```diff
@@ -233,7 +233,7 @@ end
 *(n::Real, s::StateSpace) = *(s, n)
 
 ## DIVISION ##
-/(s1::StateSpace, s2::StateSpace) = s1*(1/s2)
+/(s1::StateSpace, s2::StateSpace) = s1*inv(s2)
```
mfalt: I think we have to deprecate this function for MIMO systems and then eventually remove it for MIMO. The user should be referred to calling either `s1*inv(s2)` or `inv(s2)*s1` for MIMO instead.

PR #55, opened 2016-04-26
Plottests

mfalt:
Adds tests for plots and includes the changes in bodedims, we should not pull this quite yet.

This significantly slows down the building time and the comparisons are not very exact, could maybe be improved if some other back-end is chosen instead of Gadfly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ notifications:
 script:
     - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
     - julia -e 'Pkg.clone(pwd()); Pkg.build("ControlSystems")'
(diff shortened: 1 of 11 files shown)


baggepinnen:
See comments on speed in the backends section on plots.readthedocs

mfalt:
Interesting, so it seems we might gain a lot by switching to PyPlot. The problem is however that it needs to be installed which takes quite a while. PyPlot is currently only installed when building docs, that is, when testing on master.

mfalt:
After some ugly tricks to get travis to build it seems to work now, and with pretty much pixel perfection! Feel free to run
Pkg.test("ControlSystems")
on this branch @baggepinnen to verify that it works there too.

Issue #15, opened 2016-03-11
gain is not working


mfalt:
_From @baggepinnen on February 23, 2016 20:2_

Should be renamed to static gain, or dcgain as in matlab.
Should also be *\* working properly**

_Copied from original issue: mfalt/Control.jl#24_

mfalt:
Fixed for tf, ss remaining.

mfalt:
We can still not handle if the tf has both a pole and a zero in 0, this returns `NaN`.
The implementation for ss works better now, but can not handle when A is not invertible (for example with uncontrollable states). I think this could be solved using the psuedo-inverse. However, we still return a finite gain for unstable systems.

The previous implementation (where `nz=tzero(sys)`)
`sys.C*(sys.A^(nx - nz - 1))*sys.B`
still seems completely wrong.

aytekinar:
Assign the issue to me --- I am gonna fix this.

mfalt:
Great! @baggepinnen is currently implementing minreal for statespace which might be of use, but feel free to use some other way if it is more robust/efficient.

aytekinar:
I have already forked the repository, and I will create a branch `15-dcgain-fix`. In the functions I will not assume pole-zero cancellations (as anyways it is not encouraged to do cancellations in the right half plane or outside the unit disk), so I will pay attention to the poles directly. And I will calculate `G(s=0)` and `G(z=1)` in the limit. As a result, the user needs to call the `minreal` or some other function to explicitly demand pole-zero cancellations. The output of `minreal` will be of transfer function and/or state-space form, and hence, the `dcgain` function should work fine.

We can talk more about it when I create the pull request.

Issue #20, opened 2016-03-11
Functions to access to numerator and denomerator of tf


mfalt:
_From @baggepinnen on February 21, 2016 12:28_

e.g.  `numerator = num(G)`
and `num,denom = poly(G)`

_Copied from original issue: mfalt/Control.jl#12_

mfalt:
What would be the result of this, a Transferfunction, or a Polinomial? And how do we handle the fact that transferfunctions by default are MIMO systems, only allow it for siso systems?

mfalt:
_From @baggepinnen on February 21, 2016 13:17_

I often find the fact that all TFs are MIMO systems confusing. This is not the case in matlab, and it causes a lot of unexpected behaviour. The boxing and unboxing required to convert between scalars and arrays is very confusing in the beginning, and annoying in the end.  Students with little or no julia experience will not find the error messages very helpful in the beginning.

The result could be an array of polynomials in the MIMO case. In the siso case, I would like a polynomial, not a one element array of polynomials, which I acknowledge is a strange special case behavior.

mfalt:
My proposal is that we do the following three things:
* Deprecate "`Base.num`" and "`Base.den`" for SISO types
* Implement `numvec`/`denvec` which returns vectors of the polynomial coefficients as represented in sisotf (and throw error for sisogeneralized).
* Implement `numpoly`/`denpoly` that returns a Polynomials.Poly in the same way as above.

I will implement this soon if no one disagrees.

aytekinar:
@mfalt what is the situation for this one? do you need help with `numvec`/`denvec` and `numpoly`/`denpoly` ? I will be needing them for `dcgain` when adressing your comments in #56

mfalt:
I was hoping for some input first. But I can try to do this during the weekend/next week.

mfalt:
@aytekinar It is implemented now in #57, I will pull it into master when I have verified it.

aytekinar:
@mfalt I have seen it and have checked myself and given some minor comments. I can rebase my dcgain branch on numden and push my changes accordingly, or else, I can wait until it is on master. anyways dcgain is tagged for v0.2 milestone, which is due by the end of the month.

mfalt:
Great, I'll look at them, I think you could either rebase and use the numvec functions, or if you want to use the polyomials, wait until I change the inplementation to Polynomials.jl, which I think will be done this week.

aytekinar:
[...]

PR #56, opened 2016-04-27
Fix `dcgain(sys::TransferFunction)` (addresses #15)

aytekinar:
Fixed the function definition of `dcgain` for `sys::TransferFunction` type
inputs. Added documentation accordingly. Seems to be working fine with the
provided example set.

### DIFF
--- a/docs/src/examples/example.md
+++ b/docs/src/examples/example.md
@@ -1,7 +1,8 @@
-    {meta}
-    DocTestSetup = quote
-        using ControlSystems
-    end
+```@meta
+DocTestSetup = quote
+    using ControlSystems
+end
+```
 
 # PID design functions
 By plotting the gang of four under unit feedback for the process
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -1,26 +1,31 @@
 # ControlSystems.jl Manual
 
-    {meta}
-    CurrentModule = ControlSystems
+```@meta
+CurrentModule = ControlSystems
+```
 
 ## Examples
-    {contents}
-    Pages = ["examples/example.md"]
-    Depth = 1
+```@contents
+Pages = ["examples/example.md"]
+Depth = 1
+```
 
 ## Guide
 
-    {contents}
-    Pages = ["man/introduction.md", "man/creatingtfs.md"]
-    Depth = 1
+```@contents
+Pages = ["man/introduction.md", "man/creatingtfs.md"]
+Depth = 1
+```
 
 ## Functions
 
-    {contents}
-    Pages = ["lib/constructors.md", "lib/plotting.md"]
+```@contents
+Pages = ["lib/constructors.md", "lib/plotting.md"]
+```
 
 ## Documentation Index
 
-    {index}
-    Pages = ["lib/constructors.md", "lib/plotting.md", "lib/syntheis.md", "lib/timefreqresponse.md", "lib/analysis.md"]
-    Depth = 1
+```@index
+Pages = ["lib/constructors.md", "lib/plotting.md", "lib/syntheis.md", "lib/timefreqresponse.md", "lib/analysis.md"]
+Depth = 1
+```
--- a/docs/src/lib/analysis.md
+++ b/docs/src/lib/analysis.md
@@ -1,24 +1,26 @@
-    {index}
-    Pages = ["analysis.md"]
+```@index
+Pages = ["analysis.md"]
+```
 
 # Analysis
 
-    {docs}
-    covar
-    ctrb
-    damp
-    dampreport
-    dcgain
-    delaymargin
-    gangoffour
-    gangofseven
-    gram
-    margin
-    markovparam
-    norm
-    obsv
-    pole
-    sigma
-    stabregionPID
-    tzero
-    zpkdata
+```@docs
+covar
+ctrb
+damp
+dampreport
+dcgain
+delaymargin
+gangoffour
+gangofseven
+gram
+margin
+markovparam
+norm
+obsv
+pole
+sigma
+stabregionPID
+tzero
+zpkdata
+```
--- a/docs/src/lib/constructors.md
+++ b/docs/src/lib/constructors.md
@@ -1,19 +1,21 @@
-    {index}
-    Pages = ["constructors.md"]
+```@index
+Pages = ["constructors.md"]
+```
 
 # Constructing transfer functions
 
-    {docs}
-    append
-    c2d
-    feedback
-    feedback2dof
-    minreal
-    parallel
-    series
-    sminreal
-    ss
-    ss2tf
-    tf
-    tfg
-    zpk
+```@docs
+append
+c2d
+feedback
+feedback2dof
+minreal
+parallel
+series
+sminreal
+ss
+ss2tf
+tf
+tfg
+zpk
+```
--- a/docs/src/lib/plotting.md
+++ b/docs/src/lib/plotting.md
@@ -1,21 +1,23 @@
-    {index}
-    Pages = ["plotting.md"]
+```@index
+Pages = ["plotting.md"]
+```
 
 # Plotting functions
 
-    {docs}
-    bodeplot
-    gangoffourplot
-    impulseplot
-    leadlinkcurve
-    lsimplot
-    marginplot
-    nicholsplot
-    nyquistplot
-    pidplots
-    pzmap
-    pzmap!
-    rlocus
-    sigmaplot
-    setPlotScale
-    stepplot
+```@docs
+bodeplot
+gangoffourplot
+impulseplot
+leadlinkcurve
+lsimplot
+marginplot
+nicholsplot
+nyquistplot
+pidplots
+pzmap
+pzmap!
+rlocus
+sigmaplot
+setPlotScale
+stepplot
+```
--- a/docs/src/lib/synthesis.md
+++ b/docs/src/lib/synthesis.md
@@ -1,24 +1,26 @@
-    {index}
-    Pages = ["synthesis.md"]
+```@index
+Pages = ["synthesis.md"]
+```
 
 # Synthesis
 
-    {docs}
-    balance
-    care
-    dab
-    dare
-    dkalman
-    dlqr
-    dlyap
-    kalman
-    laglink
-    leadlink
-    leadlinkat
-    loopshapingPI
-    lqr
-    pid
-    place
-    reduce_sys
-    rstc
-    rstd
+```@docs
+balance
+care
+dab
+dare
+dkalman
+dlqr
+dlyap
+kalman
+laglink
+leadlink
+leadlinkat
+loopshapingPI
+lqr
+pid
+place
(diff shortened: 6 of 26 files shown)


aytekinar:
You can review the changes. I thought we should be calculating the DC gain of a MIMO system the way MATLAB [does][1]. The code may not be the most efficient code ever, but at least it solves most of our problem by properly applying L'Hopital when needed.

@neveritt and I have also thought that we can just use this method, i.e., the one with `TransferFunction` as its input, and use some conversion properties of the Julia language to define the same function for other types such as `StateSpace` and/or `Zpk` models. Then we will have one function to keep track of, basically.

Another modification I have made is the return value of `NaN` instead of throwing exceptions, as normally exception handling is far less efficient than checking against `NaN`. Julia has the built-in `isnan` function, which might prove useful when checking if the system is stable or not, i.e., `isnan(dcgain(tf(0)))`.

[1]: https://se.mathworks.com/help/control/ref/dcgain.html

baggepinnen:
Replacing errors with `NaN` return values is probably something we should discuss. It might be faster, but is the reduced computational time worth it when you consider that errors have error messages attached to them, whereas `nan` say just about nothing else than that an error has occurred. I believe most functions in the package, and in other julia packages, use errors, I can not remember a single instance of receiving a `nan` representing an error.

mfalt:
I have only done a quick check, but these are my comments so far (and I'm still updating this comment):

* I think its reasonable to return NaN on unstable systems, but I will have to think about this a bit more.

* Instead of iterating over m, and n, its more readable to use
```
for i in eachindex(sys.matrix)
    ...
    something(sys.matrix[i])
   ...
end
```
this will also ensure that the indices are iterated in the correct order.
* Just as I mentioned in the chat, we should instead of doing
`zerovec = sys.matrix[n,m].num.a[end:-1:1]`
be using a function `numvec` that is defined for each siso-type. The current code would fail for `sisozpk`. The `numvec` function is currently implemented in the types using `Base.num`. I think this function should be deprecated and renamed `numvec`, and then add another function `numpoly`.
* I think that we in general are trying to keep the functions type stable, so that we return the same type for all transferfunctions. This would mean an array in all cases, I think this is more julianic. We have a branch where we are trying to return scalars and vectors for sisosystems instead of vectors/arrays, but I am probably not in favor of using that style.

aytekinar:
@baggepinnen, I have already written the documentation saying that `NaN` indicates an unstable system. Throwing an error in MIMO systems, which are unintuitive in transfer function models in your view, I know, results in an error message. However, returning an array of values consisting of real numbers, `Inf`s and `NaN`s would give the user freedom to see which input-output pairs are stable and which are not.

@mfalt I will modify the code accordingly, to take `eachindex` into account. In addition, you are right in demanding a type consistency in return values. This will further optimize the code, IMO. Regarding the `numvec` and `numpoly` (and their `den`-versions), I thought we could address them altogether in a separate issue (such as `Polynomials` switch or a brand new one such as `type improvements`) to keep changes in their corresponding issues/topics. Should you want me to use these accessor functions, then I need to check what the available options are.

baggepinnen:
I agree, the use of `nan` in this case seems well motivated.

mfalt:
I agree that you do not need to implement the new numvec/denvec in this PR. But then you should probably use `num(sisotfsys)` and `den(sisotfsys)`in the `_sisodcgain` for `SisoTf` and `zp2polys(num(sisozpksys))` and `zp2polys(den(sisozpksys))` for SisoZpk. This last function (that I probably implemented) has a very stupid name since it returns a vector, not a poly.
I think this comment clearly illustrates why we need to take care of the `numvec`/`numpoly` issues that are tracked in #20.

aytekinar:
Then, I will proceed with some modifications to my fix taking into account the above comments. Since @baggepinnen also agrees that `NaN` has a good motivation, I will rewrite the code and use cleaner code.

By the way, @neveritt and I do still insist that we should write one function, say `_dcgainsisotf(sys::SisoTf)`, and call it in the outer `for idx in eachindex(...)` loop. We also need to define conversion rules from different models to `SisoTf`, which would also fix the problem of having different and maybe confusing `zp2polys(num(sisozpksys))`-type of functions/methods.

What do you think?

mfalt:
Yes, having only one function for calculating siso dcgain is all right, we have done similar things before. 
To convert to `SisoTf` either call `tf(sys)` and all parts will be converted or `convert(SisoTf , sysosys)` on the individual siso systems. (I think the first option is best, that way you can define the following functions:
```
function dcgain(sys::Transferfunction{SisoTf})
    ...
end

dcgain(sys::Transferfunction) = dcgain(tf(sys))
```

aytekinar:
One of Julia's strong features is the type conversion. That's why we thought that having a `dcgain(sys::LTISystem) = dcgain(convert(::TransferFunction, sys))` conversion function, inside which we check the subtype of `sys` and do the corresponding conversion.

`tf(sys::StateSpace)` is still needed to expliticly demand `tf` models, but at least we can have a more robust `dcgain` function, or?

Anyways, I think we get what you are trying to say. And most likely we are thinking the same way as each other.

mfalt:
I am not sure I understand exactly what you mean but I think we have roughly the same idea.
Yes, state space also has to be handled. Either by conversion or a more efficient way of calculating the dcgain tor these types.

aytekinar:
@mfalt and @neveritt ... seriously we need to talk about the type hierarchy.

I was trying to update `dcgain`. Assume we have a helper function `_dcgain(sys::SisoTf, discrete::Bool = false)`. This guy should be capable of calculating the DC gain of SISO systems. However, inside the function, you have asked me to use `isstable` for example. `isstable` is defined for `sys::LTISystem`. Since `SisoTf` is not a subtype of `LTISystem`, I cannot use that function. Even worse is, when I try to modify the definition of the abstact type as `abstract SisoTf <: LTISystem`. This time, I have an error stating that `SisoRational` types do not have `Ts` fields.

We face similar problems with the available type hierarchy.

mfalt:
To be honest, I have never had a problem with the SiSo/MiMo differences, the idea is just that a TransferFunction, which is mimo in general contains a bunch of rational functions, maybe if you think of them as RationalFunction, instead of SisoRational it would make more sense. The reasoning is we only care about representing the function there, how they are interpreted is stored in TransferFunction. They were never intended to be a subtype of LTISystem.

However, this seems to be probably the first time I think a solution to this is a bit tricky, but it could be solved like this. Since we are concerned of the one input to one output poles (which is not that common) we can look at those subsystems with map
```julia
stable = map(isstable, [f[i,j] for i in 1:size(f,1), j in 1:size(f,2)])
```
we should probably make `TransferFunction` iterable so this can be written
```julia
stable = map(isstable, sys)
```
the code could then be
```julia
function dcgain(sys::TransferFunction)
    gains = zeros(size(sys.matrix))
    stable = map(isstable, [sys[i,j] for i in 1:size(sys,1), j in 1:size(sys,2)])
    for idx in eachindex(sys.matrix)
        gains[idx] = stable[idx] ?  _dcgain(sys.matrix[idx], sys.Ts > 0) : NaN
    end
    return gains
 end
```
another alternative (which is my favorite) would be to define `_isstable` for `sisoTf` too, this could make sense and is trivial, then it would simply be
```julia
 dcgain(sys::TransferFunction) =  map(s->_dcgain(s, sys.Ts > 0), sys.matrix)
```
where `_dcgain` is something like
```julia
function _dcgain(sys::SisoTf, discrete::Bool)
    if _isstable(sys, discrete)
        ....
        #Your code here
        ....
    else
        NaN
    end
end
```
A last example, which does not require that Transferfunction is iterable is
```julia
function dcgain(sys::TransferFunction)
    gains = zeros(size(sys.matrix))
    for i in CartesianRange(size(sys))
        gains[i] = isstable(sys[i.I...]) ?  _dcgain(sys.matrix[i], sys.Ts > 0) : NaN
    end
    return gains
 end
```
or a (more ugly?) version
```julia
function dcgain(sys::TransferFunction)
    sisogain(s) = isstable(s) ?  _dcgain(s.matrix[1], sys.Ts > 0) : NaN
    map(i-> sisogain(sys[i.I...]), reshape(collect(CartesianRange(size(sys))),size(sys)))
 end
```
My point is, that yes, it can sometimes be a bit inconvenient, but it is usually not that hard to solve, and I am far from convinced that the alternative is better.

baggepinnen:
I can leave some of my reflections on the types

## SISO/MIMO 

- For the **user**, there should never be a difference in behavior between function calls on SISO systems compared to MIMO systems.
- A statespace system's SISO/MIMO-character is determined soley by the number of columns/rows in the B/C-matrices, and I therefore consider it natural to treat all statespace systems as MIMO, where SISO is just a special case. It would in this respect be natural to also treat all transfer functions visible to the user as MIMO, where a 1x1 system is a special case. How this is represented internally is maybe trickier.
1) A matrix of SISO transfer functions, which themselves are fully fledged transfer functions which could be used by the user, conflicts with all tranferfunctions being MIMO. They would also all hold individual sample-times etc. 
2) A matrix of spcial SISO types not visible to the user would be the case we have today. This took some time for me to get used to, but I do think it is well thought through and works well once one understand the idea.

## Continuous / Discrete / Generalized
In this case, I can agree with @aytekinar that letting the numeric value of a field determine whether the system is discrete or continuous is bad style. This can, I think, be solved by introducing a new parametric type argument, like `TransferFunction{SisoRational, Continuous}`, where `Continuous` is an empty type like
```julia
abstract TFtime
type Continuous <: TFtime
end
type Discrete <: TFtime
end
type Generalized <: TFtime
end
```
These could then be given typealiases if one would like
```julia
typealias TFcont TransferFunction{SisoRational,Continuous}
typealias TFdisc TransferFunction{SisoRational,Discrete}
typealias TFgen TransferFunction{SisoRational,Generalized}
```
similar to how `Vector{T}` and `Matrix{T}` are typealiases for `Array{T,1}`  and `Array{T,2}` respectively.

Please feel free to comment on this. I would also hope you agree that we can not be in more hurry developing this toolbox than that we have time to discuss this enough such that we all agree.

mfalt:
I think your solution seem pretty good @baggepinnen, as a comment to a discussion I had with @aytekinar regarding having noise models in the TransferFunction:

I think this is a good reason to rethink the type system, but I am not convinced what the right solution is. If the noise model would be a property of each input->output relation, then it seems to me that this could be done by having a new `sisoTf` type
```julia
type SisoNoise{S<:SisoTf, T<:NoiseType} <: SisoTf
    sys::S
    noise::T
end
```
or however you want to define it, but this gets trickier if we want to do more drastic changes to `TransferFunction`. We might therefore want to define some abstract TransferFunction type above our TransferFunction which the NoiseTransferFunction (and others) can a subtype of?

mfalt:
I also want to say that I completely agree that we should take all the time we need to discuss these different approaches so that we find a solution that we all like! In particular, we should talk at Reglermöte. We should also set up a some page where we can set up and summarize our ideas, maybe at the wiki?

Issue #22, opened 2016-03-11
Change polynomial implementation to Polynomials.jl


mfalt:
_From @mfalt on February 21, 2016 1:7_

The implementation of `Poly` is outdated and the `roots` function extremely inaccurate. We should consider using the Polynomial.jl package.

_Copied from original issue: mfalt/Control.jl#9_

aytekinar:
We think the same with @neveritt. Since we already require the `Polynomials` package, maybe we can internally save everything in `Polynomials.Poly` type and write methods that would convert calls `tf(num::Array, den::Array)` from the natural way of writing transfer functions to the internal (reversed order) representation of `Polynomials.Poly`.

We can have a look into this if you agree. If you give us whereabouts of the functions that use `Poly`, that would be great. Otherwise, we will check the files accordingly.

mfalt:
Yes, I agree, the interface should not change for the user, it should still be the highest power first. I am not sure where all the functions that use Poly are, but most of them will be in the types/ folder, mainly in sisotf.jl and sisozpk.jl . I think we should try to keep almost all access to the polinomials in the respective types and then provide a consistent interface to the types that can be accessed from all other methods.
For example by implementing numvec/denvec numpoly/denpolythat for each type that returns vectors/polynomials instead of accessing mytype.num.

aytekinar:
So you mean we should implement numvec/denvec to get the polynomial coefficients from highest to lowest degree and numpoly/denpoly to get the actual `Polynomials.Poly` object from the transfer functions?

In this way, we can have methods for `tf` accepting vectors to construct transfer functions from the highest to lowest degree polynomial coefficients, and those accepting Polynomials.Poly to construct from the lowest to highest degre coefficients?

mfalt:
Yes, that is pretty much what I mean. Since we have a couple of different implementations of siso transferfunctions, like tf and zpk, we should try not to assume any kind of underlying structure when writing higher level code.

I have actually not thought about accepting Polynomials.Poly in constructors, but this could be a nice feature. However, I think that we should refrain from actually supplying any variables of Polinomials type to the user unless explicitly asked (for example through a numpoly/denpoly call)

aytekinar:
In the current implementation of `Polynomials.Poly`, one can use `Polynomials.Poly([1, 0, 1], "s")` to get `1 + s^2`. That's why we thought it would be nice to accept `Polynomials.Poly` inputs to construct transfer functions.

[...]

PR #57, opened 2016-05-07
Deprecated num/den for numvec,numpoly...

mfalt:
* Solving the problem #20 of Base.num and Base.den being badly defined for different siso types by deprecating and replacing them with numvec, numpoly, denvec, denpoly.

* Replaced all occurrences of `.num` and `.den` as well as `num(...)`, `den(...)` outside the files for the respective siso types.

* This enables us to start working on moving to `Polynomials` #22 as well as dcgain #15.

Feel free to look at the changes and comment on them. This should be one of the last steps before we can tag a new major version.

### DIFF
--- a/src/synthesis.jl
+++ b/src/synthesis.jl
@@ -175,7 +175,7 @@ end
 """
 function feedback2dof(P::TransferFunction,R,S,T)
     !issiso(P) && error("Feedback not implemented for MIMO systems")
-    tf(conv(poly2vec(numpoly(P)[1]),T),zpconv(poly2vec(denpoly(P)[1]),R,poly2vec(numpoly(P)[1]),S))
+    feedback2dof(numvec(P)[1], denvec(P)[1], R, S, T)
  end
 
 feedback2dof(B,A,R,S,T) = tf(conv(B,T),zpconv(A,R,B,S))
--- a/src/types/sisotf.jl
+++ b/src/types/sisotf.jl
@@ -9,7 +9,7 @@ immutable SisoRational <: SisoTf
             # The numerator is zero, make the denominator 1
             den = one(den)
         end
-        new(num, den)
+        new(copy(num), copy(den))
     end
 end
 SisoRational(num::Vector, den::Vector) = SisoRational(Poly(map(Float64,num)), Poly(map(Float64,den)))
@@ -54,6 +54,7 @@ Base.zero(::SisoRational) = Base.zero(SisoRational)
 Base.length(t::SisoRational) = max(length(t.num), length(t.den))
 
 function Base.num(t::SisoRational)
+    Base.depwarn("`num is deprecated for getting numerator, use `numvec` or `numpoly` instead", :numvec)
     lt = length(t)
     n = zeros(lt)
     n[(lt - length(t.num) + 1):end] = t.num[:]
@@ -61,22 +62,25 @@ function Base.num(t::SisoRational)
 end
 
 function Base.den(t::SisoRational)
+    Base.depwarn("`den` is deprecated for getting denominator, use `denvec` or `denpoly` instead", :denvec)
     lt = length(t)
     d = zeros(lt)
     d[(lt - length(t.den) + 1):end] = t.den[:]
     return d
 end
 
-denpoly(G::SisoRational) = Poly(den(G))
-numpoly(G::SisoRational) = Poly(num(G))
+numvec(t::SisoRational) = t.num[:]
+denvec(t::SisoRational) = t.den[:]
+numpoly(t::SisoRational) = copy(t.num)
+denpoly(t::SisoRational) = copy(t.den)
 
 function evalfr(sys::SisoRational, s::Number)
     S = promote_type(typeof(s), Float64)
-    den = polyval(sys.den, s)
+    den = sys.den(s)
     if den == zero(S)
         convert(S, Inf)
     else
-        polyval(sys.num, s)/den
+        sys.num(s)/den
     end
 end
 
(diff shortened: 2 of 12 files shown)


src/synthesis.jl
```diff
@@ -175,7 +175,7 @@ end
 """
 function feedback2dof(P::TransferFunction,R,S,T)
     !issiso(P) && error("Feedback not implemented for MIMO systems")
-    tf(conv(poly2vec(numpoly(P)[1]),T),zpconv(poly2vec(denpoly(P)[1]),R,poly2vec(numpoly(P)[1]),S))
+    feedback2dof(numvec(P)[1], denvec(P)[1], R, S, T)
```
mfalt: Does this look correct @baggepinnen?

src/types/sisotf.jl
```diff
@@ -9,7 +9,7 @@ immutable SisoRational <: SisoTf
             # The numerator is zero, make the denominator 1
             den = one(den)
         end
-        new(num, den)
+        new(copy(num), copy(den))
```
aytekinar: what happens here if we don't use `copy`?

src/types/transferfunction.jl
```diff
@@ -344,6 +341,72 @@ function minreal(t::TransferFunction, eps::Real=sqrt(eps()))
...
+pole{T<:SisoTf}(::T) = error("pole is not implemented for type $T")
+numvec{T<:SisoTf}(::T) = error("numvec is not implemented for type $T")
+denvec{T<:SisoTf}(::T) = error("denvec is not implemented for type $T")
+numpoly{T<:SisoTf}(::T) = error("numpoly is not implemented for type $T")
+denpoly{T<:SisoTf}(::T) = error("denpoly is not implemented for type $T")
+
+"""
+`numpolys = numpoly(tf::TransferFunction)`
+
+Get an `Array` of of size `(ny,nu)` containing `Poly`s representing the
```
aytekinar: small typo --- 2x`of` in the documentation.

aytekinar:
I will change the `dcgain` properly once this commit is pulled onto `master`.

src/types/sisotf.jl
mfalt: Our Poly types are immutable, so it won't be a problem now, but when we switch to Polynomials the Poly will be mutable so if the user sends Poly to the constructor then we need a copy so it will not be changed though another reference later.

src/types/transferfunction.jl
mfalt: Fixed.

src/types/sisozpk.jl
```diff
@@ -86,13 +88,11 @@ function zp2polys(vec)
     polys
 end
 
-function numpoly(G::SisoZpk)
-    zpolys = zp2polys(G.z)
-end
+numvec(t::SisoZpk) = numpoly(t)[:]
+denvec(t::SisoZpk) = denpoly(t)[:]
```
aytekinar: `numvec` and `denvec` should return arrays of the same size for a given `sys`, that is the `length` of the system, imo.

mfalt: It did that previously, but I didn't really understand why. Do you have a good reason?

aytekinar:
There are two problems:

*  `zpk(sys::SisoRational)` returns `0`,
```julia
using ControlSystems
sys1 = tf([1;2], [1;2;2])
sys2 = zpk(sys1)
```
*  `length` returns different lengths for similar systems.
```julia
sys1 = tf([1;2], [1;2;1])
sys2 = zpk([-2], [-1;-1], 1)
length(sys1.matrix[1])
length(sys2.matrix[1])
```

I did not want to open a new issue, as I have noticed these while committing to #57.

mfalt:
I am not able to reproduce the first bug, are you on this latest version of numden?

As for the second one, yes, it seems like the definitions are different, what do we want it to be?
Base.length(t::SisoZpk) = max(length(t.z), length(t.p))
Base.length(t::SisoRational) = max(length(t.num), length(t.den))

aytekinar:
The first problem is reproduced if you accept the changes in #67. The second one is checking the cardinality of poles for the zpk case and the length of the polynomial in the rational case. Obviously, the cardinality of the poles is the correct one --- `length(ControlSystems.Poly)` returns the number of coefficients in the polynomial, which is `length(roots)+1`

aytekinar:
Fixing `Base.length(sys::SisoRational)` also fixes the `zpk(sys::SisoRational)` problem mentioned above.

mfalt:
> Obviously, the cardinality of the poles is the correct one

So not `Base.length(t::SisoZpk) = max(length(t.z), length(t.p))`? My point is that the length of a system is usualy not defined. Maybe something like `order` would be more appropriate, referring to the corresponding differential equation, but then the numerator should be taken into consideration.

aytekinar:
I thought `length` was standing there for the `degree` of the system, since it was comparing the number of poles and zeros. I was assuming we were dealing with causal systems, as MATLAB does not allow for non-causal systems directly.

mfalt:
Yes, `degree` would be a better name. So there is nothing inhibiting the user to create a non-causal /direct derivate until the user tries to simulate the system (or make it into ss). I prefer this model since it gives the user much more freedom. I think padding `numvec` and `denvec` to `degree+1` makes sense with `degree` as
```julia
degree(t::SisoZpk) = max(length(t.z), length(t.p))+1
degree(t::SisoRational) = max(length(t.num), length(t.den))
```
and then deprecate or change `length` to the same?

aytekinar:
That also makes sense for me! The way I implemented was due to this, basically.

baggepinnen:
This PR is rather outdated by now, but it implements some tests we might want to extract into a new PR?

Issue #59, opened 2016-05-17
"norm" for stable system returns "Inf"


denglerchr:
`sys = tf([1, -1.45, 0.475], [1, -2*0.5, 0.5^2], 1.0); 
norm(sys)`
returns "Inf". Both poles are at 0.5, so the norm should be bounded. Same code in Matlab returns 1.1269.

mfalt:
Thanks. I fixed this in pull request #60, it was a bug with the direct term causing it to be Inf even if the system is discrete. The new version returns `1.1269427669584644`. I will pull this as soon as possible, but it may take a few weeks before we tag a new version.

mfalt:
I pulled this into master now, to get the changes you have to do a
`Pkg.checkout("ControlSystems")`
or wait until we tag a new version and do a
`Pkg.update()`
Thanks for reporting this.

PR #60, opened 2016-05-17
Fixed bug #59

mfalt:
This fixes bug 59, but I don't want to pull this because it seems that we have type instability in returning Inf that should be fixed first. Should the whole matrix be filled with Inf or just some of the elements?


mfalt:
Closed in favor of rebase #61.

PR #66, opened 2016-05-23
Compute the Hinfininty norm for discrete-time systems

jleny:
I had left this when I implemented norm(sys,Inf). Also, norm(sys,Inf) now computes the Linifinity norm, as in Matlab, rather than Hinfinity. So it will return a finite value in general even for unstable systems (with no pole on the imaginary axis / unit circle).


jleny:
Is there anything needed to merge this pull request?

mfalt:
I think It looks good, but it would be nice if we could address the issue of type instability, do you have any ideas? Maybe a separate function `linf` that returns both the frequency and the value, since all other `norm` functions only return a single value?

jleny:
ok, I don't think (or don't know) that there is a way to simulate varargout in Julia. I don't have a better idea for now than having a different function such as linf returning the frequency as well, as you propose. I'll look into this and make the change when I have a moment.

jleny:
So in this version, norm always returns a number, even for the L-infinity norm. If you want the peak frequency as well, you can use norminf.

mfalt:
I think this looks like a good solution, thanks! Do you have any comments @baggepinnen ?

baggepinnen:
An option to keep the function type stable is to have the second argument be a `Val` type. That way, the correct method is chosen on dispatch, e.g.
`norm(G, Val{2})`
http://docs.julialang.org/en/release-0.4/manual/types/#value-types

olof3:
That is a nice idea, maybe one could define, 
```const hinf = Val{:hinf}```,
although it adds some global variables and maybe is a confusing construction.

Otherwise you could introduce a function `h2norm` (would be nice anyway), and encourage use of either `h2norm` and `norminf` and not `norm`.

Some other thoughts, 

1) Maybe the function should be called infnorm, as that's closer to how you say infinity norm. But why not have two functions, both linfnorm and hinfnorm, then one directly knows the behavior of the function

2) Would expect `norminf`, to just return the norm, but one also want to be able to obtain the peak frequency somehow. What happens if there are many peak frequencies? Say for example G(s) = 1?

jleny:
Some thoughts:
1) Type stability: I'm confused, I thought you just meant that you want the function 'norm' to always return an output of the same type, I don't see how using value types helps. Regarding the use of value types or not, the Base.LinAlg.norm function for vectors decides which one to use based on a test at run-time, I'm not sure it's worth doing something more complicated here, especially if it changes the syntax with respect to the base norm. A global constant named hinf will be confusing, I'm sure it's a variable name people use a lot.
2) Nomenclature: I will let you decide here which names you want to use, if you want to have multiple functions implementing the same things, etc. My thoughts: again in Base.LinAlg.norm, they define norm1, norm2, normInf, not 1norm, 2norm or infnorm. What I find nice with keeping norm first for example is that all the functions will show up next to each other in documentations when you look for norm related functions. Keeping norm to compute H2 and Hinf norms is also close to the Matlab syntax, which I think is something you should try to keep. That said, Matlab also has functions hinfnorm and h2norm, SciLab has h_norm and linfnorm, etc. My choice of norminf was just because norm(sys,Inf) in Matlab calls getPeakGain (which is another name for hinfnorm...), which calls norminf_, but I have no problem if we change the name to hinfnorm and add linfnorm as well.
3) Regarding cases with multiple peak frequencies, since the computations are numerical, I expect that we will just return one of the peak frequencies. However, I need to look in more details at the behavior of the algorithm in this case, and I see that G(s)=1 breaks the function. (by the way, in this case norm(sys,2) also returns an error rather than Inf, which is not particularly nice, and also in general it would be good to send a warning to the user for the presence of direct feedthrough when computing H2 norms, as in Matlab).

olof3:
1) Yes, probably not worth the effort, especially not if there are functions to call directly.

2) Good points to take into consideration.

baggepinnen:
It would help with type stability in the sense that it would be known at compile-time what method of norm will be called, instead of a runtime check of the value of the second numerical argument, i.e. `2` or `Inf`. If a certain method of norm has a different return signature, type instability occurs in the case of a runtime check, while no instability occurs with the use of `Val` types. At least this is my understanding, do I make any sense?

mfalt:
Although, it is possible to have good performance with a varying number of outputs arguments for `norm` using the `Val` types, I don't think it would result in good user experience. I think it is better to have the `norm` always return one value, as in Base, and to have separate functions for other use cases, as in this implementation. I don't think we will have any type instability issues since the output type of `norm` should be same (`Float64`), for both `Inf` and `2` norm. At a later restructuring of the code, we should have the output depend on the type of the TransferFunction, like `TransferFunction{Float64}` or `TransferFunction{Complex128}`, but this is currently not implemented so it is not an issue yet.

mfalt:
I think we could pull this if there is no things that we need to change first? I assume that we would still want these two functions, and even some breaking changes later (like the number of output arguments of norminf) shouldn't be that disruptive.

jleny:
Let me check something first. As we pointed out above, the function returns an error for G(s)=1 for example. I should be able to fix this in the next few days.

jleny:
Ok, the static gain case had to be treated separately.

jleny:
Perhaps my last message was confusing, but this can be pulled now.

mfalt:
Thanks, yes I missed your last comment. Great work!

PR #71, opened 2016-06-29
Newplotssyntax

mfalt:
Should make it possible to do plots again after Plots 0.7 update. Will pull when tests pass.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.4
-Plots
+Plots v0.7.4
 Polynomials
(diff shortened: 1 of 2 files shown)


baggepinnen:
Looked through it (fast) and seems alright. I'll have a look into recipes #72

baggepinnen:
This PR created some problems
```
ERROR: fixed packages introduce conflicting requirements for Plots: 
         ControlSystems requires versions [0.7.4,∞) [none of the available versions can satisfy this requirement]
         Robotlib requires versions [0.0.0-,∞)
         SymPy requires versions [0.0.0-,∞)

```
I think the requirements on Plots must be given such that we accept version numbers higher than 0.7.4  [0.7.4-,∞)?

PR #73, opened 2016-06-30
Fixes #69 - time response for discrete system with specified final time

### LINKED ISSUE #69: Time response for discrete system with specified final time throws error
```julia
G = tf([1], [1; zeros(3)], 1)
y, t, x = step(G, 10)
```
gives

ERROR: Time vector must match sample time for discrete system

The problem seems to be that the wrong time vector is created for the step response.

olof3:
Fix to avoid problem when final time is not multiple of sample time. Added tests.


mfalt:
Looks great, thanks, should poss treats when merged with the new plots syntax.

PR #82, opened 2016-08-04
Eigvals for computing statespace poles.

olof3:
`eigvals` is bit more efficient than using `eig` since the eigenvectors are not needed. There were not so many tests using pole on state space systems, so I added some.


mfalt:
Looks good, but we should probably sort the poles in some way before comparing them, in case the eigvals implementation changes?

olof3:
Yes, that should definitely be done. I guess that goes for the existing tests of pole and tzero as well?

Matlab has a function cplxpair which sorts in order of increasing real parts and additionally tries to pair up conjugate pairs. In order of decreasing magnitude of the imaginary part (if there are multiple pairs with the same real part.

Functionality like cplxpair could be useful for printing the poles, but it seems like there is some numerics to think about. 

To me the following could be a reasonable ordering 
1) Primarily according to increasing real part
2) In case of the same real part according to increasing imaginary part

Note that it might not work if there are not two conjugate pairs of eigenvalues with the same real part, then they may be sorted arbitrarily due to numerical noise.

Another approach would be to do an approximate comparison of two unordered collections. If the number of poles is not to big, this should be ok. Couldn't find any existing function for this though.

Another way would be 

I don't know how Lapack does it?

In https://github.com/JuliaLang/julia/blob/master/test/linalg/eigen.jl they just seem to check that eigvals and eigfactor gives the same eigenvectors, apparently assuming that they are ordered in the same way.

Any ideas?

mfalt:
I think it would be nice to sort them by increasing real part and then imaginary, but we should make sure that pairs are collected together. I'm still not sure if we already have a method for this, there should be something similar in the `minreal` implementation, although it might not be very good.

baggepinnen:
> I think it would be nice to sort them by increasing real part and then imaginary,
I'm just throwing this out there, wouldn't it be interesting to have them sorted in this order
- Positive or negative real part
- Magnitude, i.e. `abs(p)`
- Argument

mfalt:
You mean something like this?
```julia
[
-10,
-1 - 1im
-1.4142
-1 + 1im
5
2
1
]
```
It seems messy not to have the complex pairs together and to have first increasing real part and then decreasing.

test/test_analysis.jl
```diff
@@ -105,6 +105,20 @@ sys = s*(s + 1)*(s^2 + 1)*(s - 3)/((s + 1)*(s + 4)*(s - 4))
 @test_approx_eq pole([sys sys]) [-1.0, 4.0, -4.0, -1.0, 4.0, -4.0]
 @test_approx_eq pole(ex_11) eig(ex_11.A)[1]
 
+@test_approx_eq pole(ex_8) [-3.383889568918823 + 0.000000000000000im
```
baggepinnen: How about
```julia
poles = [-3.383889568918823 + 0.000000000000000im
                            -2.199935841931115 + 0.000000000000000im
                            -0.624778101910111 + 1.343371895589931im
                            -0.624778101910111 - 1.343371895589931im
                            -0.083309192664918 + 0.487701968391972im
                            -0.083309192664918 - 0.487701968391972im]
approxin(el,col) = reduce(|,false,el.≈col)
@test all(approxin(p,poles) for p in pole(ex_8))
```

baggepinnen: This ensures all poles are present but does not care about order.

test/test_analysis.jl
```diff
@@ -106,6 +106,22 @@ sys = s*(s + 1)*(s^2 + 1)*(s - 3)/((s + 1)*(s + 4)*(s - 4))
...
 @test pole(ex_11) ≈ eig(ex_11.A)[1]
 
+poles = [-3.383889568918823 + 0.000000000000000im
+                            -2.199935841931115 + 0.000000000000000im
+                            -0.624778101910111 + 1.343371895589931im
+                            -0.624778101910111 - 1.343371895589931im
+                            -0.083309192664918 + 0.487701968391972im
+                            -0.083309192664918 - 0.487701968391972im]
+approxin(el,col) = reduce(|,false,el.≈col)
+@test all(approxin(p,poles) for p in pole(ex_8))
```
baggepinnen: Add also reverse test
`@test all(approxin(p,pole(ex_8)) for p in poles)`
Move definition of `approxin` to `tests/framework.jl`

baggepinnen: Glömde flytta definitionen av `approxin`?

PR #84, opened 2016-08-08
Updates to work with julia 0.5

mfalt:
First draft of julia 0.5 compability. Accidentally included the changes in bd7b7647d715524f0143dde69296ceb1d93b82dd to run the tests even if ControlExamplePlots is missing.

### DIFF
--- a/src/freqresp.jl
+++ b/src/freqresp.jl
@@ -85,11 +85,11 @@ Notation for frequency response evaluation.
 - F(omega,true) evaluates the discrete-time transfer function F at i*Ts*omega
 - F(z,false) evaluates the discrete-time transfer function F at z
 """ ->
-function Base.call(sys::TransferFunction, s)
+function (sys::TransferFunction)(s)
     evalfr(sys,s)
 end
 
-function Base.call(sys::TransferFunction, z_or_omega::Number, map_to_unit_circle::Bool)
+function (sys::TransferFunction)(z_or_omega::Number, map_to_unit_circle::Bool)
     @assert !iscontinuous(sys) "It makes no sense to call this function with continuous systems"
     if map_to_unit_circle
         isreal(z_or_omega) ? evalfr(sys,exp(im*z_or_omega.*sys.Ts)) : error("To map to the unit circle, omega should be real")
@@ -98,7 +98,7 @@ function Base.call(sys::TransferFunction, z_or_omega::Number, map_to_unit_circle
     end
 end
 
-function Base.call(sys::TransferFunction, s::AbstractVector, map_to_unit_circle::Bool)
+function (sys::TransferFunction)(s::AbstractVector, map_to_unit_circle::Bool)
     @assert !iscontinuous(sys) "It makes no sense to call this function with continuous systems"
     freqresp(sys,s)
 end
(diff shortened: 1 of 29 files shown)


src/types/transferfunction.jl
```diff
@@ -81,7 +81,6 @@ Base.convert(::Type{SisoGeneralized}, sys::SisoZpk) = convert(SisoGeneralized, c
 
 Base.convert(::Type{SisoRational}, sys::SisoGeneralized) = SisoRational(sys.expr)
 Base.convert(::Type{SisoZpk}, sys::SisoGeneralized) = convert(SisoZpk, SisoRational(sys.expr))
-Base.convert(::Type{TransferFunction{SisoZpk}}, s::TransferFunction) = zpk(s)
```
mfalt: This was a duplicate, which now throws warnings.

src/types/transferfunction.jl
```diff
@@ -116,6 +115,10 @@ tzero(sys::SisoTf) = error("tzero is not implemented for type $(typeof(sys))")
 ==(a::SisoTf, b::SisoTf) = ==(promote(a,b)...)
 !=(a::SisoTf, b::SisoTf) = !(a==b)
 isapprox(a::SisoTf, b::SisoTf; kwargs...) = isapprox(promote(a,b)...; kwargs...)
+
+# Promote_op types
+Base.promote_op{T<:SisoTf}(::Any, ::Type{T}, ::Type{T}) = T
```
mfalt: Added this for all operators, not just `*`  for which it was needed, I don't see that this would be a problem?

src/utilities.jl
```diff
@@ -105,3 +105,12 @@ numpoly(G::TransferFunction) = map(numpoly, G.matrix)
...
 poly2vec(p::Poly) = p.a[1:end]
+
+"""
+outs = index2range(ind1, ind2)
+Helper function to convert indexes with scalars to ranges. Used to avoid dropping dimensions
+"""
+index2range(ind1, ind2) = (index2range(ind1), index2range(ind2))
+index2range{T<:Number}(ind::T) = ind:ind
+index2range{T<:AbstractArray}(ind::T) = ind
+index2range(ind::Colon) = ind
```
mfalt: These were all the use-cases I could come up with, please suggest more. We could add a wild-card, but this way we will not have any unexpected behavior.

test/framework.jl
```diff
@@ -1,103 +1,11 @@
 module CustomTest
```
mfalt: I scrapped the old system because it was not compatible with julia 0.5. The new output is slightly less informative in some ways but generates a nice overview of failes/passed tests.

src/freqresp.jl
```diff
@@ -138,7 +138,7 @@ function sigma(sys::LTISystem, w::AbstractVector)
     nw, ny, nu = size(resp)
     sv = Array(Float64, nw, min(ny, nu))
     for i=1:nw
-        sv[i, :] = svdvals(squeeze(resp[i, :, :],(1)))
+        sv[i, :] = svdvals(resp[i, :, :])
```
mfalt: We no longer have to squeeze outputs!

mfalt:
Could someone look through this and approve the changes? All tests pass but I will fix a few warnings before I pull. We might want to make this version 0.2.0 and backport some bug fixes to 0.1.* to let users catch up with julia 0.5.
@baggepinnen @olof3

olof3:
I've run it with 0.5 and it works great. I'd say just go ahead!

PR #86, opened 2016-08-09
WIP: First prototype for statespace systems with given numeric type.

olof3:
Early attempt for support of statespace models with matrices of different numeric types after discussion with @mfalt, @baggepinnen. Different numeric types gives the user more flexibility, for example the following types of matrices could be useful,

Complex - There are some applications of this in power systems and electronics. Matlab doesn't handle this case very well.
BigFloat - When there are numerical issues or higher precision is required
Symbolic

This first prototype is limited to StateSpace models.

All tests pass except one for `tzero`, which is conversion related, and some for the printing of statespace models.

Does this seem reasonable?

Should the user be warned when he creates a complex system as in matlab?

Should there be a separate function `complexss` so that complex systems are not created by mistake?

It is a surprisingly well contained change, once we decide on how to handle the conversion in tzero, it could in principle be pulled, and further additions could be added separately.

TODO:
Does `tzero` work the same for complex systems?
Plotting functions for complex systems
Implement `conj` with correct behavior for complex systems

### DIFF
--- a/src/types/statespace.jl
+++ b/src/types/statespace.jl
@@ -2,11 +2,11 @@
 ##                      Data Type Declarations                     ##
 #####################################################################
 
-type StateSpace <: LTISystem
-    A::Matrix{Float64}
-    B::Matrix{Float64}
-    C::Matrix{Float64}
-    D::Matrix{Float64}
+type StateSpace{T} <: LTISystem
+    A::Matrix{T}
+    B::Matrix{T}
+    C::Matrix{T}
+    D::Matrix{T}
     Ts::Float64
     nx::Int
     nu::Int
@@ -15,8 +15,8 @@ type StateSpace <: LTISystem
     inputnames::Vector{String}
     outputnames::Vector{String}
 
-    function StateSpace(A::Matrix{Float64}, B::Matrix{Float64},
-            C::Matrix{Float64}, D::Matrix{Float64}, Ts::Float64,
+    function StateSpace(A::Matrix{T}, B::Matrix{T},
+            C::Matrix{T}, D::Matrix{T}, Ts::Float64,
             statenames::Vector{String}, inputnames::Vector{String},
             outputnames::Vector{String})
         nx = size(A, 1)
@@ -51,11 +51,19 @@ type StateSpace <: LTISystem
     end
 end
 
-function StateSpace(A::Array, B::Array, C::Array, D::Array, Ts::Real,
+# Or is it already known when the constructor is called what type fo
+function StateSpace(A::Union{Number,Array}, B::Union{Number,Array}, C::Union{Number,Array}, D::Union{Number,Array}, Ts::Real,
         statenames::Vector{String}, inputnames::Vector{String},
         outputnames::Vector{String})
-    return StateSpace(float64mat(A), float64mat(B), float64mat(C),
-            float64mat(D), map(Float64, Ts), statenames, inputnames, outputnames)
+
+        (A, B, C, D) = promote(ensure_matrix(A), ensure_matrix(B),
+                               ensure_matrix(C), ensure_matrix(D))
+
+        # Why do we need to specify T, isn't the inner constructor more specific anyway?
+        T = promote_type(typeof(A[1]), typeof(B[1]), typeof(C[1]), typeof(D[1]))
+
+        return StateSpace{T}(A, B, C, D, map(Float64, Ts),
+                        statenames, inputnames, outputnames)
 end
 
 #####################################################################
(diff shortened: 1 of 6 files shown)


src/types/statespace.jl
```diff
@@ -15,8 +15,8 @@ type StateSpace <: LTISystem
     inputnames::Vector{String}
     outputnames::Vector{String}
 
-    function StateSpace(A::Matrix{Float64}, B::Matrix{Float64},
-            C::Matrix{Float64}, D::Matrix{Float64}, Ts::Float64,
+    function StateSpace(A::Matrix{T}, B::Matrix{T},
```
mfalt: Is this really working? I was expecting to see StateSpace{T}

src/types/statespace.jl
```diff
@@ -65,37 +73,29 @@ end
...
+    # For accepting scalar D
+    if D == 0
+        D = zeros(size(C,1),size(B,2))
+    end
+
     nu = size(B, 2)
     ny, nx = size(C, 1, 2)
     kvs = Dict(kwargs)
-    statenames = validate_names(kvs, :statenames, nx)
+    statenames = validate_names(kvs, :statenames, nx) # Are these needed?, a transfer function shouldn't have to carry along variable names
```
mfalt: I'm pretty sure they wil default to empty strings in this function

src/types/statespace.jl
```diff
@@ -51,11 +51,19 @@ type StateSpace <: LTISystem
...
+function StateSpace(A::Union{Number,Array}, B::Union{Number,Array}, C::Union{Number,Array}, D::Union{Number,Array}, Ts::Real,
         statenames::Vector{String}, inputnames::Vector{String},
         outputnames::Vector{String})
-    return StateSpace(float64mat(A), float64mat(B), float64mat(C),
-            float64mat(D), map(Float64, Ts), statenames, inputnames, outputnames)
+
+        (A, B, C, D) = promote(ensure_matrix(A), ensure_matrix(B),
+                               ensure_matrix(C), ensure_matrix(D))
+
+        # Why do we need to specify T, isn't the inner constructor more specific anyway?
```
mfalt: You have to specify `T` because you are creating an element of the type `StateSpace{T}`, which tells the inner constructor what `T` is. Elements will for example be converted to this type (I think)

src/utilities.jl
```diff
@@ -56,6 +56,11 @@ float64mat(A::Matrix) = map(Float64,A)
 float64mat(A::Matrix{Float64}) = A
 float64mat(A::Real) = float64mat([A])
 
+ensure_matrix(A::Vector) = reshape(A, size(A, 1), 1)
```
mfalt: With this method, we should probably replace all occurrences of `float64mat`.

src/types/statespace.jl
```diff
@@ -51,11 +51,19 @@ type StateSpace <: LTISystem
...
+        T = promote_type(typeof(A[1]), typeof(B[1]), typeof(C[1]), typeof(D[1]))
```
mfalt: might be better to use `eltype(A)` here instead of `typeof(A[1])`. The only case where I think your code will fail is for a `0x0` system, but I think `eltype` is more elegant.

src/types/statespace.jl
```diff
@@ -65,37 +73,29 @@ end
...
     kvs = Dict(kwargs)
-    statenames = validate_names(kvs, :statenames, nx)
+    statenames = validate_names(kvs, :statenames, nx) # Are these needed?, a transfer function shouldn't have to carry along variable names
     inputnames = validate_names(kvs, :inputnames, nu)
     outputnames = validate_names(kvs, :outputnames, ny)
-    return StateSpace(A, B, C, D, Ts, statenames, inputnames, outputnames)
-end
 
-# Function for accepting scalars
-function ss(A::Union{Real,Array}, B::Union{Real,Array}, C::Union{Real,Array}, D::Union{Real,Array}, args...; kwargs...)
```
mfalt: Nice that you were able to remove this!

src/types/tf2ss.jl
```diff
@@ -31,13 +31,22 @@ function Base.convert(::Type{StateSpace}, t::TransferFunction)
...
 
-Base.convert(::Type{StateSpace}, t::Real) = ss(t)
 
-Base.promote_rule{T<:SisoTf}(::Type{StateSpace}, ::Type{TransferFunction{T}}) = StateSpace
-Base.promote_rule{T<:Real}(::Type{StateSpace}, ::Type{T}) = StateSpace
+
+Base.convert{T}(::Type{StateSpace{T}}, n::Number) = ss(convert(T, n))
+Base.promote_rule{T1<:Number, T2<:SisoTf}(::Type{StateSpace{T1}}, ::Type{TransferFunction{T2}}) = StateSpace{promote_type(T1,Float64)}#promote_type vs promote_rule?
+Base.promote_rule{T1<:Number, T2<:Number}(::Type{StateSpace{T1}}, ::Type{T2}) = StateSpace{promote_type(T1,T2)}
+Base.promote_rule{T1<:Number, T2<:Number}(::Type{StateSpace{T1}}, ::Type{StateSpace{T2}}) = StateSpace{promote_type(T1,T2)}
```
mfalt: We should probably be consistent in restricting the types, so if we have `Number` here, we should have the same in the definition of `StateSpace`, or we should leave it as `Any` here.

src/types/tf2ss.jl
```diff
@@ -31,13 +31,22 @@ function Base.convert(::Type{StateSpace}, t::TransferFunction)
...
     return ss(A, B, C, D, t.Ts, inputnames=t.inputnames, outputnames=t.outputnames)
 end
 
-Base.convert(::Type{StateSpace}, t::Real) = ss(t)
 
-Base.promote_rule{T<:SisoTf}(::Type{StateSpace}, ::Type{TransferFunction{T}}) = StateSpace
-Base.promote_rule{T<:Real}(::Type{StateSpace}, ::Type{T}) = StateSpace
+
+Base.convert{T}(::Type{StateSpace{T}}, n::Number) = ss(convert(T, n))
+Base.promote_rule{T1<:Number, T2<:SisoTf}(::Type{StateSpace{T1}}, ::Type{TransferFunction{T2}}) = StateSpace{promote_type(T1,Float64)}#promote_type vs promote_rule?
```
mfalt: I think this is correct, you define a `promote_rule` but call `promote_type`. I think the logic might be that `promote_type` can handle multiple arguments in any order but you only have to define one `promote_rule`.

test/test_complex_sys.jl
```diff
@@ -0,0 +1,38 @@
...
+@test ss(1,1,1,0.5im) + 0.5im == ss(1,1,1,1.0im)
+
+
+# Just try to add a transfer function with a complex
+# state-space model to see that nothing crashes
+s = tf("s")
+ss(1.0 + 1im,1,1,1.0) + 1/(s+1)
+
+
+# Add test for printing complex system
```
mfalt: We should probably also have many tests for promotion and conversion of systems. And modify all tests that use StateSpace to also test with for example integers or big floats. We could for example wrap the tests in
```julia
for T in [Float64, Int64, Complex128, BigFloat]
A = getSomeData(T)
#do some tests
#test that outputs are of type T
#and so on
end
```

mfalt:
It looks really nice, I added several comments to the code. Should be consider doing the same thing for `TransferFunction` before we actually pull this?

src/utilities.jl
```diff
@@ -56,6 +56,11 @@ float64mat(A::Matrix) = map(Float64,A)
...
+ensure_matrix{T<:Number}(A::Matrix{T}) = A
+ensure_matrix{T<:Number}(A::T) = ensure_matrix([A])
```
mfalt: Alternative: `fill(A, 1,1)` or `[A]'`

mfalt:
Given that `A'` conjugates complex numbers, we should probably look through the code for any such occurrences and change them to `transpose` now that we are allowing complex numbers.

mfalt:
All this should be pulled from another branch

PR #99, opened 2017-05-25
use 0.6.0-pre as minimum julia version

tkelman:
since `abstract type` syntax wouldn't work on early 0.6.0-dev versions,
better to stick to the julia-0.5-compatible versions of the package there


mfalt:
Thanks!

PR #113, opened 2017-08-29
Implement adaptive step-siz rootlocusplot.

baggepinnen:
About 4+ orders of magnitude faster than previous implementation with several orders of magnitude better accuracy.


baggepinnen:
I don't think there's a test for rootlocusplot. Maybe we can set that up
tomorrow.

It only works for siso and have been so before as well. Should be
reasonable easy to extend to mimo, but it gets hard to interpret the result
IMO.

PR #116, opened 2017-11-06
Better logticks and defaults for bodeplot

mfalt:
Fixes so xticks and yticks obeys ylims and xlims. Adds ylimsphase keyword to bodeplot.

### DIFF
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -38,11 +38,12 @@ function setPlotScale(str::AbstractString)
     _PlotScale, _PlotScaleFunc, _PlotScaleStr = plotSettings
 end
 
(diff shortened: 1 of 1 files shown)


src/plotting.jl
```diff
@@ -260,9 +261,17 @@ bodeplot
                 if !plotphase
                     continue
                 end
+                #Let user specify limits on phase independently
+                ylimsphase = if isa(ylimsphase, Void)
+                    (floor(minimum(phasedata)/180)*180,
```
baggepinnen: After trying
```julia
g = tf(1,[1,1])
bodeplot(g)
```
I wonder if not 90 would be better than 180?

PR #118, opened 2018-01-31
[WIP] Autodiff

baggepinnen:
Major overhaul of types. All LTISystem types now accept arbitrary array types, e.g., TrackedArrays, GPUArrays or Matrix{DualNumber}. WIP PR created to trigger tests


test/test_zpk.jl
```diff
@@ -24,10 +24,10 @@ z = zpk("z", 0.005)
 
 # TESTS
 # Constructors
-@test s == zpk([0], [], 1)
-@test z == zpk([0], [], 1, 0.005)
-@test C_022 == zpk(vecarray(Int64, 2, 2, [], [], [], []), vecarray(Int64, 2, 2, [], [], [], []), [4 0; 0 4])
-@test D_022 == zpk(vecarray(2, 2, [], [], [], []), vecarray(2, 2, [], [], [], []), [4 0; 0 4], 0.005)
+@test s == zpk([0], Int64[], 1)
```
mfalt: I understand why, but maybe we can do a smarter check for empty vectors in the constructors when types are different, e.g.
`T = promote_type(typeof.(inputs[find(x-> x != Any[], inputs)])...)`

src/types/promotion.jl
```diff
@@ -0,0 +1,202 @@
+Base.eltype{T}(::Type{Poly{T}}) = T
+Base.eltype{T}(::Type{TransferFunction{T}}) = T
+Base.eltype{T}(::Type{SisoZpk{T}}) = T
+Base.eltype{T}(::Type{SisoRational{T}}) = T
+Base.eltype{T}(::Type{StateSpace{T}}) = T
+
+
+primitivetype{T}(::Type{Poly{T}}) = primitivetype(T)
```
mfalt: A small docstring of the intent could be very helpful here for other developers, same for arraytype, responsetype, primitivereal

src/types/transferfunction.jl
```diff
@@ -437,7 +407,8 @@ function *(t1::TransferFunction, t2::TransferFunction)
     elseif t1.Ts != t2.Ts
         error("Sampling time mismatch")
     end
-    matrix = t1.matrix*t2.matrix
+    T = promote_type(eltype(t1.matrix), eltype(t2.matrix))
+    matrix = convert(Matrix{T}, *(promote(t1.matrix,t2.matrix)...))
```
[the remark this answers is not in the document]
baggepinnen: True, much simpler with new convert rules

src/utilities.jl
```diff
@@ -47,14 +47,13 @@ end
 
 ## Helpers (not public) ##
 
-# Convert the argument to a Matrix{Float64}
-function float64mat(A::Vector)
+# Convert the argument to a Matrix{T}
+function Tmat(A::Vector, T)
```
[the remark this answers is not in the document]
baggepinnen: Changed name to `to_matrix` Did not want to add constructors to a base type with possible base type arguments.

test/test_connections.jl
```diff
@@ -100,15 +100,15 @@ Czpk_222 = [Czpk_221; Czpk_221]
...
 @test [tf(1), zpk(2)] == arr
-arr2 = Array{TransferFunction{ControlSystems.SisoRational},1}(2)
+arr2 = Array{typeof(tf(1, 0.1)),1}(2)
 arr2[1] = tf(1, 0.1); arr2[2] = tf(2, 0.1);
 @test [tf(1, 0.1), tf(2, 0.1)] == arr2
-arr3 = Array{StateSpace,1}(3)
-arr3[1] = ss(0); arr3[2] = ss(1); arr3[3] = ss(2)
-@test [0, zpk(1), ss(2)] == arr3
+arr3 = Array{typeof(ss(0.)),1}(3)
+arr3[1] = ss(0.); arr3[2] = ss(1.); arr3[3] = ss(2.)
```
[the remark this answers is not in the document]
baggepinnen: Added back old test

test/test_zpk.jl
baggepinnen: Issue #121

src/types/promotion.jl
baggepinnen: Done :)

test/test_simplification.jl
```diff
@@ -21,6 +21,6 @@ sysmin = minreal(sys)
 t = 0:0.1:10
 y1,x1 = step(sys,t)[[1,3]]
 y2,x2 = step(sysmin,t)[[1,3]]
-@test sum(abs2,y1.-y2) < √(eps()) # Test that the output from the two systems are the same
+@test sum(abs2,y1.-y2) < 1e-6 # Test that the output from the two systems are the same
```
mfalt: Did we actually lose accuracy here? Currently testing on my machine.

test/test_synthesis.jl
```diff
@@ -1,6 +1,6 @@
 @testset "test_synthesis" begin
-P = tf(1,[1,1])
-C = tf([1,1],[1,0])
+P = tf(1.,[1.,1])
+C = tf([1.,1],[1.,0])
```
mfalt: We should add issue that minreal no longer works properly for Int systems.

test/test_simplification.jl
mfalt: This is some serious loss in accuracy, from
maximum(abs.(y1.-y2))  = 7.0921046813055e-13
to
maximum(abs.(y1.-y2)) = 0.0002794242369802413

baggepinnen: Before simulation was carried out by discretizing. ZOH sampling provides exact solution on time steps, but since we now solve in cont. time, we get lower acc. unless we decrease ODE tolerance

work_on_type_system/src/connections.jl
```diff
@@ -0,0 +1,194 @@
...
+series(s1::LTISystem, s2::LTISystem) = s2*s1
+
+@doc """
+`series(s1::LTISystem, s2::LTISystem)`
+
+Connect systems in parallel, equivalent to `s2+s1`
+""" ->
+parallel(s1::LTISystem, s2::LTISystem) = s1 + s2
+
+append() = LTISystem[]
```
baggepinnen: This is a quite generic name for a function and adding a method with zero args might cause issues

work_on_type_system/src/connections.jl
```diff
@@ -0,0 +1,194 @@
...
+    end
+    A = blkdiag([s.A for s in systems]...)
+    B = blkdiag([s.B for s in systems]...)
+    C = blkdiag([s.C for s in systems]...)
+    D = blkdiag([s.D for s in systems]...)
+    return StateSpace(A, B, C, D, Ts, states, inputs, outputs)
+end
+
+function append(systems::TransferFunction...)
+    Ts = systems[1].Ts
```
baggepinnen: Should we have an API for accessing this field?

work_on_type_system/src/connections.jl
```diff
@@ -0,0 +1,194 @@
...
+    if !all([s.Ts == Ts for s in systems])
```
baggepinnen: I don't think the comprehension is needed, a generator should work (just remove the `[ ]`)

work_on_type_system/src/connections.jl
```diff
@@ -0,0 +1,194 @@
...
+    return TransferFunction(mat, Ts)
+end
+
+append(systems::LTISystem...) = append(promote(systems...)...)
+
+function Base.vcat(systems::StateSpace...)
+    # Perform checks
+    nu = systems[1].nu
+    if !all([s.nu == nu for s in systems])
+        error("All systems must have same input dimension")
```
baggepinnen: This is done in many places in the code, maybe have a function that checks this?

work_on_type_system/src/connections.jl
```diff
@@ -0,0 +1,194 @@
...
+    end
+    mat = blkdiag([s.matrix for s in systems]...)
+    return TransferFunction(mat, Ts)
+end
+
+append(systems::LTISystem...) = append(promote(systems...)...)
+
+function Base.vcat(systems::StateSpace...)
+    # Perform checks
+    nu = systems[1].nu
```
baggepinnen: Field access, expose API?

work_on_type_system/src/connections.jl
```diff
@@ -0,0 +1,194 @@
...
+    if !all([s.nu == nu for s in systems])
+        error("All systems must have same input dimension")
+    end
+    Ts = systems[1].Ts
+    if !all([s.Ts == Ts for s in systems])
+        error("Sampling time mismatch")
+    end
+    A = blkdiag([s.A for s in systems]...)
```
baggepinnen: These 4 lines of creating ABCD are repeated in many places in the code

work_on_type_system/src/connections.jl
```diff
@@ -0,0 +1,194 @@
...
+
+function Base.vcat(systems::TransferFunction...)
+    # Perform checks
+    nu = systems[1].nu
+    if !all([s.nu == nu for s in systems])
+        error("All systems must have same input dimension")
+    end
+    Ts = systems[1].Ts
+    if !all([s.Ts == Ts for s in systems])
+        error("Sampling time mismatch")
```
baggepinnen: Also repeated code that could be a function

work_on_type_system/src/connections.jl
```diff
@@ -0,0 +1,194 @@
...
+#     if S <: TransferFunction
+#         hcat(map(e->convert(TransferFunction,e),systems)...)
+#     else
+#         cat(2,systems...)
+#     end
+# end
+
+# TODO: could use cat([1,2], mats...) instead
+# Empty definition to get rid of warning
+Base.blkdiag() = []
```
baggepinnen: Probably not advisable to define this

baggepinnen: Below is defining method for base function on base types -> can change behavior in other packages after loading ControlSystems

work_on_type_system/src/connections.jl
```diff
@@ -0,0 +1,194 @@
...
+function Base.blkdiag(mats::Matrix...)
+    rows = Int[size(m, 1) for m in mats]
+    cols = Int[size(m, 2) for m in mats]
+    T = eltype(mats[1])
+    for ind=1:length(mats)
+        T = promote_type(T, eltype(mats[ind]))
```
baggepinnen: This looks like some kind of reduce operation?

work_on_type_system/src/types/StateSpace.jl
```diff
@@ -0,0 +1,263 @@
...
+            to_matrix(T, D), Float64(Ts))
+end
+
+# Getter functions
+get_A(sys::StateSpace) = sys.A
+get_B(sys::StateSpace) = sys.B
+get_C(sys::StateSpace) = sys.C
+get_D(sys::StateSpace) = sys.D
+
+get_Ts(sys::StateSpace) = sys.Ts
```
baggepinnen: slightly non-julian to name functions like this, more typical would be `sampletime(sys)`

work_on_type_system/src/connections.jl
olof3: Yes, do we actually need an append function? I don't think the behavior of this `append` function is obvious from its name. There is also an extension/modification of `blkdiag` in the existing code which acts on matrices.

Perhaps remove append and just have one function
`function Base.blkdiag(systems::Union{LTISystem,Number,AbstractMatrix{<:Number}}...)`
?

work_on_type_system/src/connections.jl
olof3: So we have discussed some more elaborate ways to implement different sampling times rather than just have a field of Float64 in both the Discrete and Continuous systems. This switch would probably require getter functions, so adding them would facilitate a potential switch.

However otherwise I guess it makes the code a bit more verbose, also if we want to have getter functions for the matrices in statespace systems. I wouldn't mind keeping the direct field access for the moment, at least for the statespace matrices. But, I guess it's good practice with getter functions.

In terms of input output dimensions, I agree with mfalt to remove the field nu, ny, nx and just infer the dimensions from the stored matrices. Does it seem reasonable to call the getter functions `nstates`, `noutputs`, `ninputs` in this case?

work_on_type_system/src/connections.jl
olof3: See above, probably remove the field `nu` altogether and use `noutputs`?

work_on_type_system/src/connections.jl
baggepinnen: A single function for accessing all matrices might suffice, `A,B,C,D = matrices(sys)` They are not copied so it's free to get them all anyway. 
`nstates, noutputs, ninputs ` sound reasonable to me

work_on_type_system/src/connections.jl
baggepinnen: e.g.
```julia
validate_sampletimes
check_sampletimes
```
It's an internal function so any self explaining name would be okay

src/types/zpk.jl
```diff
@@ -45,6 +45,24 @@ function zpk(gain::Matrix{T}, Ts::Real=0; kwargs...) where {T <: Number}
...
+    elseif eltype(z) == Any
+        @assert z == []
+        TR = eltype(p)
+        return zpk(TR[], p, k)
+    elseif eltype(p) == Any
+        @assert p == []
+        TR = eltype(z)
+        return zpk(z, TR[], k)
+    else
+        error("Non numeric vectors must be empty.")
```
mfalt: Are we sure this is a good restriction?

PR #134, opened 2018-06-19
Update REQUIRE

ufechner7:
Don't require an outdated version of Plots.jl

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.6.0
-Plots 0.7.4
+Plots
 Polynomials
 LaTeXStrings
 Requires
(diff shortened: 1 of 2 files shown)


baggepinnen:
We don't require an old version, from the julia documentation

> For example, the line:
> 
> Distributions 0.1
> is satisfied by any version of Distributions greater than or equal to 0.1.0

We thus require versions *newer* than v0.7.4 which we probably still want to do.

ufechner7:
Well, when I installed the master branch of ControlSystems.jl, packages like DataFrames got downgraded. This didn't happen when I installed my modified branch:
`
Pkg.rm("ControlSystems")
Pkg.clone("https://github.com/ufechner7/ControlSystems.jl")
Pkg.checkout("ControlSystems.jl", "stable")
`
Perhaps the Julia documentation does not describe the real behaviour of the package manager?

baggepinnen:
In going to close this as we don't want to remove the lower bound on Plots.jl

PR #135, opened 2018-06-20
Fixed error in docstring, improved example in docstring of LQG.

ufechner7:
Parameters in the docstring were swapped. The example can now be copy-pasted and it works.


ufechner7:
I added a more detailed example in the example folder. It also shows how to convert a scalar in a matrix before passing it to the LQR constructor, when needed.

PR #139, opened 2018-06-22
Hvcat

mfalt:
Removes the h(v)cat functions that overloads Base and creates new macro for this
`@sys [1/s tf(1); ones(1,2)]`
instead of `[1/s tf(1); ones(1,2)]` 

Comments are very welcome. (Some small changes are not relevant to this branch)


src/connections.jl
```diff
@@ -43,7 +43,58 @@ end
...
+            return esc(Expr(:call, vcatsys, rows...))
+        #else
+        #    #vcat
+        #    Expr(:call, vcat, ex.args...)
+        #end
+    elseif ex.head == :hcat
+        # hcat only
+        return esc(hcatsysex(ex))
+    else
+        :(error("Undexpected argument to @sys, expected `vcat` or `hcat`"))
```
baggepinnen: This can error at compile time, i.e remove the expression.

mfalt:
It is mostly to avoid overloading `hcat`,`vcat` and `hvcat`, since they are also catching calls without any of our types. I think even the call `[1 2]` might cause type-instability without this.

Adding `Array{TransferFunction}` to `TransferFunction` conversion should be trivial, but `@sys` is slighty more powerful when the size of the matrix doesn't agree with the size of the systems, e.g
```julia
f1 # 1x1 system
f2 # 1x2 system
f3 # 1x3 system
@sys [f1 f2; f3]
```

src/connections.jl
mfalt: is `esc(Expr(:call, error, "Text"))` the right approach?

baggepinnen:
You can just error as if it wasn't in a macro at all.
error("text")

olof3:
As I've understood it this pull request addresses two issues:

1. Avoiding that ControlSystem's `hcat` and `vcat` catches calls like `[randn(2,2), randn(2,2)]`
2. Allowing the creation of matrices of systems, without creating a single MIMO system

I think that it should be possible to address this without introducing a macro, keeping the original syntax. To me this makes this also makes the most sense, see 2. below.

1. I think that it should be possible to avoid this behavior, while properly concatenating systems. I believe that I got it to work while overloading `cat_t`. Or would this be disallowed in future Julia versions?

2. I think this situation is completely analogous to how one wants `[randn(2,2) randn(2,2); randn(2,2) randn(2,2)]` to behave, i.e. return a 4x4 matrix. Following this `hcat` and `vcat` on `LTISystems` should return a MIMO system. While there are many instances where one wants to create matrices of systems, it seems to me that it should be done in an analogous way to creating matrices of Matrix{<:Number}.

mfalt:
Thanks to @olof3 this works without a macro now.

PR #141, opened 2018-06-26
Mat2ss conversion

olof3:
Conversion from Matrix to StateSpace.

PR #142, opened 2018-06-26
Conversion Matrix{Number} -> Statespace + tests

mfalt:
Duplicate of https://github.com/JuliaControl/ControlSystems.jl/pull/141

### DIFF
--- a/src/types/conversion.jl
+++ b/src/types/conversion.jl
@@ -18,29 +18,22 @@
 # Base.convert(::Type{<:TransferFunction}, b::Number) = tf([b])
 # Base.convert(::Type{<:TransferFunction{<:SisoRational}}, b::Number) = tf(b)
 # Base.convert(::Type{<:TransferFunction{<:SisoZpk}}, b::Number) = zpk(b)
-Base.convert(::Type{<:TransferFunction{<:SisoRational{T}}}, b::Number) where {T<:Number} = tf(T(b))
-Base.convert(::Type{<:TransferFunction{<:SisoZpk{T, TR}}}, b::Number) where {T<:Number, TR} = zpk(T(b))
-function Base.convert(::Type{StateSpace{T, MT}}, b::Number) where {T<:Number, MT}
-    A = convert(MT, Matrix{T}(0,0))
-    B = convert(MT, Matrix{T}(0,1))
-    C = convert(MT, Matrix{T}(1,0))
-    D = convert(MT, fill(convert(T, b), (1,1)))
-    Ts = 0.0
-    return StateSpace{T,MT}(A,B,C,D,Ts)
-end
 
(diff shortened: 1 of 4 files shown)


src/types/conversion.jl
```diff
@@ -20,20 +20,29 @@
...
 Base.convert(::Type{<:TransferFunction{<:SisoRational{T}}}, b::Number) where {T<:Number} = tf(T(b))
 Base.convert(::Type{<:TransferFunction{<:SisoZpk{T, TR}}}, b::Number) where {T<:Number, TR} = zpk(T(b))
+function Base.convert(::Type{StateSpace{T, MT}}, b::Number) where {T<:Number, MT}
+    A = convert(MT, Matrix{T}(0,0))
+    B = convert(MT, Matrix{T}(0,1))
+    C = convert(MT, Matrix{T}(1,0))
+    D = convert(MT, fill(convert(T, b), (1,1)))
+    Ts = 0.0
+    return StateSpace{T,MT}(A,B,C,D,Ts)
+end
```
mfalt: To avoid duplication of code, this could just be:
```julia
Base.convert(::Type{StateSpace{T, MT}}, b::Number) where {T<:Number, MT} = convert(StateSpace{T, MT}, fill(b, (1,1)))
```

mfalt: I don't think it gets optimized away, but a call to `size(D)` should be negligible.

olof3: I think the code would be much more compact, perhaps even allowing some duplication if we could have  something like

```julia
StateSpace{T,MT}(MT(0,0), MT(0,1), MT(1,0), convert(MT, fill(convert(T, b), (1,1))), Ts)
```

however this does not seem to work for, e.g., `MT = SparseMatrixCSC{Float64,Int64}`

mfalt: Yes, this is precisely the reason why I did `convert(MT, fill(`, there is no canonical way to create a general AbstractMatrix of a certain size

src/types/conversion.jl
```diff
@@ -20,20 +20,29 @@
...
-function Base.convert(::Type{StateSpace{T1, MT}}, b::T2) where {T1, MT, T2<:Number}
-    A = convert(MT, fill(zero(T1), (0,0)))
-    B = convert(MT, fill(zero(T1), (0,1)))
-    C = convert(MT, fill(zero(T1), (1,0)))
-    D = convert(MT, fill(convert(T1, b), (1,1)))
+function convert(::Type{StateSpace{T,MT}}, D::AbstractMatrix{<:Number}) where {T, MT}
+    (ny, nu) = size(D)
+    A = convert(MT, Matrix{T}(0,0))
+    B = convert(MT, Matrix{T}(0,nu))
+    C = convert(MT, Matrix{T}(ny,0))
```
mfalt: Should be noted that the previous version here was used since the one you use is deprecated in julia 0.7 to `Matrix{T}(undef, 0,nu))`
not sure if
`fill(zero(T1), (0,1)))`
looks better though

PR #143, opened 2018-06-26
eltype and numeric_type update

mfalt:
Any comments?

### DIFF
--- a/src/types/conversion.jl
+++ b/src/types/conversion.jl
@@ -62,15 +62,14 @@ function convert(::Type{StateSpace{T,MT}}, sys::StateSpace) where {T, MT}
 end
 
 
-function Base.convert(::Type{<:StateSpace}, G::TransferFunction)
+function Base.convert(::Type{<:StateSpace}, G::TransferFunction{<:SisoTf{T0}}) where {T0<:Number}
     if !isproper(G)
         error("System is improper, a state-space representation is impossible")
     end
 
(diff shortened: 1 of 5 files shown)


olof3:
Looks alright to me.

src/types/conversion.jl
```diff
@@ -103,9 +102,8 @@ end
 siso_tf_to_ss(f::SisoTf) = siso_tf_to_ss(convert(SisoRational, f))
 
 # Conversion to statespace on controllable canonical form
-function siso_tf_to_ss(f::SisoRational)
-    T0 = numeric_type(f)
-    T = typeof(one(T0)/one(T0))
+function siso_tf_to_ss(f::SisoRational{T0}) where {T0<:Number}
```
baggepinnen: Shouldn't this be either
- convert
- Constructor
?

olof3: I guess it's some kind of helper function. It doesn't return a StateSapce but just matrices A, B, C, D. Perhaps an underscore before the function name would make it a bit more clear.

olof3: in general it would be great with a bit more clarity for the conversions between the different system types. Perhaps split into different files? Perhaps move the balanced realization code?

mfalt: Agree, there are many things that could improve here, but that seems like part of a bigger work on the constructors in general.

PR #144, opened 2018-06-26
Misc. fixes.

olof3:


src/freqresp.jl
```diff
@@ -39,10 +39,11 @@ function _preprocess_for_freqresp(sys::StateSpace)
     if isempty(sys.A) # hessfact does not work for empty matrices
         return sys
     end
+    TT = promote_type(numeric_type(sys), Float64)
 
     A, B, C, D = sys.A, sys.B, sys.C, sys.D
     F = hessfact(A)
-    H = F[:H]::Matrix{Float64}
+    H = F[:H]::Matrix{TT}
```
mfalt: Is this always what hessfact givs in base?

olof3: Seems like old code. I changed it to allow for complex matrices, but perhaps we should just drop it altogether?

PR #148, opened 2018-08-31
Added precision argument to print and show functions

jacobhuesman:
I added a precision argument to `print` and `show` to allow for limiting the number of digits displayed by the two functions. Previously for coefficients that didn't come out to be nice numbers, the default number of digits made the output annoying interpret. 

Here's an example:
```
using ControlSystems
A = [-20 -2 0 0; 10 0 -10 0; 0 5 -100 -5; 0 0 4 0]
B = [2 0 0 0]'
C = [0 1 0 0]
D = 0
G = ss(A,B,C,D)
```
Outputting the transfer function to the terminal looked like this:
```
# Input
show(zpk(G))

# Output
TransferFunction:
                                 (s + 99.79959839197049)(s + 1.407374883553052e15)(s + 0.20040160804507176)
1.4210854715202004e-14------------------------------------------------------------------------------------------------
                      (s + 99.29631508731481)(s + 1.6701219833671543)(s + 18.905983873522935)(s + 0.12757905579519613)

```
I added some formatting to make the display a little cleaner:
```
# Input
show(zpk(G))

# Output
TransferFunction:
               (s + 99.7996)(s + 1.40737e+15)(s + 0.200402)
1.42109e-14----------------------------------------------------
           (s + 99.2963)(s + 1.67012)(s + 18.906)(s + 0.127579)

```
And adjustable:
```
# Input
show(zpk(G), 2)

# Output
TransferFunction:
         (s + 100)(s + 1.4e+15)(s + 0.2)
1.4e-14-----------------------------------
       (s + 99)(s + 1.7)(s + 19)(s + 0.13)

```

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,3 +3,4 @@ Plots 0.7.4
 Polynomials
 LaTeXStrings
 Requires
+Formatting
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -32,3 +32,4 @@ pages:
     - Synthesis: 'lib/synthesis.md'
     - 'Time and Frequency response': 'lib/timefreqresponse.md'
     - Plotting: 'lib/plotting.md'
+    - Display: 'lib/display.md'
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -20,12 +20,12 @@ Depth = 1
 ## Functions
 
 ```@contents
-Pages = ["lib/constructors.md", "lib/plotting.md"]
+Pages = ["lib/constructors.md", "lib/plotting.md", "lib/display.md"]
 ```
 
 ## Documentation Index
 
 ```@index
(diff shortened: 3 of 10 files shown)


mfalt:
Thanks for the great work, it looks a lot better.
Unfortunately, we just pulled a set of major changes into the package (this beast https://github.com/JuliaControl/ControlSystems.jl/pull/118) so I'll have to take care of the conflicts, probably be rewriting most of it.

One problem is that we switched from our own implementation of polynomials to Polynomials.jl, and partly use that system for printing now.
The new syntax is from Formatting.jl right? How much of a standard is this, are other packages using it?

jacobhuesman:
Oh boy haha... I saw that branch when I was starting on this but figured it would be a while before it was ready to be merged in. 

I can look into refactoring this later in the week if you think that this functionality is worthwhile to your package. I just picked Formatting.jl since it was readily available.

The `show` and `print` syntax isn't standard as far as I know. I was originally going to use keyword arguments like
```
print(tf(G), precision=2)
```
but that doesn't work well with the `sprint` function that you guys were using.

PR #149, opened 2018-09-01
Travis fix

jacobhuesman:
Fixed a few of the issues with Travis-CI:
- Added some dependencies that Plot.jl doesn't seem to pull in automatically, but are still necessary for GTK to build correctly.
  - texlive-latex-base
  - dvipng 
  - libqt5widgets5 
  - libqt5network5
- Changed the travis container to `sudo: required` as some of the installs require sudo
- Switched to a forked branch of [ControlExamplePlots.jl](https://github.com/jacobhuesman/ControlExamplePlots.jl/tree/fix_and_update) that I've implemented some fixes and updates on.
- Marked the gangoffour test as broken. I'm not sure if it's local to my machine, but the plot does not display cleanly at all for me.

Master should build successfully once this is merged in. Although it might make more sense to wait and update this pull request once this https://github.com/JuliaControl/ControlExamplePlots.jl/pull/1 is merged in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ script:
     - julia -e 'Pkg.clone(pwd()); Pkg.build("ControlSystems")'
     #- julia -e 'ENV["PYTHON"] = ""; Pkg.clone("PyPlot"); Pkg.build("PyPlot")'
     - julia -e 'Pkg.clone("https://github.com/jacobhuesman/ControlExamplePlots.jl.git")'
-    - julia -e 'Pkg.checkout("ControlExamplePlots", "revert_lsimgen")'
+    - julia -e 'Pkg.checkout("ControlExamplePlots", "fix_and_update")'
     - julia -e 'Pkg.test("ControlSystems"; coverage=true)'
 
 after_success:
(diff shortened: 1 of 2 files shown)


mfalt:
Thanks, I'll take a look and see which of these are still needed after I bump the julia version to 0.7.

We have currently disabled the plots tests since they were a major pain, and will cause the tests to fail, even as improvements are made. We should probably enable them soon again though.

It seems like the tolerances are sometimes increased an sometimes decreased, we should make sure that they work on different systems, so it is good to have some margins.

Once we stabilize the package and activate plot tests, we should update the figures in ControlExamplePlots to reflect the desired behavior.

jacobhuesman:
Sounds good to me, I do think the plot tests add a lot of complexity to the test suite. 

I would still recommend adding `sudo: required` to your `.travis.yml` file. Otherwise the system builds in a [sudo-less](https://docs.travis-ci.com/user/reference/overview/) environment that might cause build failures if any of your packages dependencies need to pull in debian packages like `Plots.jl` does.

baggepinnen:
We have now moved CI to GH actions. Thanks for your attempt to help with the old CI infrastructure!

PR #161, opened 2018-10-05
Made sure that plot! works for all recipes

mfalt:


mfalt:
Updated it to look slightly better

Issue #163, opened 2018-10-23
feedback(sys) not working correctly if sys is in transfer function representation


bernhard-thiele:
Consider following code snippet which demonstrates the problem:

```
julia> using ControlSystems

julia> g = tf([1],[1,1])
ControlSystems.TransferFunction{ControlSystems.SisoRational{Int64}}
  1
-----
s + 1

Continuous-time transfer function model

julia> feedback(g)
ControlSystems.TransferFunction{ControlSystems.SisoRational{Int64}}
   1
-------
2*s + 1

Continuous-time transfer function model

julia> tf(feedback(ss(g)))
ControlSystems.TransferFunction{ControlSystems.SisoRational{Float64}}
    1.0
-----------
1.0*s + 2.0

Continuous-time transfer function model
```

mfalt:
Thanks for finding this, we corrected it and made some changes to avoid these kinds of errors in https://github.com/JuliaControl/ControlSystems.jl/pull/164

PR #164, opened 2018-10-31
Fixed feedback and added Polynomial constructors for simpler code

mfalt:


mfalt:
Solves https://github.com/JuliaControl/ControlSystems.jl/issues/163

src/types/tf.jl
```diff
@@ -69,3 +69,23 @@ function tf(var::AbstractString, Ts::Real)
     Ts == 0 && error("Ts must not be 0 for discrete time tf.")
     return tf([1, 0], [1], Ts)
 end
+
+## Constructors for polynomial inputs
+function tf(num::AbstractArray{PT}, den::AbstractArray{PT},  Ts::Real=0.0) where {T<:Number, PT <: Polynomials.Poly{T}}
+    ny, nu = size(num, 1), size(num, 2)
+    if (ny, nu) != (size(den, 1), size(den, 2))
```
baggepinnen: `(size(den, 1), size(den, 2)) == size(den) `?

src/types/tf.jl
```diff
@@ -69,3 +69,23 @@ function tf(var::AbstractString, Ts::Real)
...
+        error("num and den dimensions must match")
+    end
+
+    matrix = Matrix{SisoRational{T}}(undef, ny, nu)
+    for o=1:ny
```
baggepinnen: This double loop can be replaced by one over eachindex

src/types/tf.jl
mfalt: Not if den is vector

src/types/tf.jl
mfalt: That doesn't have to result in a `Matrix` type

PR #165, opened 2018-11-13
Mimo poles

albheim:
Implementation of MIMO poles according to Glad and Ljung book. Implementation can definitely be improved, but seems to be working at least.


mfalt:
Do you know what failed and why? It would be nice if we could pull this.

src/analysis.jl
```diff
@@ -2,10 +2,89 @@
...
+        end
+    end
+    return minors
+end
+
+"""`det(sys)`
+
+Compute the determinant of the Matrix `sys` of SisoTf systems, returns a SisoTf system."""
+# TODO: improve this implementation, should be more efficient ones
+function det(sys::Matrix{SisoZpk{T, TR}}) where {T<:Number, TR}
```
mfalt: Should be `Matrix{S} ... where {T,TR,S<:SisoZpk}` since S is used later.

test/test_analysis.jl
```diff
@@ -177,6 +179,15 @@ z, p, k = zpkdata(G)
      "|  -2.000e+00   |  1.000e+00    |  2.000e+00    |  5.000e-01    |\n"*
      "|  3.000e+00    |  -1.000e+00   |  3.000e+00    |  -3.333e-01   |\n")
 
+s = tf("s")
```
mfalt: This line can be removed

mfalt:
Should be ready to pull if no one has any comments.

baggepinnen:
Ready to pull?

PR #167, opened 2018-12-09
Update synthesis.jl

pilwonhur:
In Line 165
tf(P,P+Q),L.Ts) needs to be changed to tf(P[1][end:-1:0],(P+Q)[1][end:-1:0],L.Ts)
Note that Polynomials use 0 index in reverse order.

In Line 140,
q += A^(n-i)*poly[iL] -> q += A^(n-i)*poly[i]
Polynomials use 0-index.

### DIFF
--- a/src/synthesis.jl
+++ b/src/synthesis.jl
@@ -137,7 +137,7 @@ function acker(A,B,P)
     poly = mapreduce(p -> Poly([1, -p]), *, P, init=Poly(one(eltype(P))))
     q = zero(Array{promote_type(eltype(A),Float64),2}(undef, n,n))
     for i = n:-1:0
-        q += A^(n-i)*poly[i+1]
+        q += A^(n-i)*poly[i]
     end
     S = Array{promote_type(eltype(A),eltype(B),Float64),2}(undef, n,n)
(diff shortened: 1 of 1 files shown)


mfalt:
I will look into this as soon as I can, we have had several errors like this, nice catch.

Could you add a test for this that currently fails and is fixed with your update?

mfalt:
I verified that the acker code indeed has a bug and your code seems to fix it, if you could add som simple tests like
```julia
A = randn(3,3)
B = randn(3,1)
p = [3.0,2,1]
K = ControlSystems.acker(A,B,p)
@test K ≈ "some data"
@test eigvals(A-B*K)  ≈ p
```
with some deterministic data, that would be great.

I think the feedback was fixed by the commit https://github.com/JuliaControl/ControlSystems.jl/commit/3c36152ff9e3539bff5b8e245aed0ff4c28db485 by making sure that we use the polynomials directly in the consructor, exacty to avoid this indexing problem. Once the acker bug is fixed, I'll make sure that we tag a new version of ControlSystems.

baggepinnen:
Bump, this bug fix would be nice to have asap

PR #168, opened 2018-12-11
Enable freqresp for LTIsystems with special types

baggepinnen:
With this patch, `bode` and `sigma` works for stuff like
```julia
using Measurements
ζ = 0.8 ± 0.1
G = tf(1, [1., 2ζ, 1])
sigma(G)
```

### DIFF
--- a/src/freqresp.jl
+++ b/src/freqresp.jl
@@ -149,7 +149,7 @@ frequencies `w`
 function sigma(sys::LTISystem, w::AbstractVector)
     resp = freqresp(sys, w)
     nw, ny, nu = size(resp)
-    sv = Array{Float64}(undef, nw, min(ny, nu))
+    sv = Array{numeric_type(sys)}(undef, nw, min(ny, nu))
     for i=1:nw
         sv[i, :] = svdvals(resp[i, :, :])
     end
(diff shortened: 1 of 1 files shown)


mfalt:
I think we need to keep these vectors Float-ish, for example with integer systems. (This is where the tests seem to fail) `svdvals` autoatically converts to Float32/Float64/Complex in the standard cases. We have had so  much problems with promotions and pre-allocation of matrices, so I think it's time so sacrifice some (potential) speed and efficiency for robustness where possible, i.e. in this case, we should not try to do some `promote_op(/, numeric_type(..),...)`, but instead
```julia
sv = dropdims(mapslices(svdvals, resp, dims=(2,3)),dims=(3))
```
and for the second case
```julia
bounds = map(sys -> ControlSystems._bounds_and_features(sys, plot)[1], systems)
w1 = minimum(minimum.(bounds))
w2 = maximum(maximum.(bounds))
```
The would however propagate the ridiculusly complete failiure on type inference in `_bounds_and_features`

baggepinnen:
Am on it, will change some others at the same time

mfalt:
Looks good, thanks!

PR #170, opened 2018-12-19
Some type instability issues

mfalt:


mfalt:
Did you manually test the plots before pulling this? I think I just pushed a bugfix to this branch

PR #173, opened 2019-01-16
Update DARE

HppyCtrlEngnrng:
Remove the assumption of invertibility of A matrix utilizing generalized Schur decomposition.


baggepinnen:
Looks good to me, @mfalt?

src/matrix_comps.jl
```diff
@@ -34,37 +34,34 @@ end
...
 Algorithm taken from:
 Laub, "A Schur Method for Solving Algebraic Riccati Equations."
 http://dspace.mit.edu/bitstream/handle/1721.1/1301/R-0859-05666488.pdf
 """
 function dare(A, B, Q, R)
-    G = try
-        B*inv(R)*B'
-    catch
-        error("R must be non-singular.")
+    if (minimum(eigvals(Q)) < 0)
```
mfalt: What happens when eigvals are complex here?

olof3:
Great! However two test cases where the new code breaks and the old code worked are
```
I2 = Matrix{Float64}(I, 2, 2)
dare([1.0 im; im 1.0], I2, I2, I2) ≈ [1 + sqrt(2) 0; 0 1 + sqrt(2)]
```
```
dare(1.0, 1, 1, 1) ≈ fill((1 + sqrt(5))/2, 1, 1)
```


The first test case is fixed by changing `transpose(A)` to `A'` and  `transpose(B)` to `B'`. (Laub's paper only considered the real case, but the complex case is roughly the same, see Lancaster's book on Algebraic Riccatti equations).

The second test case still breaks since somehow the new code doesn't handle scalar inputs due to that 
`[1.0I 1; 0.0 1.0]` gives a matrix of type `Any`!? Somehow this was not an issue in the old code. It can be discussed if it is necessary to support scalar inputs.


Two other test cases which would be good to include are
```
dare(1.0im, 1, 1, 1) ≈ fill((1 + sqrt(5))/2, 1, 1)
dare(1.0, 1im, 1, 1) ≈ fill((1 + sqrt(5))/2, 1, 1)
```

HppyCtrlEngnrng:
Dear All, Thank you very much for your reviews!
As you pointed out, I focused only on real case for discrete time LQR.
Now, I made some updates and it supports both complex and scalar A, B, Q, R.

Also, I wrote a few test codes.
Could you please check them out?
[dare_update_tests.zip](https://github.com/JuliaControl/ControlSystems.jl/files/2822013/dare_update_tests.zip)

src/matrix_comps.jl
```diff
@@ -34,37 +34,35 @@ end
 """`dare(A, B, Q, R)`
 
 Compute `X`, the solution to the discrete-time algebraic Riccati equation,
-defined as A'XA - X - (A'XB)(B'XB + R)^-1(B'XA) + Q = 0, where A and R
-are non-singular.
+defined as A'XA - X - (A'XB)(B'XB + R)^-1(B'XA) + Q = 0, where Q>=0 and R>0
 
 Algorithm taken from:
```
baggepinnen: Does this need updating now?

src/matrix_comps.jl
```diff
@@ -34,37 +34,35 @@ end
...
-
-    (m, n) = size(U)
-    U11 = U[1:div(m, 2), 1:div(n,2)]
-    U21 = U[div(m,2)+1:m, 1:div(n,2)]
-    return U21/U11
+    
+    n = size(A, 1);
+    
+    E = [
+        Matrix{Float64}(I, n, n) B/R*B';
```
baggepinnen:
```suggestion
        I B/R*B';
```
shouldn't it be enough to write `I`?

src/matrix_comps.jl
```diff
@@ -34,37 +34,35 @@ end
...
+        zeros(size(A)) A'
+    ];
+    F = [
+        A zeros(size(A));
+        -Q Matrix{Float64}(I, n, n)
```
baggepinnen:
```suggestion
        -Q I
```
same here?

src/matrix_comps.jl
mfalt: There was a bug in julia where this didn't work for 0 size of I. We could test it against 1.0.x to see if the proposed change works, but I don't see a problem in leaving it like this.

mfalt:
Thanks for fixing this @HppyCtrlEngnrng

PR #175, opened 2019-01-24
Smallfixes

baggepinnen:


src/matrix_comps.jl
```diff
@@ -389,11 +389,11 @@ function normLinf_twoSteps_dt(sys::StateSpace,tol=1e-6, maxIters=1000, approxcir
 end
 
 
-"""`T, B = balance(A[, perm=true])`
+"""`S, P, B = balance(A[, perm=true])`
```
mfalt: This seems to not agree with the output, but maybe it didnt before either?

baggepinnen: I'm not sure what you mean? The current output is the three matrices s, p, b

mfalt: Ah, I looked at the wrong output, github hid too much.

PR #176, opened 2019-01-24
add simple sys id

baggepinnen:
This PR adds simple arx tf estimation through equation error estimation. Useful for quick SISO and MISO estimation of transfer functions from data


baggepinnen:
Will just put this in [ControlSystemIdentification.jl](https://github.com/baggepinnen/ControlSystemIdentification.jl) instead

PR #177, opened 2019-01-24
Minor bug in plot

mfalt:


baggepinnen:
Ready to pull?

mfalt:
@baggepinnen You already pushed this commit right?

PR #178, opened 2019-01-24
WIP: Delayed LTI

mfalt:
Do not pull yet, still WIP

List of problems, feel free to add to it:

Should be resolved: 
 - [ ] More extensive tests needed
 - [ ] Some discussions on the code below

Would be nice to have:
 - [ ] lsim doesn't handle D22 term
 - [x] lsim interpolates x to get y (with D21 != 0), which is the only inhibitor for great accuracy (it is still very much usable, and x is correct at the grid points) 
 - [ ] impulse: This is quite tricky, the first impulse can be handled, but if there is a direct term D21 to the delay, then we have to some way "store" the impulse train in there. I am not even sure how to begin to think about D22 != 0
 - [ ] Some sort of approximation of the delays would be useful to implement, e.g. for use with the design tools


src/types/DelayLtiSystem.jl
```diff
@@ -0,0 +1,95 @@
...
+    DelayLtiSystem{T}(psys, Tau)
+end
+
+DelayLtiSystem(sys::StateSpace{T,MT}, Tau::Vector{Float64}) where {T, MT} = DelayLtiSystem{T}(sys, Tau)
+DelayLtiSystem(sys::StateSpace{T,MT}) where {T, MT} = DelayLtiSystem{T}(sys, Float64[])
+DelayLtiSystem{T}(sys::StateSpace) where T = DelayLtiSystem{T}(sys, Float64[])
+
+
+# TODO: Think through these promotions and conversions
+Base.promote_rule(::Type{<:StateSpace{T1}}, ::Type{DelayLtiSystem{T2}}) where {T1<:Number,T2<:Number} = DelayLtiSystem{promote_type(T1,T2)}
```
mfalt: Should problably be caught by a `Type{<:LTISystem, DelayLTISystem}` to work with TF also.

mfalt: But we would need `LTISystem{T}`` which we don't, applies to the rest of the comments too.

src/types/DelayLtiSystem.jl
```diff
@@ -0,0 +1,95 @@
...
+Base.promote_rule(::Type{AbstractMatrix{T1}}, ::Type{DelayLtiSystem{T2}}) where {T1<:Number,T2<:Number} = DelayLtiSystem{promote_type(T1,T2)}
+Base.promote_rule(::Type{T1}, ::Type{DelayLtiSystem{T2}}) where {T1<:Number,T2<:Number} = DelayLtiSystem{promote_type(T1,T2)}
```
mfalt: These should probably be caught somewhere else by
`Type{AbstractMatrix}, ::Type{<:LTISystem}`
and
`Type{<:Number}, ::Type{<:LTISystem}`

src/types/DelayLtiSystem.jl
```diff
@@ -0,0 +1,95 @@
...
+
+
+function feedback(sys1::DelayLtiSystem, sys2::DelayLtiSystem)
+    psys_new = feedback(sys1.P, sys2.P)
+    Tau_new = [sys1.Tau; sys2.Tau]
+
+    DelayLtiSystem(psys_new.P, Tau_new)
+end
+
+function delay(tau::Real, T::Type{<:Number}=Float64)
```
mfalt: Is this the canonical way, or maybe we should ahve:
`delay(tau::Real)`
and
`delay(::Type{T}, tau::Real) whete T<:Number`

src/types/Lti.jl
```diff
@@ -4,6 +4,9 @@ abstract type LTISystem <: AbstractSystem end
 *(sys1::LTISystem, sys2::LTISystem) = *(promote(sys1, sys2)...)
 /(sys1::LTISystem, sys2::LTISystem) = /(promote(sys1, sys2)...)
 
+feedback(sys1::Union{LTISystem,Number,AbstractMatrix{<:Number}},
+         sys2::Union{LTISystem,Number,AbstractMatrix{<:Number}}) = feedback(promote(sys1, sys2)...)
+
```
mfalt: We should think about being consistent here, but that discussion probably belongs elsewehere.
For `/(sys1::LTISystem, sys2::LTISystem) = /(promote(sys1, sys2)...)`
we dont allow matrices and scalars (and assume the specific implementation will solve it), but for
`feedback` we provide a fallback here.

src/types/PartionedStateSpace.jl
```diff
@@ -0,0 +1,156 @@
...
+    elseif d == :D22
+        return P.D[ny1+1:end, nu1+1:end]
+    else
+        return getfield(P, d)
+    end
+end
+
+
+# There should already exist a function like this somewhare?
+function blkdiag(A1::Matrix{T1}, A2::Matrix{T2}) where {T1<:Number, T2<:Number}
```
mfalt: `AbstractMatrix` maybe for future-proofing

mfalt: Although we would have to figure out a way of constructing the new matrix to be the promotetype, which might be tricky. So maybe we skip that

src/types/DelayLtiSystem.jl
olof3: Yes, any of those options are probably better.

src/types/Lti.jl
olof3: Yes, I think there are many places where more we should be more consistent.

src/types/DelayLtiSystem.jl
olof3: Is `delay` a reasonable name for the function? It seems quite appropriate, but another option would be `exp(-tau*tf("s")`. However this transfer function notation is a bit unfortunate since time delays and transfer functions tend to be numerically troublesome.

example/delayed_lti_system.jl
```diff
@@ -0,0 +1,19 @@
+using Plots
```
baggepinnen:
```suggestion
using Plots, ControlSystems
```

example/delayed_lti_timeresp.jl
```diff
@@ -0,0 +1,10 @@
+using Plots
```
baggepinnen:
```suggestion
using Plots, ControlSystems
```

example/delayed_lti_timeresp.jl
```diff
@@ -0,0 +1,10 @@
...
+plotly()
```
baggepinnen:
```suggestion
```
should maybe not change the users default

example/delayed_lti_timeresp.jl
```diff
@@ -0,0 +1,10 @@
...
+
+sys = feedback(1.0, ss(-1.0, 2, 1, 0) * (delay(2.0) + delay(3.0) + delay(2.5)))
+
+t = 0:0.02:8
+@time t, x, y = lsim(sys, t; u=t->[t>=0 ? 1.0 : 0.0])
```
baggepinnen:
```suggestion
t, x, y = lsim(sys, t; u=t->[t>=0 ? 1.0 : 0.0])
```
gues this was a leftover from your benchmarking?

src/ControlSystems.jl
```diff
@@ -75,9 +78,11 @@ export  LTISystem,
 
 
 # QUESTION: are these used? LaTeXStrings, Requires, IterTools
-using Polynomials, OrdinaryDiffEq, Plots, LaTeXStrings, LinearAlgebra
+using Polynomials, Plots, LaTeXStrings, LinearAlgebra
+using DifferentialEquations, OrdinaryDiffEq, DelayDiffEq
```
baggepinnen: Isn't `DifferentialEquations` an umbrella that exports everything else?

src/delay_systems.jl
```diff
@@ -0,0 +1,73 @@
+function freqresp(sys::DelayLtiSystem, ω::AbstractVector{T}) where {T <: Real}
+    ny = noutputs(sys)
+    nu = ninputs(sys)
+
+    P_fr = ControlSystems.freqresp(sys.P.P, ω);
```
baggepinnen:
```suggestion
    P_fr = freqresp(sys.P.P, ω);
```

src/delay_systems.jl
```diff
@@ -0,0 +1,73 @@
...
+    return G_fr
+end
+
+
+
+function lsim(sys::DelayLtiSystem, t::AbstractArray{<:Real}; u=(t -> fill(0.0, ninputs(sys))), x0=fill(0.0, nstates(sys)), alg=MethodOfSteps(Tsit5()), kwargs...)
+    P = sys.P
+
+    if ~iszero(P.D22)
+        error("non-zero D22-matrix block is not supported") # Due to limitations in differential equations
```
baggepinnen:
```suggestion
        throw(ArgumentError("non-zero D22-matrix block is not supported")) # Due to limitations in differential equations
```

src/delay_systems.jl
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    dt = t[2] - t[1]
+    if ~all(diff(t) .≈ dt) # QUESTION Does this work or are there precision problems?
+        error("The t-vector should be uniformly spaced, t[2] - t[1] = $dt.") # Perhaps dedicated function for checking this?
+    end
+
+    # Slightly more complicated definition since the d signal is not directly available
+    dde = function (dx, x, h, p, t)
```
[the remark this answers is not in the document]
baggepinnen: There really shouldn't be

src/delay_systems.jl
```diff
@@ -0,0 +1,73 @@
...
+    # Saves y (excluding the d(t-τ) contribution) and d
+    saved_values = SavedValues(Float64, Tuple{Vector{Float64}, Vector{Float64}})
+    cb = SavingCallback((x,t,integrator) -> (P.C1*x + P.D11*u(t), P.C2*x + P.D21*u(t)), saved_values, saveat=t)
+
+    prob = DDEProblem(dde, x0, h_initial, (0.0, 8.0), constant_lags=sys.Tau)
+
+    sol = solve(prob, alg, callback=cb; saveat=t, kwargs...)
+
+    x = sol.u # the states are labeled u in DifferentialEquations
+    y = hcat([saved_values.saveval[k][1] for k=1:length(t)]...)
```
baggepinnen:
```suggestion
    y = reduce(hcat, [saved_values.saveval[k][1] for k=1:length(t)])
```

src/delay_systems.jl
```diff
@@ -0,0 +1,73 @@
...
+    d = hcat([saved_values.saveval[k][2] for k=1:length(t)]...)
```
baggepinnen:
```suggestion
    d = reduce(hcat, [saved_values.saveval[k][2] for k=1:length(t)])
```
`reduce(hcat, ...)` is vastly more efficient and also do not compile new methods for varying length arguments.

src/delay_systems.jl
```diff
@@ -0,0 +1,73 @@
...
+
+    # Account for the effect of the delayed d-signal on y
+    for k=1:length(sys.Tau)
+        N_del = Integer(sys.Tau[k] / dt)
```
baggepinnen: `round(Int, x)` is more robust if `x` is not representable as integer

src/types/DelayLtiSystem.jl
```diff
@@ -0,0 +1,95 @@
...
+end
+
+# QUESTION: would psys be a good standard variable name for a PartionedStateSpace
+#           and perhaps dsys for a delayed system, (ambigous with discrete system though)
+function DelayLtiSystem{T}(sys::StateSpace, Tau::Vector{Float64}) where T
+    nu = ninputs(sys) - length(Tau)
+    ny = noutputs(sys) - length(Tau)
+
+    if nu < 0  || ny < 0
+        error("Time vector is too long.")
```
baggepinnen:
```suggestion
        throw(ArgumentError("Time vector is too long."))
```

src/types/PartionedStateSpace.jl
```diff
@@ -0,0 +1,156 @@
...
+    A_new = zeros(Float64, dims1 .+ dims2)
+    A_new[1:dims1[1], 1:dims1[2]] = A1
+    A_new[dims1[1]+1:end, dims1[2]+1:end] = A2
+
+    return A_new
+end
+
+
+function +(s1::PartionedStateSpace, s2::PartionedStateSpace)
+    A = blkdiag(s1.A, s2.A)
```
baggepinnen: Does `blkdiag` exist? Didn't see the definition, sorry if I missed it

src/delay_systems.jl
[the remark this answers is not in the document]
mfalt: I have rewritten most of this code, will probably push an update tomorrow.

src/types/DelayLtiSystem.jl
mfalt: What is the difference here?

src/types/PartionedStateSpace.jl
mfalt: Yes, we have it defined in the toolbox somewhere, an update of it is in my upcoming patch.

src/types/DelayLtiSystem.jl
baggepinnen: An argument error is more informative, and people doing error handling in applications sometimes rely on the proper type of error being thrown. See https://discourse.julialang.org/t/error-vs-throw-argumenterror/18915

baggepinnen: > Being more specific also allows you to write tests with @test_throws that don’t just pass spuriously if the same type of error is thrown in some other part of the code under test.

src/types/DelayLtiSystem.jl
```diff
@@ -0,0 +1,165 @@
...
+    end
+end
+
+# Efficient addition
+function +(sys::DelayLtiSystem{T}, n::T) where {T<:Number}
+    ny, nu = size(sys)
+    ssold = sys.P.P
+    # Add to direct term from input to output
+    new_D = ssold.D
+    new_D[1:ny, 1:nu] .+= n
```
olof3: Don't do broadcast

baggepinnen:
I rebased this on master, I think I managed to pull it off without screwing up.

PR #184, opened 2019-01-29
add inovation_form function with tests

baggepinnen:


baggepinnen:
This should be ready to pull now

PR #185, opened 2019-01-29
Avoid returning ::Adjoint in kalman

baggepinnen:


mfalt:
Why do we want this?

baggepinnen:
For consistency and to return the type the user expects. Some downstream
functions also do not accept adjoint. Creating an ss(A, K, C, 0) fails if K
is adjoint, since it's not the same matrix type as the others.

PR #189, opened 2019-01-31
Fix plotticks for consequtive plots

baggepinnen:
This PR makes the nice ticks in bode plots stay nice when several systems are plotted after each other
Previous behavior: the ticks were only drawn to cover one curve
New behavior: the ticks take all curves in a subplot into account

Have only made some small tests (positive) so far, should test on PlotTests as well


baggepinnen:
I have run through `runplotgen` and all plots look the same. My use case that looked bad before looks better now, the following looks terrible on master but good in PR
```julia
julia> G = tf(1,[1,1])

julia> bodeplot(G)

julia> bodeplot!(100G)
```
I'l pull if there are no objections

PR #192, opened 2019-03-04
WIP: Robust toolbox dev

mfalt:
Let's discuss the changes here.


example/hinf_example_MIMO.jl
```diff
@@ -0,0 +1,109 @@
...
+C = rand(ny,nx)
+D = rand(ny,nu)
+
+G = hInf_bilinear_z2s(hInf_bilinear_s2z(ss(A,B,C,D), 0.005))
+
+# Sensitivity weight function
+WSe = tf([0.1,1],[1000,1])
+#WS = [WSe 0 0 ; 0 WSe 0 ; 0 0 WSe]
+WS = [WSe 0 0 0 ; 0 WSe 0 0 ; 0 0 WSe 0 ; 0 0 0 WSe]
+WS = [WS 0*WS; WS*0 WS]
```
olof3: Perhaps best to try to avoid redefining local variables.

src/hinfinity_design.jl
```diff
@@ -0,0 +1,937 @@
+""" The code in this module is based on the bibentry [1] (glover1988state)
+below, and is best read alongside the complementary report in [2]
+(mgreiff2019report). For the LMI-experiments see the third bibentry [3].
+
+  @article{glover1988state,
```
olof3: How should we reference papers @mfalt , @baggepinnen ? Is this reasonable, or perhaps something a bit more concise?

example/hinf_example_MIMO.jl
mgreiff: Thanks, the general MIMO example should be removed, as it is currently dysfunctional - I have yet to add a loop shift in the controller synthesis - I'll remove this script in its entirety or now.

src/hinfinity_design.jl
```diff
@@ -0,0 +1,937 @@
...
+Check the assumptions for using the γ-iteration synthesis in Theorem 1. In
+future revisions, we could suggest possible changes to P should the system not
+be feasible for synthesis. However, this has not been too successful so far..!
+"""
+function hInf_assumptions(P::ExtendedStateSpace; verbose=true)
+
+    A, B1, B2, C1, C2, D11, D12, D21, D22 = ssdata(P)
+
+    # Check assumption A1
+    if !_is_stabilizable(A, B2)
```
mfalt: Should we consider making `is_stabilizable` a publicly available?

src/hinfinity_design.jl
```diff
@@ -0,0 +1,937 @@
...
+    return true
+end
+
+"""`[Mpinv] = function _compute_pseudoinverse(M::AbstractMatrix)`
+
+Compute the g-inverses for a non-rectangular matrix M, satisfying
+min(size(M))==rank(M). Used for checking assumptions A5 and A6 in [1], see the
+section on the generalized
+"""
+function _compute_pseudoinverse(M::AbstractMatrix)
```
mfalt: Does `pinv` work here?
Is this function (or enclosing function) somewhat performance critical? If so, there are some issues like type stability and `inv` insead of `/` that we should consider.

mfalt:
Some random comments
* All the examples (pdfs) should probably be put in https://github.com/JuliaControl/ControlExamples.jl
* Some sort of rebase? would probably be smart to keep the binaries out of the git history?, and to avoid conflics with changes that have been made since you created your branch. (For example in make.jl, makeplots.jl, plotting.jl)
* Should make indentation consistent (for example in `@recipe function specificationplot`)
* We should discuss if we should merge/separate ExtendedStateSpace and PartionedStateSpace (from delayed_lti branch)

src/hinfinity_design.jl
```diff
@@ -0,0 +1,937 @@
...
```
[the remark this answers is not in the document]
mgreiff: Very good point - it is definitely acceptable, and I have now removed the compute_pseudoinverse function.

src/hinfinity_design.jl
```diff
@@ -0,0 +1,937 @@
...
+"""
+
+
+"""`[flag] = function hInf_assumptions(P::ExtendedStateSpace; verbose=true)`
+
+Check the assumptions for using the γ-iteration synthesis in Theorem 1. In
+future revisions, we could suggest possible changes to P should the system not
+be feasible for synthesis. However, this has not been too successful so far..!
+"""
+function hInf_assumptions(P::ExtendedStateSpace; verbose=true)
```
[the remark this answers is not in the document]
mgreiff: Thanks for linking the style guide. All function names have been changed, and hopefully, they are now compliant with the style guide.

src/hinfinity_design.jl
mgreiff: Thanks, pinv works and has now been implemented instead of the _compute_pseudoinverse() function!

src/hinfinity_design.jl
mgreiff: Quite possibly, and I think that the bilinear discretization could be useful outside of the hinf_design module too, and should probably be implemented in the c2d() function. Similarly, we also have its inverse in bilineard2c(), which could be implemented in some function d2c().

I can start to migrate the functions to discrete.jl and matrix_comp.jl if you like.

freemin7:
Is MIMO still broken?
How stale is the PR?

baggepinnen:
@freemin7 see https://juliacontrol.github.io/RobustAndOptimalControl.jl/dev/#H_\\infty-and-H_2-design

PR #193, opened 2019-03-08
Updated the call of lsim in test_timeresp.jl to use keyword arguments…

hurak:
… for x0. The same modification done in timeresp.jl, wherein additionally the string for the label was corrected (transpose removed) and `using Plots` added.

### DIFF
--- a/src/timeresp.jl
+++ b/src/timeresp.jl
@@ -88,6 +88,8 @@ To simulate a unit step, use `(x,i)-> 1`, for a ramp, use `(x,i)-> i*h`, for a s
 Usage example:
 ```julia
 using LinearAlgebra # For identity matrix I
+using Plots
+
 A = [0 1; 0 0]
 B = [0;1]
 C = [1 0]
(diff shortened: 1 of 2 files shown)


hurak:
> Thanks for the fix. The semicolon is not needed at the call site, although it is accepted

Indeed, the semicolon is not needed. In fact, as I read it now in Julia (v1.1) manual, they explicitly state "When the function is called, the semicolon is optional: one can either call plot(x, y, width=2) or plot(x, y; width=2), but the former style is more common." You should have left it your way then. Hopefully, at least the other part of my fix was useful. 

Well, my first PR ever :-) 

Thanks for a good work. As a teacher of optimal and robust control (https://moodle.fel.cvut.cz/course/view.php?id=3866), I am tempted to use your package in the course. Towards that goal, I hope to be able to start submitting to your package.

baggepinnen:
Great to hear! Check out https://github.com/JuliaControl/ControlSystems.jl/pull/192 for upcoming robust control functionality

mfalt:
Thanks!

PR #204, opened 2019-04-16
Remove unnecessary type restrictions

baggepinnen:
This loosening of restrictions is often needed for handling weird float types (MonteCarloMeasurements.jl/dual numbers etc.). Also removes some boilerplate where generic type parameters were declared but never used.


baggepinnen:
All tests pass, will merge then

PR #207, opened 2019-04-18
Switch to Project.toml

baggepinnen:
This PR makes the switch to the new requirements specification. This is required to be able to tag new releases (which we should do imminently)


baggepinnen:
All checks pass, I'll merge this and tag a new patch release upon thumbs up
@mfalt

baggepinnen:
@JuliaRegistrator register()

PR #209, opened 2019-05-07
[WIP] Autovec

baggepinnen:


src/utilities.jl
```diff
@@ -116,3 +116,68 @@ index2range(ind1, ind2) = (index2range(ind1), index2range(ind2))
 index2range(ind::T) where {T<:Number} = ind:ind
 index2range(ind::T) where {T<:AbstractArray} = ind
 index2range(ind::Colon) = ind
+
+
+"""
+@autovec (idxs...) nout f() = (a, b, c)
+
+A macro that helps in creating versions of functions where excessive dimensions 
+are removed automatically for specific outputs. `idxs` are the indexes of the
```
baggepinnen: Indices

src/utilities.jl
```diff
@@ -116,3 +116,68 @@ index2range(ind1, ind2) = (index2range(ind1), index2range(ind2))
...
+    for i in 1:nout
+        if i in idxs
+            return_exp = :($return_exp..., vec(ret[$i]))
+        else
+            return_exp = :($return_exp..., ret[$i])
+        end
+    end
+
+    args = map(dict[:args]) do a
+        if a isa Expr
```
baggepinnen:
```suggestion
        a isa Expr ? a.args[1] : a
```

src/utilities.jl
```diff
@@ -116,3 +116,68 @@ index2range(ind1, ind2) = (index2range(ind1), index2range(ind2))
...
+
+    # Build the returned expression on the form (ret[1], vec(ret[2]), ret[3]...) where 2 ∈ idxs
+    return_exp = :()
+    for i in 1:nout
+        if i in idxs
+            return_exp = :($return_exp..., vec(ret[$i]))
+        else
+            return_exp = :($return_exp..., ret[$i])
+        end
+    end
```
baggepinnen:
```suggestion
    return_exp = :($return_exp..., i in idxs ? vec(ret[$i]) : ret[$i])
```

src/utilities.jl
```diff
@@ -116,3 +116,68 @@ index2range(ind1, ind2) = (index2range(ind1), index2range(ind2))
...
+            a.args[1]
+        else
+            a
+        end
+    end
+
+    quote
+        export $(esc(dict[:name])), $(esc(Symbol(dict[:name], "v")))
+
+        $(esc(f)) # Original function
```
baggepinnen: What does this line do on its own?

src/utilities.jl
```diff
@@ -116,3 +116,68 @@ index2range(ind1, ind2) = (index2range(ind1), index2range(ind2))
...
+        # `$($(esc(dict[:name])))v($($(esc.(dict[:args])...)); $($(esc.(dict[:kwargs]))...))`
+        """ 
+
+        For use with SISO systems where it acts the same as `$($(esc(dict[:name])))` 
+        but with the extra dimensions removed in the returned values.
+        """
+        function $(esc(Symbol(dict[:name], "v")))($(esc.(dict[:args])...); 
+                                                  $(esc.(dict[:kwargs])...))::$rtype
+            # TODO add where {...} to function definition
+            # TODO check if siso? what argument do we check?
```
baggepinnen: I think it's safe to assume we check the first argument for `issiso`. I can't think of a function which does not follow this convention.

src/utilities.jl
```diff
@@ -116,3 +116,68 @@ index2range(ind1, ind2) = (index2range(ind1), index2range(ind2))
...
+            # use issiso on something that is sys <: LTISystem
+            # loop through args to find?
+            for a in ($(esc.(args)...),)
+                if a isa LTISystem
+                    @assert issiso(a)
```
baggepinnen:
```suggestion
issiso(a) || throw(ArgumentError("Only SISO systems accepted to $(esc(Symbol(dict[:name], "v")))"))
```

test/test_autovec.jl
```diff
@@ -0,0 +1,12 @@
+@testset "test_autovec" begin
+
+# Check output of bode and make sure dimensions are correct
+sys = tf([1], [1, 2])
+mag, phase, w = bode(sys)
+@test size(mag) == size(phase) == (size(w)[1], 1, 1)
```
baggepinnen:
```suggestion
@test size(mag) == size(phase) == (size(w,1), 1, 1)
```

src/utilities.jl
albheim: The export or the $(esc(f))?

src/utilities.jl
albheim: So I should remove the loop over the arguments then?

src/utilities.jl
baggepinnen: The `$(esc(f))`, it's not saved in a local variable and not really part of any other expression?

src/utilities.jl
baggepinnen: While the `issiso` must be resolved at runtime, I think the loop can be compiled away since `isa` operates on an input argument (which type is known at compile time) and a hard coded type. I don't think it's an issue to have the loop there at all so might as well keep it.

mfalt: Without `@generated` I think it is hard to force doing at compile time, and I don't think its worth the effort. I would say a runtime check as below is fine.

baggepinnen: I mean that the compiler will be able to get rid of it without a generated function, since `a isa LTIsystem` is resolvable at runtime
```
julia> f(a) = a isa Float64
f (generic function with 1 method)

julia> @code_llvm f(1.)

;  @ REPL[1]:1 within `f'
define i8 @julia_f_15995(double) {
top:
  ret i8 1
}
```

src/utilities.jl
albheim: I just realized this does not seem to be working as intended, not really sure whats going on but I'll try to figure it out.

albheim: Changed my mind, it does seem to work.

src/utilities.jl
```diff
@@ -153,22 +152,19 @@ macro autovec(indices, nout, f)
 
         $(esc(f)) # Original function
 
-        # TODO fix following row and add to documentation
-        # `$($(esc(dict[:name])))v($($(esc.(dict[:args])...)); $($(esc.(dict[:kwargs]))...))`
+        # TODO following row does not work, existing works... don't understand why
+        # `$($(esc(dict[:name])))v($($(esc.(get(dict, :args, []))...)); $($(esc.(get(dict, :kwargs, []))...)))`
         """ 
+        `$($(esc(dict[:name])))v($(join([arg for arg in $(esc(dict[:args]))], ", ")); $([kwarg for kwarg in $(esc(dict[:kwargs]))]...))`
```
albheim: Not sure why this works and the above one does not.

mfalt: You seem to be doing a join with the string ", ", over a set of symbols? Im not sure if that could be a problem

baggepinnen:
I added some tests for autovec in c2ea925 and managed to find some cases where the macro is not really working yet. When travis finishes, you'll see that the user specified  keyword arg does not carry through on the line
`@test t1v(kw=2) == (2,[2, 2])`

Also, the macro throws a weird error when defining a method with parametric type constraints, e.g.,
```julia
ControlSystems.@autovec (2,) 2 function t2(; kw::T=1) where {T <: Integer}
    return kw, [kw kw]
end
```

albheim:
> I added some tests for autovec in [c2ea925](https://github.com/JuliaControl/ControlSystems.jl/commit/c2ea9253dab3cc3145c0e31ebc4e318963f5bf90) and managed to find some cases where the macro is not really working yet. When travis finishes, you'll see that the user specified keyword arg does not carry through on the line
> `@test t1v(kw=2) == (2,[2, 2])`
> 
> Also, the macro throws a weird error when defining a method with parametric type constraints, e.g.,
> 
> ```julia
> ControlSystems.@autovec (2,) 2 function t2(; kw::T=1) where {T <: Integer}
>     return kw, [kw kw]
> end
> ```

Nice! I know there is some problems with the keywords. Have not really sanitized them at all so some problems are expected, especially when defining types.

src/utilities.jl
```diff
@@ -123,3 +123,77 @@ index2range(ind1, ind2) = (index2range(ind1), index2range(ind2))
...
+
+function extractvarname(a)
+    if typeof(a) == Symbol
+        return a
+    elseif a.head == :(::)
+        return a.args[1]
+    else
+        return extractvarname(a.args[1])
+    end
+end
```
albheim: I'm pretty sure this covers all the cases, but might be some syntax stuff I'm still not 100% on. 
Also realised after i pushed that most other internal utility functions seems to be named with a starting underscore and use snake case, so I assume I should do that here also?

src/utilities.jl
```diff
@@ -123,3 +123,77 @@ index2range(ind1, ind2) = (index2range(ind1), index2range(ind2))
...
+        if i in indices
+            return_exp = :($return_exp..., vec(result[$i]))
+        else
+            return_exp = :($return_exp..., result[$i])
+        end
+    end
+
+    args = map(dict[:args]) do a
+        a isa Expr ? a.args[1] : a
+    end
```
albheim: Old stuff that should be removed...

albheim (dismissed):
Have some more stuff to fix...

src/utilities.jl
baggepinnen: > starting underscore and use snake case

The julia design guidlines recommend to write names without underscore if they are readable, and with underscores if they are not readable without. Small letters everywhere except for in type and module names.

albheim: Okay, thanks. Saw one place which used leading underscore in that file and assumed that might be some convention used here, but looking again there was only that place.

baggepinnen:
@albheim can you remember what the status was on this PR? It would be nice to have it merged if at all possible :)

albheim:
@baggepinnen I have some vague memory about fixing it to a point to where I thought it should work, but then I got errors from travis which I did not necessarily think was from my thing and was too lazy to fix around that time.

Looking at the errors now it seems like when precompiling controlsystems -> ordinarydiffeq -> diffeqbase it crashes on the use of eachcol somewhere in that package for the julia 1.0 version. Looking at the [docs](https://docs.julialang.org/en/v1.1/NEWS/#New-library-functions-1) it seems eachcol was added in 1.1.

Not super good at the versioning system and package system, but seems like that is something that should be a requirement in diffeqbase rather than controlsystems? They should have something saying that this version is not compatible with julia 1.0? We could probably fix it by constraining our version of ordinarydiffeq, but seems ugly.

albheim:
Mostly to rerun travis, not sure how to manually trigger it. But also saw I had not fixed a comment from @baggenpinnen about underscores.

albheim:
Okay, pulled in latest master and also committed a small change i had been working on locally to remove one of the arguments for autovec. Seems to be working fine now and for me this could be merged. 

Should it be documented more in some way? Given that it is only an internal macro I assume we don't want anything about it in the official docs, so I only have some documentation for it in docstrings connected to the function. The docstrings generated for the new functions are also not the most interesting, but I think it is enough for now at least.

baggepinnen:
LGTM 😃 thanks for working though this! We'll wait for an approval from @olof3 or @mfalt as well before proceeding

src/ControlSystems.jl
```diff
@@ -71,9 +71,9 @@ export  LTISystem,
         # Frequency Response
         freqresp,
```
mfalt, line 72: We should probably add the following also:
freqrespv, numvecv, denvecv, numpolyv, denpolyv.
Otherwise LGTM.

mfalt (approved):

albheim (approved):
Didn't mean to self request a review, but not sure how to get rid of it without commenting so LGTM

PR #210, opened 2019-05-10
Specialize simulator types

baggepinnen:
For performance


baggepinnen:
Will pull soon if no objections

PR #211, opened 2019-05-10
Abstract statespace

baggepinnen:
This introduces `AbstractStateSpace` so that it becomes *significantly* easier to implement new statespace types


src/types/TransferFunction.jl
```diff
@@ -141,6 +141,12 @@ end
 -(G::TransferFunction) = TransferFunction(-G.matrix, G.Ts)
 
 ## MULTIPLICATION ##
+function (*)(A::AbstractMatrix{<:SisoRational}, B::AbstractMatrix{<:SisoRational})
+    # TS = promote_op(matprod, eltype(A), eltype(B)) # TODO; this promote_op does not work 100% and often returns ControlSystems.SisoRational{_A} where _A even though it should be ControlSystems.SisoRational{eltype(A)}
+    TS = promote_type(eltype(A), eltype(B))
```
mfalt: This is not technically correct, but maybe enough. It assumes the type is closed under addition and multiplication.

baggepinnen: Yeah it's unfortunate, but as the comment says, using promote_op causes type instabilities that are critical in some code (not only for performance, but for correct promotion). The cocstring on promote_op explains why it sucks.. 😜

mfalt: I don't even see the reason why we need this function. Can't we use the standard callback (generic-matmul)? Or would that result in the problem we are trying to overcome now?

mfalt: Ok, I understand now. I think this is all right for now. It could be a bit problematic for SisoZpk (which is not included in this PR). Is the problem we are fixing not existing for Zpk?

mfalt: And could you add a test, or at least an example here?

baggepinnen: This method is a copy-paste of base's generic matmul, with the `TS` line changed. I added it because it caused type instability in critical cases.

This method is a copy-paste of base's generic matmul, with the `TS` line changed. I added it because it caused type instability in critical cases.

baggepinnen: > And could you add a test, or at least an example here?
### Before patch
```julia
julia> C
TransferFunction{ControlSystems.SisoRational{Float64}}
        300.0*s + 100.0
-------------------------------
1.0*s^3 + 1600.0*s^2 + 1000.0*s

julia> S
TransferFunction{ControlSystems.SisoRational{StaticParticles{Float64,20}}}
       1.0*s^4 + 1600.0 ± 2.0*s^3 + 6590.0 ± 3200.0*s^2 + 3490.0 ± 2000.0*s
-----------------------------------------------------------------------------------
1.0*s^4 + 1600.0 ± 2.0*s^3 + 6590.0 ± 3200.0*s^2 + 3860.0 ± 1900.0*s + 123.0 ± 54.0

julia> C*S
TransferFunction{ControlSystems.SisoRational}
               300.0*s^5 + 481000.0 ± 610.0*s^4 + 2.14e6 ± 970000.0*s^3 + 1.71e6 ± 930000.0*s^2 + 349000.0 ± 200000.0*s
--------------------------------------------------------------------------------------------------------------------------------------
1.0*s^7 + 3200.0 ± 2.0*s^6 + 2.57e6 ± 6500.0*s^5 + 1.21e7 ± 5.2e6*s^4 + 1.28e7 ± 6.2e6*s^3 + 4.06e6 ± 1.8e6*s^2 + 123000.0 ± 54000.0*s

```
Inferred type is abstract `TransferFunction{ControlSystems.SisoRational}`
### After patch
```julia
julia> C*S
TransferFunction{ControlSystems.SisoRational{StaticParticles{Float64,20}}}
               300.0*s^5 + 481000.0 ± 610.0*s^4 + 2.14e6 ± 970000.0*s^3 + 1.71e6 ± 930000.0*s^2 + 349000.0 ± 200000.0*s
--------------------------------------------------------------------------------------------------------------------------------------
1.0*s^7 + 3200.0 ± 2.0*s^6 + 2.57e6 ± 6500.0*s^5 + 1.21e7 ± 5.2e6*s^4 + 1.28e7 ± 6.2e6*s^3 + 4.06e6 ± 1.8e6*s^2 + 123000.0 ± 54000.0*s

```
Inferred type is concrete and correct. Can't come up with an example with simple types that fails

src/types/promotion.jl
```diff
@@ -25,7 +25,15 @@
 #Base.promote_rule(::Type{StateSpace{T1}}, ::Type{StateSpace{T2}}) where {T1,T2} = StateSpace{promote_type(T1, T2)}
 
 # NOTE: Is the below thing correct always?
-Base.promote_rule(::Type{StateSpace{T1,MT1}}, ::Type{StateSpace{T2,MT2}}) where {T1,T2,MT1,MT2} = StateSpace{promote_type(T1, T2),promote_type(MT1, MT2)}
+function Base.promote_rule(::Type{StateSpace{T1,MT1}}, ::Type{StateSpace{T2,MT2}}) where {T1,T2,MT1,MT2}
+    MT = promote_type(MT1, MT2)
+    if isconcretetype(MT) # This sometimes fails and produces a Unionall. This condition can be checked at compile time, checked by f(x,y) = (MT = promote_type(x, y); isconcretetype(MT)); @code_llvm f(Int,Float64)
+        StateSpace{promote_type(T1, T2),MT}
+    else # If fail, fall back to an ordinary matrix
+        FT = promote_type(T1, T2)
```
mfalt: If we add `@warn` here we could tell users that unexpected behavior is happening, and possibly try to fix/instruct the maintainers of corresponding packages to fix `promote_type` of those types.

baggepinnen: Yes, but warnings are extremely annoying if there is no mechanism to turn them off. Imagine having that warning displayed 10000 times in a loop.. I'm not sure it's worth the effort on our part to implement the global state to keep track of if the warning has been issued or not

mfalt: You are right, there is no clear way for the user to avoid it. But it could probably also result in pretty obscure bugs and performance hits.

mfalt:
I think this looks good, not convinced of the name, but can't come up with anything better.
The only thing that is lacking is tests for the new type.

baggepinnen:
Tests have been added in 0845fd3

baggepinnen:
Satisfied with the [tests](https://github.com/JuliaControl/ControlSystems.jl/commit/0845fd3ba0ba11dc5cdfbf1315ab0c8995731f76) @mfalt ?

mfalt:
I think it suffices. Would be nice to see if it works through promotion and stuff for sparse matrices or static matrices in the tests, but I assume you have tested for static in your code?

baggepinnen:
Yeah I've used it for a week or so with a lot of weird stuff in there, it seems to be working as expected. If someone *really needs* support sparse stuff and find that it creates problems they can open an issue, until then I'm fine with assuming that works

PR #216, opened 2019-05-23
Better tol in delay solve

baggepinnen:
This PR gives the user a hint on how to improve the tolerance for difficult problems. Improves by 2 orders of magnitudes in test


mfalt:
Can't we just default to that solver for impulse?

baggepinnen:
One could indeed do that :P Consider it done.

baggepinnen:
Done, pull @mfalt?

PR #218, opened 2019-05-23
Delayfixes

baggepinnen:
This PR 
- Generalizes the type of DelayLTIsystem
- Defines an inverse of delay system
- fixes an bug where a matrix was written to in place


mfalt:
Do we want `DelayLTI` to not restrict the delays to `Real` as you have?

baggepinnen:
I think it's unlikely to cause any problems, but done in a7e13d7

PR #219, opened 2019-05-27
WIP Fixplots and example

mfalt:


mfalt:
These look good on my computer, pulling

PR #220, opened 2019-05-27
Moredelays

baggepinnen:
Some fixes and generalizations for delay systems

### DIFF
--- a/src/connections.jl
+++ b/src/connections.jl
@@ -136,8 +136,7 @@ end
 
 # Used in typed_hvcat
 function Base.typed_hcat(::Type{T}, X...) where {T<:LTISystem}
-    S = promote_type(typeof.(X)...)
-    hcat(convert.(S, X)...)
+    hcat(convert.(T, X)...)
 end
 # Ambiguity
 Base.typed_hcat(::Type{T}, X::Number...) where {T<:LTISystem, N} = hcat(convert.(T, X)...)
--- a/src/connections.jl
+++ b/src/connections.jl
@@ -135,7 +135,10 @@ function Base._cat_t(::Val{2}, T::Type{<:LTISystem}, X...)
 end
 
 # Used in typed_hvcat
-Base.typed_hcat(::Type{T}, X...) where {T<:LTISystem} = hcat(convert.(T, X)...)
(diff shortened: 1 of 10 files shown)


src/connections.jl
```diff
@@ -135,7 +135,10 @@ function Base._cat_t(::Val{2}, T::Type{<:LTISystem}, X...)
 end
 
 # Used in typed_hvcat
-Base.typed_hcat(::Type{T}, X...) where {T<:LTISystem} = hcat(convert.(T, X)...)
+function Base.typed_hcat(::Type{T}, X...) where {T<:LTISystem}
+    S = promote_type(typeof.(X)...)
+    hcat(convert.(S, X)...)
+end
```
mfalt: I'm not sure this is actually what we want to do.

src/types/DelayLtiSystem.jl
```diff
@@ -156,10 +156,14 @@ function feedback(sys1::DelayLtiSystem, sys2::DelayLtiSystem)
...
 end
 
-function delay(tau::S, T::Type{<:Number}=Float64) where S
+function delay(T::Type{<:Number}, tau)
     return DelayLtiSystem(ControlSystems.ss([zero(T) one(T); one(T) zero(T)]), [T(tau)])
 end
 
+function delay(tau::S) where S
+    return DelayLtiSystem(ControlSystems.ss([zero(S) one(S); one(S) zero(S)]), [S(tau)])
+end
```
mfalt: I agree with the argument order above, but I'm nbot sure we should drop the Float default. For example, the lsim above will break for a `delay(1)` right?

src/connections.jl
baggepinnen: No, probably me who messed up while experimenting

src/types/DelayLtiSystem.jl
baggepinnen: I defaulted to `promote_type(Float64, typeof(tau))` instead.

baggepinnen:
Ready to pull and tag

src/connections.jl
baggepinnen: See here?

PR #221, opened 2019-06-04
RFC: Reference generator

RykteT:


mfalt:
We should figure out how this can be used in some convenient way for `lsim` or similar.

src/reference_generators.jl
```diff
@@ -0,0 +1,227 @@
...
+end
+
+function (ref::PWCGenerator)(t::Real)
+    if t < ref.t[1]
+        return 0
+    elseif t >= ref.t[end]
+        return ref.v[end]
+    end
+
+    for (idx, time) in enumerate(ref.t)
```
baggepinnen: Collecting this generator will have quadratic complexity in the length of t?

src/reference_generators.jl
```diff
@@ -0,0 +1,227 @@
...
+INPUT:
+    f::T = The frequency of the square wave
+    A::T = The amplitude of the square wave
+"""
+struct SquareGenerator{T <: Real} <: Function
+    f::T #Frequency
+    A::T #Amplitude
+end
+
+SquareGenerator(f) = SquareGenerator(f, (1)::typeof(f))
```
baggepinnen:
```suggestion
SquareGenerator(f) = SquareGenerator(f, oftype(f, 1))
```

src/reference_generators.jl
```diff
@@ -0,0 +1,227 @@
...
+
+INPUT:
+    v::T = The variance of the GNGenerator
+    m::T = The mean of the GNGenerator
+    N    = The normal distribution
+"""
+struct GNGenerator{T <: Real} <: Function
+    v::T #Variance
+    m::T #Mean
+    N::Distribution    #TODO: How to fix this?
```
baggepinnen: Insert another type parameter which is restricted to subtypes of Distribution

src/reference_generators.jl
```diff
@@ -0,0 +1,227 @@
...
+end
+
+GNGenerator(v, m) = GNGenerator(m, v, Normal(m, v))
+GNGenerator(v)    = GNGenerator((0)::typeof(v), v, Normal((0)::typeof(v), v))
```
baggepinnen: Use `oftype`

test/test_reference_generators.jl
```diff
@@ -0,0 +1,124 @@
...
+    gn2 = GNGenerator(1, 1)
+
+    t = zeros(1_000_000)
+    @test abs(mean(gn1.(t))) < 1e-2
+    @test abs(mean(gn2.(t)) - 1)< 1e-2
+end 
+
+# Noise Tests
+@testset "NoiseGenerator Tests" begin
+    using Distributions
```
baggepinnen: For this to work, Distributions must be added as an extra test dependency in Project.toml

src/reference_generators.jl
RykteT: It will! Could it be fixed by overloading the Base.collect function? It would become linear instead I would assume.

test/test_reference_generators.jl
RykteT: > How does broadcasting of generators work? Should be tested as well.
> For example
> 
> ```julia
> gen.(t::Vector)
> ```

Is this not tested?

RykteT:
> How does broadcasting of generators work? Should be tested as well.
> For example
> 
> ```julia
> gen.(t::Vector)
> ```

Is this not already tested?

src/reference_generators.jl
```diff
@@ -0,0 +1,227 @@
...
+
+INPUT:
+    f::T = The frequency of which to repeat the sawtooth signal
+    p::T = The phase of the sawtooth signal
+    frac::Function = The sawtooth function. Please do not alter this
+"""
+struct SawtoothGenerator{T <: Real} <: Function
+    f::T #Frequency
+    p::T #Phase
+    frac::Function #Not to be added from outside
```
baggepinnen: Why is this a field in the struct if it is not supposed to be changed? Couldn't it be hardcoded then?

RykteT: You are definitely correct! I was not certain on how to implement this in a julia-fashion. Will fix it!

PR #223, opened 2019-06-11
Remove defenition of dlqr for vector case

mfalt:
It seems completely unnessesary, i.e the original definition seems to work for vectors already?


mfalt:
We should probably add a simple test and maybe look into why 7df41efe4dc81e3966f62aba06e381e328eda53e said this was a bug. Maybe because of old julia version?

baggepinnen:
This was perhaps legacy from before transposes were taken seriously ? :P

PR #227, opened 2019-08-27
Fix pidplots for cases where plot type is not :controller

### LINKED ISSUE #226: pidplots returns nothing unless :controller in args
The function ends with 4 sequential if statements. In Julia, the result of the expression is returned. If `:controller` was in args, the result of the last expression is the result of the call to `bodeplot`. Otherwise, the result of the last expression is `nothing`.

non-Jedi:
Fix #226

Logic checking for `:nichols` was removed as nothing was being done with `nichols_` in current code.


mfalt:
I think this looks good, can we pull this @baggepinnen ?

baggepinnen:
I made some further changes so that the example in the docs works as intended. It produces all plots well on my local machine.

Issue #229, opened 2019-10-21
showterm2 function out of date


andrersimoes:
Hi gentlemen,

I'm receiving an error inside the **showterm2** (polyprint.jl) function regarding the number of parameters used at **Polynomials.printsign** call.

``
ERROR: MethodError: no method matching printsign(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Float64, ::Int64, ::Bool, ::MIME{Symbol("text/plain")})                                                                    
``

The j element is no longer being used by the correlated **Polynomials** package code.

Would you mind if I remove the extra argument and push the correction ?

Regards,
André.

mfalt:
I noticed that, but forgot to push a fix. Feel free to do it, just make sure to set v0.5.3 as the lower limit for Polynimials.jl

andrersimoes:
Hi Mattias.

Do I need any procedure or specific permission to accomplish git push ?

``
remote: Permission to JuliaControl/ControlSystems.jl.git denied to andrersimoes
fatal: unable to access 'https://github.com/JuliaControl/ControlSystems.jl.git/': The requested URL returned error: 403
``
By the way, it seems that your manifest is already up-to-date with Polynomials v0.5.3. Only polyprint required a patch at my side.

baggepinnen:
You need to push to your for of Controlsystems.jl and create a pull request to this repository.

phelipe:
It was merged?

mfalt:
Just merged it @phelipe

phelipe:
I did a pull-request to update project.toml( #237), I think it solve problems with DSP.jl.

PR #230, opened 2019-10-22
update of showterm2 due to 'j' parameter removed from Polynomials.printsign

andrersimoes:
See issue #229.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,3 +18,6 @@ Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[compat]
+Polynomials = "0.5.3"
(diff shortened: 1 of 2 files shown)


mfalt:
Could you also please set v0.5.3 as the lower limit for Polynimials.jl? I.e. add 
```
[compat]
Polynomials= "0.5.3"
```
to the Project.toml file

andrersimoes:
Project.toml altered and tested inside an empty .julia directory. Please, check if it works on your side too.

baggepinnen:
Nice, can we have a patch release for this as well?

mfalt:
> 
> 
> Nice, can we have a patch release for this as well?

Yes, I think we have quite a few updates, especially regarding delay systems since the last version

mfalt:
It seems like delay systems still dosn't have a tagged vesrion, so I'll take a look tomorrow and see what issues/PRs we can include in a 0.6 release this weekend. I think the delay systems should be bigger than a patch version. Any thoughts @baggepinnen ?

baggepinnen:
The tagging of minor release indicates a breaking release for major version 0. If we should follow julias version of semver, we should only tag 0.6 if something has broken, otherwise it's a patch release. We could also consider tagging 1.0 if we think that the API will be stable from now on.

The downside of tagging a breaking release is that everyone that has a dependency on ControlSystems will have to update their compat file, otherwise they will block controlsystems from updating.

PR #234, opened 2019-10-31
Proposed 0.5.4 version

mfalt:
- Updated version
- Added Julia 1.2 to test
- Fixed really weird bug in norminf


.travis.yml
```diff
@@ -2,6 +2,7 @@ language: julia
 julia:
   - 1.0
   - 1.1
```
baggepinnen:
```suggestion
```
Looks good, can stop testing on 1.1 if we test on 1.2

mfalt:
We have a few problems:
- I don't know how to reliably sort the eigenvalues so that we get complex conjugated poles together, which is needed for some methods (fix_conjugate_pairs)
- Acker does not return a real K even when the poles are complex-conjugated.

baggepinnen:
Take complex conjugate of all poles with negative imaginary part and then
solve the assignment problem using the Hungarian algorithm?

mfalt:
> Take complex conjugate of all poles with negative imaginary part and then solve the assignment problem using the Hungarian algorithm?

This seems like it would require us to put some epsilon for when they should be matched or not. It is certainly a possibility, but seems a bit unnecessary when solvers always solve the conjugates together. I created an alternative PR with "no sorting"

PR #236, opened 2019-11-04
Release 0.5.4 option2

mfalt:
Similar to https://github.com/JuliaControl/ControlSystems.jl/pull/234 but without sorting of eigenvalues. The sorting therefore defaults to that of the solver, which in LAPACK keeps complex conjugated poles together, and the same seems to be true for GenericLinearAlgebra.

This would potentially be breaking for users in Julia 1.2 if they expect a specific ordering, although no ordering was documented. ~~The same is true for users of Julia < 1.2 with the other alternative.~~

~~I had to copy and edit `roots` from `Polynomials.jl` since it orders by default.~~


src/utilities.jl
```diff
@@ -16,35 +16,30 @@ to_matrix(T, A::Number) = fill(T(A), 1, 1)
...
+    eigvalsnosort(args...; kwargs...) = eigvals(args...; kwargs...)
+end
 
 # NOTE: Tolerances for checking real-ness removed, shouldn't happen from LAPACK?
 # TODO: This doesn't play too well with dual numbers..
 # Allocate for maxiumum possible length of polynomial vector?
 #
 # This function rely on that the every complex roots is followed by its exact conjugate,
-# and that the first complex root in each pair has positive real part. This formaat is always
+# and that the first complex root in each pair has positive real part. This format is always
```
mfalt: positive real part -> positive imaginary part

mfalt:
Set Polynomials to use new print interface (>=0.6.0).
There was a bad bug in DSP (0.6.0) should we set DSP (>=0.6.1) to avoid https://github.com/JuliaDSP/DSP.jl/issues/320 ?
Bump out version to 0.5.4

If tests pass I think we should pull this. @baggepinnen ? @olof3 ?

PR #242, opened 2019-12-11
Norminf refactor

olof3:
The algorithms for continuous and discrete time are now closer to the cited papers, more readable, and also works for complex-coefficient systems. Many tests have been added.

It could be considered if it is worth to merge the algorithms for discrete and continuous time into one. One option to simplify this would be to consider actual frequencies and not normalized frequencies in the discrete-time formulation and define something like `sigmamax(sys, w)`. But perhaps better to wait until there is dispatch on continuous and discrete time.

Another thing which could be discussed is the naming. I think L-infinity and H-infinity norms should be distinguished, in particular as the current function computes the L-infinity norm and most people are probably expecting that it is the H-infinity norm. So perhaps renaming the current function to `linfnorm`and defining something like
`hinfnorm(sys) = isstable(sys) ? linfnorm(sys) : Inf`

Switching the naming of norminf to (h/l)infnorm is consistent with opnorm etc. and seems easier to say and remember.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Some of the available commands are:
 ##### Constructing systems
 ss, tf, zpk, ss2tf
 ##### Analysis
-pole, tzero, norm, norminf, ctrb, obsv, gangoffour, margin, markovparam, damp, dampreport, zpkdata, dcgain, covar, gram, sigma, sisomargin
+pole, tzero, norm, hinfnorm, linfnorm, ctrb, obsv, gangoffour, margin, markovparam, damp, dampreport, zpkdata, dcgain, covar, gram, sigma, sisomargin
 ##### Synthesis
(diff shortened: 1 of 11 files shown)


olof3:
@mfalt was also supportive of having `hinfnorm` and `linfnorm` instead of `norminf` so that has been implemented, and `norminf` has been deprecated.

Instead of defining `hinfnorm(sys) = isstable(sys) ? linfnorm(sys) : Inf`
I went with the following solution, again after some discussion with @mfalt  

```julia
function hinfnorm(sys::AbstractStateSpace; tol=1e-6)
if iscontinuous(sys)
         _infnorm_two_steps_ct(sys, :hinf, tol)
    else
        return _infnorm_two_steps_dt(sys, :hinf, tol)
     end
end
end
```
and then having `_infnorm_two_steps_dt` depend on if `:hinf` or `:linf` is passed as an argument. This avoids re-computation of poles and that the peak gain for marginally stable systems can be returned without duplicating code. These advantages seem to slightly outweigh the worse separation of functionality.

PR #247, opened 2019-12-29
prevent promotion to Float64 in unwrap

baggepinnen:


baggepinnen:
CI Failure seems unrelated. Latest master failed with same error when retriggered
https://travis-ci.org/JuliaControl/ControlSystems.jl/jobs/618353490?utm_medium=notification&utm_source=github_status

Seems we need a new compat entry somewhere

mfalt:
Looks good. This big was hard to pin down. It seems to happen randomly, and best solution seems to be sorting the poles in damp. Olof fixed this in https://github.com/JuliaControl/ControlSystems.jl/pull/242

PR #256, opened 2020-01-21
WIP type stable sample time

mfalt:
Attempt to make sample times part of the type system.
~~This is still WIP, 390 tests pass, 146 does not.~~
~~I particular we need to fix concatenation of systems with numbers. Although, we can make them of the correct type, they still need to have the correct same time, or some implicit sample type (like -1).~~ Implemented with `UNDEF_TIME`

* ~~Docs are not updated~~ See pr https://github.com/JuliaControl/ControlSystems.jl/pull/293 for update to docs
* We should add tests for `Static` systems if we keep it


src/types/AbstractSampleTime.jl
```diff
@@ -0,0 +1,67 @@
...
+ts_same(x::AbstractSampleTime, y::AbstractSampleTime) = throw(ArgumentError("Sampling time mismatch"))
+ts_same(x::AbstractSampleTime, y::AbstractSampleTime, z...) = ts_same(ts_same(x, y), z...)
+
+function ts_same(x::Discrete{T}, y::Discrete{S}) where {T,S}
+    if x != y
+        throw(ArgumentError("Sampling time mismatch"))
+    end
+    return Discrete{promote_type(T,S)}(x)
+end
+ts_same(x::Continuous, y::Continuous) = Continuous()
```
mfalt: Add n arg version

mfalt:
Tests now pass except for HeteroStateSpace.

We still have some open questions:
**Is there a need to denote systems as "Static" in the type system?** The advantage would be that they work both with Continuous and Discrete systems, but numbers/matrices already satisfy this to some degree.

**I went with the implementation of having sample-time as "-1"when it is unspecified.** This should mostly occur during promotion, in particular concatenation, but should never be visible to the user. 
Implemented internally using `const UNDEF_TIME = -1`.
Any system with this sample time would inherit other sample times, for example:
```julia
sys1 = tf([1],[1 ,1], 0.5)  # Discrete
sys2 = tf([1],[1, 1], -1)   # Implicit discrete time
sys1 + sys2
[sys1 sys2]
```
However, it is not possible to do
```julia
sys1 = tf([1],[1 ,1])      # Continuous
sys2 = tf([1],[1, 1], -1)  # Implicit discrete time
sys1 + sys2  # Error
[sys1 sys2]  # Error
```

**Should `HeteroStateSpace` be limited to have `AbstractMatrix` or `AbstractArray` as elements?**

**Should we still print systems that have no poles or zeros with the text "static system"?**

src/freqresp.jl
```diff
@@ -80,9 +80,9 @@ function (sys::TransferFunction)(s)
 end
 
 function (sys::TransferFunction)(z_or_omega::Number, map_to_unit_circle::Bool)
-    @assert !iscontinuous(sys) "It makes no sense to call this function with continuous systems"
+    @assert !iscontinuous(sys) && !isstatic(sys) "It makes no sense to call this function with continuous systems"
```
olof3: Not from this PR, but continuous -> continuous-time. Not entirely clear to me what does this function does.

src/freqresp.jl
```diff
@@ -178,8 +178,8 @@ function _bounds_and_features(sys::LTISystem, plot::Val)
         w1 = 0.0
         w2 = 2.0
     end
-    if !iscontinuous(sys) # Do not draw above Nyquist freq for disc. systems
-        w2 = min(w2, log10(π/sys.Ts))
+    if !iscontinuous(sys) && !isstatic(sys) # Do not draw above Nyquist freq for disc. systems
```
olof3: Not related to this PR, but think it should be possible to plot frequency response above the Nyquist frequency. For example in order to understand folding effects of down sampling. Again, not sure what this function does, so this might be unrelated.

src/types/TimeType.jl
```diff
@@ -0,0 +1,76 @@
...
+Base.promote_rule(::Type{Continuous}, ::Type{Continuous}) = Continuous
+Base.promote_rule(::Type{Discrete{T1}}, ::Type{Discrete{T2}}) where {T1,T2}= Discrete{promote_type(T1,T2)}
+
+Base.convert(::Type{Discrete{T1}}, Ts::Discrete{T2}) where {T1,T2} = Discrete{T1}(Ts.Ts)
+
+# Promoting two or more systems systems should promote sample times
+ts_same(x::TimeType, y::TimeType) = throw(ErrorException("Sampling time mismatch"))
+ts_same(x::TimeType, y::TimeType, z...) = ts_same(ts_same(x, y), z...)
+
+function ts_same(x::Discrete{T1}, y::Discrete{T2}) where {T1,T2}
```
olof3: I would prefer something like `find_common_samplingtime` or `find_common_Ts` to make it a bit more clear what is being done.

src/timeresp.jl
```diff
@@ -227,7 +227,7 @@ end
 
 function _default_Ts(sys::LTISystem)
     if !iscontinuous(sys)
```
olof3: Should be `isdiscrete`

src/types/TimeType.jl
```diff
@@ -0,0 +1,76 @@
+
+abstract type TimeType end
+
+const UNDEF_TIME = -1
```
olof3: `UNDEF_SAMPLETIME` ?

src/types/StateSpace.jl
```diff
@@ -178,38 +214,36 @@ function isapprox(sys1::ST1, sys2::ST2) where {ST1<:AbstractStateSpace,ST2<:Abst
 end
 
 ## ADDITION ##
-function +(s1::StateSpace{T,MT}, s2::StateSpace{T,MT}) where {T, MT}
+function +(s1::StateSpace{TimeT,T,MT}, s2::StateSpace{TimeT,T,MT}) where {TimeT,T, MT}
     #Ensure systems have same dimensions
     if size(s1) != size(s2)
         error("Systems have different shapes.")
```
olof3: Would be helpful to write the sizes of the systems in the error message.

src/connections.jl
```diff
@@ -22,23 +22,17 @@ Append systems in block diagonal form
...
     ST = promote_type(typeof.(systems)...)
-    Ts = systems[1].Ts
-    if !all(s.Ts == Ts for s in systems)
-        error("Sampling time mismatch")
-    end
     A = blockdiag([s.A for s in systems]...)
     B = blockdiag([s.B for s in systems]...)
     C = blockdiag([s.C for s in systems]...)
     D = blockdiag([s.D for s in systems]...)
+    Ts = ts_same([s.Ts for s in systems]...)
```
olof3: Perhaps put this before concatenating the matrices for consistency.

olof3:
Good updates in the last commit.

1. I'm leaning towards yes, having created a system `K = ss([1 2; 3 4])` it would be annoying if it can't be used a P-controller for a discrete-time system.

2. Seems reasonable.

4. For systems of type StateSpace{Continuous} and StateSpace{<:Discrete} that have empty A, B and C matrices it would be weird to write out the empty matrices, and just not writing them out would also be a bit weird, perhaps putting `(static)` in the print somewhere would be reasonable, to indicate that it is not actually of type Static. Note however that `tf(1)` does not indicate that the system is static. Perhaps Static shouldn't be a `TimeType` but rather a subtype of `LTISystem`, without having thought too much about it, this would seem quite logical.

5. `Ts` seems convenient and well established. Is this against some design guideline? The `A`, ... `D` fields are also upper case.

mfalt:
All good comments. @baggepinnen do you have any opinion on point 3 above? It seems that there was no such formal limitation before, but that the conversion to matrixes happened in practice.

baggepinnen:
While I could dream up examples with general linear operators instead of matrices, I do not have any such applications in mind. Would implementation be easier if the restriction is enforced? Display and construction of the systems are decoupled from the internal representation of them so those two points should not pose any problems either way?

mfalt:
> While I could dream up examples with general linear operators instead of matrices, I do not have any such applications in mind. Would implementation be easier if the restriction is enforced? Display and construction of the systems are decoupled from the internal representation of them so those two points should not pose any problems either way?

The problems I encountered we mostly that for example the B matrix often is entered as a vector. Not only does this cause problems with the printing, which should be easily solved, but for example converting back to a dense system is non trivial. Most of our methods probably assume that the arrays are two dimensional. I can try to make it work for abstract array without conversion to matrix first.

src/matrix_comps.jl
```diff
@@ -397,7 +397,7 @@ function _infnorm_two_steps_dt(sys::AbstractStateSpace, normtype::Symbol, tol=1e
     # Check if there is a pole on the unit circle
     pidx = findfirst(on_unit_circle, pole_vec)
     if !(pidx isa Nothing)
-        return (T(Inf), angle(pole_vec[pidx])/T(sys.Ts))
+        return (T(Inf), angle(pole_vec[pidx])/sampletime(sys))
```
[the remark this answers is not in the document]
mfalt: A bit unclear, I guess the T is there to make the function type stable, ill add the sampletype to the promotion of frequencies.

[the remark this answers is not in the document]
baggepinnen: Maybe not only type stable, but to prevent accidental promotion of numeric types etc.

src/types/TimeType.jl
mfalt: Changed to `common_sample_time`

olof3:
@mfalt, regarding your quick question about the frequency axis for bode plots,

In `_bounds_and_features'
```julia
    if !iscontinuous(sys) # Do not draw above Nyquist freq for disc. systems
        w2 = min(w2, log10(π/sys.Ts))
    end
```
could arguably be changed to something like
```julia
    if isdiscrete(sys)
        w2 = log10(π/sys.Ts)
    end
```
In my opinion, plotting up to the Nyquist frequency would be the preferred default behavior.

mfalt:
After talking to @olof3, I think we should change the name of the `sys.Ts` field to something else, for example `time` or `timetype`. `sys.Ts` will then simply be a wrapper for `sampletime(sys)` / `sys.time.Ts`.
Other than that, I think we should go ahead and update and merge into master.

mfalt:
We seem to have lost some sort of type-stability with the last commit. For some reason, the constant propagation doesn't work for `isdiscrete`. I am not able to figure out why. This accounts for the only error I get when testing locally.

mfalt:
@JuliaControlBot test-plots

mfalt:
@JuliaControlBot test-plots

mfalt:
I'll look over the missing coverage, it seems like we are mostly missing tests for `Static` systems. This implementation should be working, and I think it might be necessary to have the static type for some types of promotions, but if we don't want it, I will see if it is avoidable. It should at least be possible to make sure that the user never creates or receives a static system, but that it is only created temporarily in some promotions.
We should also make decisions on the following:
1. Name of field that holds the sampletime type, `sys.time` currently
2. Name for checking type of sampletime, currently `iscontinuous`, `isdiscrete` (`isstatic`)

PR #259, opened 2020-02-04
WIP: Padé approximations, general feedback interconnections and test systems

olof3:
More general feedback interconnections were necessary to conveniently apply Padé approximations to delay systems. 

Added functions:
Padé approximations: `pade`
Feedback interconnections: `feedback`, `starprod`, `lft`
Test systems: `ssrand`, `systemdepot`


src/demo_systems.jl
```diff
@@ -0,0 +1,110 @@
...
+end
+
+first_order_system(;T=1) = ss(-1/T, 1, 1/T, 0)
+fotd(;T=1, τ=1) = ss(-1/T, 1, 1/T, 0) * delay(τ)
+sotd(;T=1, T2=10, τ=1) = ss(-1/T, 1, 1/T, 0)*ss(-1/T2, 1, 1/T2, 0)*delay(τ)
+resonant(;ω0=1, ζ=0.25) = ss([-ζ -ω0; ω0 -ζ], [ω0; 0], [0 ω0], 0) # QUESTION: Is this the form that we like? Perhhaps not.
+
+
+"""
+Doyle's spinning body (satellite) example
```
baggepinnen: Might be helpful to have the function signature in the docstring

olof3:
> Would it be a good idea to have demo systems in their own module so that they are easy to find and do not crowd the regular names pace?

As another module in `ControlSystems.jl`?  That sounds like a good idea. The current ambition is a for very basic functionality and a small number of systems with different characteristics. I.e., something much much less elaborate than `MatrixDepot` for example. But a separate module still sounds very reasonable.

olof3:
This should be good to pull now. I went over it with @mfalt, any thoughts from @baggepinnen ?
I haven't changed the implementation of feedback of `DelayLTISystem` and `PartionedStateSpace` to use the new more general feedback function, I think we have to think a little bit more about exactly how to do that.

baggepinnen:
This is an enormous PR by now 😜 no objections from me

baggepinnen:
Tests are not passing though

mfalt:
Coverage fails which is not too bad, but it seems to be mostly bacause of the pade code. `_linscale` and `pade` is added but never tested. I think these should be tested if we add them

mfalt:
Since this is quite a big PR, we should proably wait a litte while before tagging to make sure everything works as expected. We have quite a few PRs in the pipeline that can be released together soon.

src/types/DelayLtiSystem.jl
```diff
@@ -173,27 +184,33 @@ end
...
-#     if a > 0
-#         error("Delay needs to be causal.")
-#     end
-#
-#     return exp(b) * delay(a)
-# end
+function delay(T::Type{<:Number}, τ)
+    return DelayLtiSystem(ControlSystems.ss([zero(T) one(T); one(T) zero(T)]), [T(τ)])
+end
+delay(τ::Number) = delay(promote_type(Float64,eltype(τ)), τ)
```
mfalt, line 194: @olof3 Why eltype here?

olof3: Good question, shouldn't be like that. Contradicts the doc string as well. I'll try to remember to fix it.

Additionally, I don't really see why we need to promote the numeric type of the `StateSpace` to the type of the delay, but perhaps this is natural?

Additionally, \tau is restricted to `Real` in the `DelayLTISystem` type, so perhaps it doesn't make sense to have any restriction here. I guess one can conceive of use cases where the delays are `Dual` numbers, but `Dual` is not a subtype of `Real` (but of `Number`), feels weird to only restrict the delays to `Number` though.

PR #272, opened 2020-03-17
choose more sensible yticks for Phasediagram in Bodeplot (unconditionaly)

freemin7:
This commit makes the yticks of the phase diagram in the BodePlot a multiple of 45° and 90°.
Somehow a previous pull request (changing the documentation of bodeplot to mention `setPlotScale(str)`) ended being included. No idea how that happened. This pull request was already accepted though.
This behavior has currently no switch, since 45° or 90° are always sensible unless you are dealing with exotic transfer functions (i think it was fractions powers of s).

### DIFF
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -69,6 +69,28 @@ function seriesextrema(xylims, plotattributes, subplot)
 end
 extremareducer(x,y) = (min(x[1],y[1]),max(x[2],y[2]))
 
+function getPhaseTicks(x, minmax)
+    minx, maxx =  minmax
+    min               = ceil(minx/90)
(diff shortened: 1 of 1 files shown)


baggepinnen:
Looks good to me. The one who merges should run the plot reference app as well. Have you tested it manually that it produces nice looking plots when several systems are plotted in the same plot? Especially if the different systems have phase curves spanning different intervals of the phase axis.

freemin7:
I think i tested it.
Both the case where it spans less than a full rotation and the the case where is does span more than a full rotation. I was happy with the results. But that feels like forever ago. I can't believe it was just 3 days. My god does Corona make me productive.

PR #273, opened 2020-03-25
Plotfixes

baggepinnen:


baggepinnen:
I've been running this code for quite some time and it seems to do what's expected. Any objections to pulling?
@mfalt

mfalt:
Looks good to me. Go ahead!

PR #279, opened 2020-03-31
Script to respond to comments in PRs

mfalt:
Checks any comments on PRs for the string `@JuliaControlBot test-plots` and if the user is an admin, then pings ControlExamplePlots.jl to do tests and respond with a comment.

See corresponding script here: https://github.com/mfalt/ControlSystems.jl/blob/master/.github/workflows/CommentResponder.yml
which is triggered here:
https://github.com/mfalt/ControlSystems.jl/pull/3


mfalt:
I think this should be ready now @baggepinnen , updated the existing script to not report failure when the PR is created from a fork. In that case the script would previously fail because it is "dangerous" to let forks access to the login for the JuliaControlBot. Now instead of failing it skips the step. We can use the feature in this PR of commenting on PRs from forks instead to run the script "manually".

.github/workflows/PingControlExamplePlots.yml
```diff
@@ -10,7 +10,11 @@ jobs:
     steps:
       - name: PR ping ControlExamplePlots
         run: |
+         if [ "$SECRET" != "" ]; then
```
mfalt, line 13: This is to avoid weird issue of environment variables not being directly available in if statements.

.github/workflows/PingControlExamplePlots.yml
```diff
@@ -10,7 +10,11 @@ jobs:
...
           curl -X POST https://api.github.com/repos/JuliaControl/ControlExamplePlots.jl/dispatches \
           -H 'Accept: application/vnd.github.everest-preview+json' \
           -u ${{ secrets.ACCESS_TOKEN_BOT }} \
           --data '{"event_type": "prupdate", "client_payload": { "repository": "'"$GITHUB_REPOSITORY"'", "pr_number":  ${{ github.event.number }} }}'
+         fi
+        env:
+          SECRET: ${{ secrets.ACCESS_TOKEN_BOT }}
```
mfalt, line 20: This is to avoid weird issue of environment variables not being directly available in if statements.

mfalt:
Aby objections to pulling this @baggepinnen ?

baggepinnen:
Go ahead

PR #284, opened 2020-04-07
WIP: improvmenets to time responses

### LINKED ISSUE #264: Some issues with lsim (mostly for discrete-time systems)
Errors are thrown in the following cases: 

for Int-systems
```
lsim(ss(1,1,1,1,1), randn(10), 0:9)
```

for Complex-value input signals
```
lsim(ss(1.0,1,1,1,1), randn(ComplexF64,10), 0:9)
```

Also, I think it should be possible to use `lsim` for discrete-time systems without having to provide a time vector.

olof3:
`ltitr` was re-written to use in-place operations and allow different numeric types of the input data.

This allows complex-valued simulations (closes #264) and makes things faster.

Minor changes were done to `lsim`:
* Avoid checking `_issmooth` when simulating discrete-time systems (`_issmooth` fails for complex signals) 
* `x0` was restricted to be a `Vector` to avoid broadcast problems, this seems quite reasonable and should be good for performance? Or should it be `AbstractVector`?
* Also added a check that the time vector is regularly spaced, if it isn't then the result will be horribly wrong.

It will be possible to improve things by quite a bit when we can dispatch on continuous or discrete. For example, `ltitr` could probably be replaced with the discrete-time version of `lsim`.


baggepinnen:
- `AbstractVector` is always to be preferred, you never know when there's a `FillArray` etc. appearing as an initial state.
- Do you have some before and after benchmarks to share? (It's very satisfying to see successful perf. improvements :D )

src/timeresp.jl
```diff
@@ -187,18 +198,31 @@ If `x0` is not provided, a zero-vector is used.
...
-    x = similar(A, size(A, 1), n)
-    for i=1:n
-        x[:,i] = x0
-        x0 = A * x0 + B * u[i,:]
+
+    # Transposing u allows column-wise operations, which apparently is faster.
+    # See issue for discssuing the orientation of the input u and the output x.
+    ut = transpose(u)
+
+    x = Matrix{T}(undef, size(A, 1), n)
```
baggepinnen, line 213: This is totally fine, but are you aware of `similar` which is supposed to make this more convenient?

src/timeresp.jl
```diff
@@ -118,20 +118,31 @@ function lsim(sys::StateSpace, u::AbstractVecOrMat, t::AbstractVector;
...
 
     dt = Float64(t[2] - t[1])
-    if !iscontinuous(sys) || method == :zoh
-        if iscontinuous(sys) # Looks strange to check iscontinuous again
+    if !all(x -> x ≈ dt, diff(t))
+        error("time vector t must be uniformly spaced")
+    end
+
+    if iscontinuous(sys)
+        if method == :unspecified
```
baggepinnen, line 126: While `var == :sym` is fine, `var === :sym` is said to be less stressful for the compiler (Ref: Jeff, but I cant remember where). The same goes for comparison to all singleton types, such as `nothing` and `missing`. 
```suggestion
        if method === :unspecified
```

src/timeresp.jl
mfalt: I think we might actually want something dense here, but maybe not. For example it might be good if GPU arrays or similar might persist.

baggepinnen: Good point, `similar` is to be preferred then so that CuArrays etc. work out.

src/timeresp.jl
olof3: We use `==` in quite many other places (and `===` in some), guess consistency would be the main thing (should be an easy find-and-replace). I wouldn't have figured this out by myself, but if it is the way to go then we should change. What do you say @mfalt

src/timeresp.jl
olof3: Yes, it's a convenient function but, I guess in this case `x` isn't that similar to `A`, the eltype and one dimension are different. Also I guess one would want a dense matrix even if A would be sparse?

olof3:
Some benchmarks of `ltitr`, quite many allocations from the slices, but looked through this with @mfalt , and didn't see any obvious improvements. 5-argument `mul!` gives another 10 percent or so.

```
Benchmark 1:
old:   37.197 μs (802 allocations: 89.20 KiB)
new:   16.504 μs (1199 allocations: 54.75 KiB)
Benchmark 2:
old:   49.565 μs (802 allocations: 134.52 KiB)
new:   27.535 μs (1199 allocations: 65.73 KiB)
```

```julia
using BenchmarkTools

function ltitr(A::AbstractMatrix{T}, B::AbstractMatrix{T}, u::AbstractVecOrMat,
        x0::VecOrMat=zeros(T, size(A, 1))) where T
    N = size(u, 1)
    x = similar(A, size(A, 1), N)
    for i=1:N
        x[:,i] = x0
        x0 = A * x0 + B * u[i,:]
    end
    return transpose(x)
end


@views function ltitr_new(A::AbstractMatrix, B::AbstractMatrix, u::AbstractVecOrMat,
        x0::AbstractVecOrMat=zeros(eltype(A), size(A, 1)))

    T = promote_type(LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(A), eltype(x0)),
                      LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(B), eltype(u)))

    n = size(u, 1)

    # Transposing u allows column-wise operations, which apparently is faster.
    # See issue for discssuing the orientation of the input u and the output x.
    ut = transpose(u)

    x = Matrix{T}(undef, size(A, 1), n)
    tmp = Vector{T}(undef, size(A, 1)) # temp vector for storing B*u[:,k]

    x[:,1] .= x0
    for k=1:n-1
        mul!(x[:,k+1], A, x[:,k])

        mul!(tmp, B, ut[:,k]) # these two lines could be replaced with 5-argument mul! (>= v1.3)
        x[:, k+1] .+= tmp
    end
    return transpose(x)
end

nx = 3; nu = 1
A = rand(nx,nx)
B = rand(nx,nu)
u = rand(200,nu)
x0 = rand(nx)

println("Benchmark 1:")
print("old: ")
@btime ltitr(A, B, u, x0)
print("new: ")
@btime ltitr_new(A, B, u, x0)

nx = 10; nu = 5
A = rand(nx,nx)
B = rand(nx,nu)
u = rand(200,nu)
x0 = rand(nx)

println("Benchmark 2:")
print("old: ")
@btime ltitr(A, B, u, x0)
print("new: ")
@btime ltitr_new(A, B, u, x0)
```

baggepinnen:
Nice, that's quite an okay speedup. The extra allocations are expected and could potentially be removed using UnsafeArrays.jl, even though I very seldom see that working in practice. The small allocations are amortized for larger ste vectors so this is still very nice 😃

src/timeresp.jl
mfalt: There is definetely some places where i might have used `==` instead of `===` for symbols and singeltons, if we find them we could fix them.

olof3: I'll to a find-replace then.

src/timeresp.jl
baggepinnen: Would it be similar to u?

olof3: Yes, this would be better, or perhaps even better with `x0`. it is set to `Matrix` by default, so one would have to specify `x0` to get another matrix type. This is perhaps good enough for now.

olof3:
Seems like it is possible to shave off another 30-40 % or so if all the B*u operations are done as a matrix--matrix multiplication (avoiding many of the view allocations?), although the code becomes less readable.

Continuation from the benchmark above.

```julia
# with 5-argument mul!
@views function ltitr_newest1(A::AbstractMatrix, B::AbstractVecOrMat, u::AbstractVecOrMat,
        x0::AbstractVecOrMat=zeros(eltype(A), size(A, 1)))

    T = promote_type(LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(A), eltype(x0)),
                      LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(B), eltype(u)))

    n = size(u, 1)

    x = Matrix{T}(undef, size(A, 1), n)
    x[:,1] .= x0

    mul!(x[:, 2:end], B, transpose(u[1:end-1, :])) # Initialize the result matrix x with B*u

    for k=1:n-1
        mul!(x[:,k+1], A, x[:,k], 1, 1) # muladd Ax[,k] to B*u that is already stored x[:,k+1]
    end
    return transpose(x)
end


# without 5 argument mul!
@views function ltitr_newest2(A::AbstractMatrix, B::AbstractVecOrMat, u::AbstractVecOrMat,
        x0::AbstractVecOrMat=zeros(eltype(A), size(A, 1)))

    T = promote_type(LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(A), eltype(x0)),
                      LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(B), eltype(u)))

    n = size(u, 1)

    tmp = Vector{T}(undef, size(A, 1)) # temp vector for storing B*u[:,k]

    x = Matrix{T}(undef, size(A, 1), n)

    mul!(x[:, 2:end], B, transpose(u[1:end-1, :])) # Initialize x with B*u

    x[:,1] .= x0
    for k=1:n-1
        mul!(tmp, A, x[:,k])
        x[:,k+1] .+= tmp
    end
    return transpose(x)
end
```


```
Benchmark 1 (nx=3, nu=1, nt=200):
old:       37.937 μs (802 allocations: 89.20 KiB)
new:      17.248 μs (1199 allocations: 54.75 KiB)
newest1:  9.905 μs (403 allocations: 23.66 KiB)
newest2:  12.718 μs (603 allocations: 33.09 KiB)

Benchmark 2 (nx=10, nu=5, nt=200):
old:      49.805 μs (802 allocations: 134.52 KiB)
new:      27.397 μs (1199 allocations: 65.73 KiB)
newest1:  15.670 μs (403 allocations: 34.59 KiB)
newest2:  19.249 μs (603 allocations: 44.08 KiB)

Benchmark 3 (nx=50, nu=5, nt=200):
old:      117.297 μs (803 allocations: 393.84 KiB)
new:      80.540 μs (1200 allocations: 128.52 KiB)
newest1:  65.013 μs (404 allocations: 97.05 KiB)
newest2:  74.126 μs (604 allocations: 106.86 KiB)
```

src/timeresp.jl
```diff
@@ -105,8 +105,8 @@ y, t, x, uout = lsim(sys,u,t,x0=x0)
 plot(t,x, lab=["Position" "Velocity"], xlabel="Time [s]")
 ```
 """
-function lsim(sys::StateSpace, u::AbstractVecOrMat, t::AbstractVector;
-        x0::VecOrMat=zeros(sys.nx), method::Symbol=_issmooth(u) ? :foh : :zoh)
+function lsim(sys::AbstractStateSpace, u::AbstractVecOrMat, t::AbstractVector;
+        x0::AbstractVector=zeros(Bool, sys.nx), method::Symbol=:unspecified)
```
baggepinnen, line 109: Won't this cause `x0` to be type unstable when it's promoted in the function body?

src/timeresp.jl
```diff
@@ -185,20 +196,39 @@ lsim(sys::TransferFunction, u, t, args...; kwargs...) = lsim(ss(sys), u, t, args
...
+                      LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(B), eltype(u)))
+
     n = size(u, 1)
-    x = similar(A, size(A, 1), n)
-    for i=1:n
-        x[:,i] = x0
-        x0 = A * x0 + B * u[i,:]
+
+    # Transposing u allows column-wise operations, which apparently is faster.
+    # See issue for discssuing the orientation of the input u and the output x.
```
baggepinnen, line 213: Which issue?

baggepinnen (approved):
It would be nice to pull this PR :)
You think it's ready @olof3 ?

src/timeresp.jl
olof3: Hmm, I don't remember if I ever posted that issue. But we can have a short discussion here instead :) @mfalt 

I was thinking about if it would make more sense that the input `u` and the output `x` are specified the other way around, with each column corresponding to different time instances. I think it does make more sense, but the print becomes more awkward. I don't think that we should change it in the short term, but we should start thinking about if this is really the way we want it. It is parallel with the discussion about frequency responses. 

```suggestion
```

src/timeresp.jl
olof3: It might get overwritten if `method == :foh`, but not sure that I see the problem? Feel free to suggest a fix. I guess it better than what was there before?

olof3:
Yes, I think this could be pulled. 

I saw that you @baggepinnen had made some changes to the promotion in `lsim` in another PR, I removed that change as I believe that that is already handled in this PR, making it redundant. Tests seem to pass, so hopefully I didn't break anything.

However, we should revisit `lsim` sometime soon. With the introduction of `TimeEvolution` the `lsim` function could be made a lot cleaner by moving a lot of the functionality to the `lsim(::AbstractStateSpace{Discrete}, ...)` function (that you coincidentally just introduced in an other PR). Then the continuous-time version could call the discrete version when appropriate. That would avoid some logic and clutter. 

But perhaps that could be for another time.

mfalt (approved):

mfalt:
@JuliaControlBot test-plots

(Updating the reference plots, since the bot was broken)

PR #286, opened 2020-04-09
Force less options in plots

baggepinnen:


baggepinnen:
@mfalt would be nice with some plots tests on this one. The plot bot fails with `ERROR: LoadError: UndefVarError: EXACT_MATCH not defined` here https://github.com/JuliaControl/ControlExamplePlots.jl/runs/572954953?check_suite_focus=true#step:5:2918

baggepinnen:
I have no idea where [this](https://travis-ci.org/github/JuliaControl/ControlSystems.jl/jobs/672824869#L490) test failure comes from, it's appears unrelated to the PR but still seems like a proper failure that should be addressed.

mfalt:
>  The plot bot fails with `ERROR: LoadError: UndefVarError: EXACT_MATCH not defined`

I thought I had fixed that, but must have missed it, should work now
@JuliaControlBot test-plots

mfalt:
It also seems like [this](https://travis-ci.org/github/JuliaControl/ControlSystems.jl/jobs/672824869#L511) is a problem with polynomials now, did you actually test that 0.7 worked when you pulled https://github.com/JuliaControl/ControlSystems.jl/pull/285 ?

mfalt:
Wee seem to be getting a lot of warnings that i think are new except for the standard 
```
latex: failed to create a dvi file
sh: 1: latex: not found
```
we are now getting
```
┌ Warning: Attribute alias `annotation` detected in the user recipe defined for the signature (::ControlSystems.Nicholsplot). To ensure expected behavior it is recommended to use the default attribute `annotations`.
```
and the same for a couple `xlabel->xguide`,`ylabel->yguide`,`color->seriescolor`

baggepinnen:
> did you actually test that 0.7 worked when you pulled #285 ?

[CI passed ](https://travis-ci.org/github/JuliaControl/ControlSystems.jl/builds/672273019) but apparently installed an old version of Polynomials 
> Installed Polynomials ────────────────── v0.6.1
I'll have a look at it and see what the issue is

mfalt:
> [CI passed ](https://travis-ci.org/github/JuliaControl/ControlSystems.jl/builds/672273019) but apparently installed an old version of Polynomials

I'm not sure why that happens sometimes, but I guess that is why the message says
> Note: I have not tested your package with this new compat entry. It is your responsibility to make sure that your package tests pass before you merge this pull request.

baggepinnen:
> I'm not sure why that happens sometimes

It's probably due to one of our other deps not having declared compat with Poly 0.7 yet. I have not thought about this problem before but it can actually be a rather big pain point in keeping compats up to date. Deps that many other packages depend on are likely to cause this issue, and then the CI reported on the CompatHelper PR is likely to have used an old version of the updated dep. The ways I can see to test with the latest version is either of
1) Wait for a long time until all other deps have updated compat and run the tests again
2) Locally edit the project file of all deps that keep the new version from being installed

I really hope there is a better way than any of those two?

mfalt:
>     1. Wait for a long time until all other deps have updated compat and run the tests again
The only reasonable way I can see, is to wait until the other deps support the new version. This should be less and less of an issue now that more packages are getting to 1.0.

Regarding this PR, it took a while for me to understand the Hz part. It seems thet the user should still supply the frequency vector in rad/s, and that only the axis is changed? We should probably add a small docstring for that.

baggepinnen:
> It seems thet the user should still supply the frequency vector in rad/s, and that only the axis is changed

That's a good point, it's probably awkward to have it that way. Maybe if `hz` is supplied the freq vector is assumed to be in Herz as well?

mfalt:
> That's a good point, it's probably awkward to have it that way. Maybe if `hz` is supplied the freq vector is assumed to be in Herz as well?

Probably, as long as it is documented I am fine either way, but I think supplying in hz is more reasonable, however, we would have to put it in `_processfreqplot` and `_default_freq_vector`

mfalt:
Also, it might more consistent to treat it the same way as we treat dB/abs, however, I am not fond of the old way we treat that with a global variable.

baggepinnen:
The python control package makes `hz` control only plotting, while the input frequency vector is always `rad/s`
https://web.math.princeton.edu/~cwrowley/python-control/generated/control.bode.html
Needs to be documented though

PR #292, opened 2020-04-10
Updates for Polynomials v1.0

baggepinnen:
A few test failures remain.


baggepinnen:
Tests seem to pass on this PR now. It has gotten really hard to install ControlSystems or any package depending on it lately due to the update of Polynomials, so it would be nice to have this merged soon.

baggepinnen:
If there are no objections, I'll merge this in a few hours.

mfalt:
I think it looks good, so we should probably tag s version after this to make it easier to install.

PR #294, opened 2020-04-14
Suggetstions for Sampletimes 

olof3:
This PR gives some variations of @mfalt's original branch ([diff against master](https://github.com/JuliaControl/ControlSystems.jl/compare/master...JuliaControl:sampletimes_test1)). See also his PR to this branch.

UPDATE:

Together with @mfalt a lot of problems and solutions have been found and things are starting to converge to something reasonable. Below is a summary of our thoughts.

SUMMARY:

(note that the code is somewhat simplified)

There will be an `abstract type` that captures if a system is continuous-time or discrete-time, and in the latter case contains a field `Ts` for the sample time. This type parameter will enable dispatch on continuous/discrete time.
```
abstract type TimeType end

struct Continuous <: TimeType end
struct Discrete{T} <: TimeType
    Ts::T
end
```

`LTISystem`s will keep track of a field `time::TimeType`
```
struct StateSpace{TimeT <: TimeType, T}
    A::Matrix{T}
    B::Matrix{T}
    C::Matrix{T}
    D::Matrix{T}
    time::TimeT
end
```
and have an interface `time` that returns this time field, for example
`time(sys::StateSpace) = sys.time::TimeType`.

Information about the system can be inferred from the `time` field, and it will be possible to dispatch on the type parameter,

```julia
lqr(sys::StateSpace{Continuous}, Q, R) = lqr(sys.A, sys.B, Q, R)
lqr(sys::StateSpace{<:Discrete}, Q, R) = dlqr(sys.A, sys.B, Q, R)
```

Another advantage of having `time::TimeType` with information about both continuous/discrete time and the sample time `Ts` is the convenience of something like

```
function Base.vcat(systems::TransferFunction...)
    time = common_time(systems...)
    mat = vcat([s.matrix for s in systems]...)
    return TransferFunction(mat, time)
end
```

OTHER POINTS:
* `getproperty` is used to allow accessing `Ts` directly for discrete-time systems through `sys.Ts`
* One should not access the sampletime of continuous time systems (since there is none), one should instead check `isdiscrete` before trying to do so. For continuous-time systems `sys.Ts` is deprecated.
* `propertynames` could be used to suggest auto completion in the REPL of `Ts` for a discrete-time `sys.`  (without doing so for continuous-time systems). Is this a good idea? With `propertynames` we could also hide the field `time` that is primarily for internal use, however we're leaning towards not doing so.
* A type `struct Static <: TimeType end` was considered, but the simplicity of only continuous and discrete time seemed to outweigh the small advantages. It should be possible to just use a `Matrix` for static systems.

NOTATIONAL PROBLEMS:
* There could be confusion over `time` and sample time (`Ts`). A field `time` also looks like a time vector. However we had a hard time to come up with something better, suggestions are welcome. a suggestion from @mfalt is  `timedomain`.
* `Ts` that denotes the sample time of discrete-time systems is not a great name byjulia naming guidelines and for confusion with type parameters that are often denoted by captial `T`. Alternatives would be `h`, `ts`, `dt` (Python), or `sampletime`. Matlab uses `Ts`. In the literatuere $$T$$ (and $T_s$) is the most common, $h$ is also used sometimes. So `Ts` is perhaps the option that causes the least confusion. `h` looks nice, but might be less self explanatory than `Ts`, it is also a bit short for kwargs and `sys.h` looks like a C header file. `ts` would also be an option in order to adhere to style guidelines, but it looks a bit a time vector or a point in time.

Feedback on this would be greatly appreciated. Perhaps there are some good solutions that we haven't thought about.

Here are some last points where @mfalt and I have slightly different ideas ([COMPARISON](https://github.com/JuliaControl/ControlSystems.jl/compare/3690cc8..sampletimes3#diff-876f1732a4c3a667d5e02cb22d582abcR77-R78)).

1. Definting a function `sampletime(sys::LTISystem) = sampletime(time(sys))` and using this internally instead of `sys.Ts`.

pros:
* It is posssible to have `sampletime(sys)` throw an error for continuous time systems while deprecating `sys.Ts` for discrete-time systems.
* Perhaps better style than `sys.Ts` (overloading `getproperty` is mostly used for matrix factorizations)
* It is easier to do broadcasting as `sampletime.(systems)` than `getproperty.(systems,:Ts)`

cons:
* It is confusing that the system types have a field `time` (that could be considered as sample time in a wide sense) but that this field is not returned by  `sampletime(sys)`. With `sys.Ts` it should be quite clear what is returned.
* `sampletime(sys)` is significantly longer than `sys.Ts` and might require introducing a new variable for readability in some places.

2. To avoid confusion with `time` I think it could be an idea to avoid using `sampletime` in the code and perhaps use `sampleperiod` or something else.

Suggestions and ideas are welcome.


src/types/conversion.jl
```diff
@@ -20,24 +20,24 @@
...
     C = MT(fill(zero(T), (ny,0)))
     D = convert(MT, D)
-    return StateSpace{TimeT,T,MT}(A,B,C,D,undef_sampletime(TimeT))
+    return StateSpace{TimeT,T,MT}(A,B,C,D,undef_Ts(TimeT))
 end
 
 # TODO We still dont have TransferFunction{..}(number/array) constructors
 Base.convert(::Type{TransferFunction{TimeT,SisoRational{T}}}, b::Number) where {TimeT, T} =
-    tf(T(b), undef_sampletime(TimeT))
+    tf(T(b), undef_Ts(TimeT))
```
baggepinnen, line 38: Isn't it a bit incosistent to change the name here from `sampletime -> Ts` while changing above from `Tf -> tfinal` and `Ts -> dt`?

olof3: Time is usually lower case `t`, hence `tfinal` (`tf` would not have been so good),  `dt` is the spacing of the time-vector (put to `sys.Ts` for discrete time systems), so those changes seemed motivated regardless.

`undef_Ts` is probably not the right thing, but it is consistent with the suggestions in this commit. I'm quite sure that will be changed very soon :) Perhaps to undef_sampleperiod

mfalt:
See https://github.com/olof3/ControlSystems.jl/pull/7 for counter-suggestions

olof3:
In the current PRs, `time::TimeType` is used for keeping track of continuous/disrete time, and in the latter case also keeping track of `Ts`.
However I just noted that the standard library `Dates` already exports `TimeType`, adding possibility for confusion. It is also a further indication of that `time` & `TimeType` are primarily understood as things related to time in the normal sense of the word,

I think it would be preferable with some better terminology.

One suggestions could be `TimeType` -> `GeneralSampleTime` and the field `time` could be changed to `gstime`. For the function `time(::LTISystem)` it could either be just `gstime(sys)` or `general_sampletime(sys)`. This should make it clear that it refers to sampletime-like concept, but also make it clear that it is not _the_ sampletime/sampleperiod `Ts`.

baggepinnen:
I have no idea what a general sample time means? Does it refer to the type of the sampletime? if `TimeType` is taken, is `SampleTimeType` a better description?

olof3:
Just to clarify, `TimeType`/`GeneralSampleTime` would be the supertype of `Continuous` and `Discrete{T<:Number}`, and `time`/`gstime` the name of the field that contains an instance of a such type.

Such an instance is not quite a sampletime $T_s$ of a discrete-time system. I think sampletime would be a quite acceptable abuse of notation, but @mfalt thought it was too much of a stretch, and i do see his point. Hence the `General` to indicate that it is not quite a sampletime $T_s$. It's a made up term, some other made up name would perhaps be better, I don't know of any "standard term" for this thing.

baggepinnen:
Hmm OK, made up names can be okay I guess, but it has to be somewhat descriptive then, general doesn't really tell me anything, generalized would perhaps be more indicative. Is the problem that a continuous-time system is not sampled with any finite interval? Is the common supertype perhaps `SampleSpace` or `TimeSpace`? A space in the mathematical sense can be either continuous, like R, or discrete, like Z, and the time axis  is defined on the entire space

olof3:
You're right, Generalized would be the right word. Then it gets a bit long, but it seems quite accurate, so I'd be happy with that.

I think the main objection for avoiding sampletime in this context is that it is $T_s$ that should be the sampletime. But with the prefix generalized (or just g) it should be fine.

Sample space could be a bit [ambiguous](https://en.wikipedia.org/wiki/Sample_space). `TimeAxis` feels a bit undescriptive, but I think it could be an alternative if we don't come up with anything else.

`SamplingOperator` and `sampleop` could be another idea.

olof3:
As discussed it might be a bit confusing with `time::TimeType` to describe if a system is discrete or continuous time
(`time` seems like it could indicate a point in time or a time vector, and `TimeType` is exported by stdlib package `DateTime`).

[Time evolution](https://en.wikipedia.org/wiki/Time_evolution) seems like the concept that we are after, so perhaps `timeevol::TimeEvolution`?

Another option could be `timeprop::TimePropagation` or `timepropag`.

`systime::SysTime` is another option which is obviously a bit confusing since it is a common name, otherwise `timeaxis` as @baggepinnen suggested, or `timeflow`.

mfalt:
Closed by https://github.com/JuliaControl/ControlSystems.jl/pull/315

PR #295, opened 2020-04-17
Sampletimes: Counter-suggestion to suggestion PR 294 to PR 256 

mfalt:
Alternative names to suggestion https://github.com/JuliaControl/ControlSystems.jl/pull/294
Chnages:
* Exports `sampletime(sys)` and uses it internally instead of `sys.Ts` to simplify interface which throw errors when the system is `Continous`. (It is still possible to use `.Ts`, but it throws a warning on `Continous` systems)
* Interface: `time(sys)` for `LTISystems`
  * New `<:LTISystem` systems only need to define `time(sys)` to get working `sampletime, isdiscrete, iscountinous, common_time`
  * Defaults to `sys.time`
  * In the context of general `LTISystems` it is only used in the fallbacks for the above mentioned functions, and once as `typeof(time(sys))` in `zpkdata`.
* `UNDEF_SAMPLEPERIOD` to `UNDEF_SAMPLETIME`
* Removed `timetype(sys)`, it was only used once, and can be replaced by `typeof(time(sys))`
* Removed hiding of the field `time`. I am open to adding back `Ts` to `propertynames` to make autocompletion easier.

Possible additional changes:

Chage the internal functions that have `AbstractStateSpace` to as inputs to use `time(sys)` instead of `sys.time` to make it robust to new types `<:AbstractStateSpace`. However, this is not nessesary since all current implementations have the field `time`, and new implementations could implement `getproperty`.

Edit: I am open to changing `.time` and `time(sys)` to `timedomain`.

### DIFF
--- a/src/ControlSystems.jl
+++ b/src/ControlSystems.jl
@@ -110,8 +110,9 @@ abstract type AbstractSystem end
 
 include("types/TimeType.jl")
 ## Added interface:
-#   sampletime(Lti) -> Number
 #   time(Lti) -> TimeType (not exported)
+#   sampletime(Lti) = sampletime(time(Lti)) -> Number (exported)
+
 
 include("types/Lti.jl")
 
--- a/src/ControlSystems.jl
+++ b/src/ControlSystems.jl
@@ -87,6 +87,7 @@ export  LTISystem,
         denvec,
         numpoly,
         denpoly,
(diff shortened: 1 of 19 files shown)


mfalt:
Fixed through 40d355076888f3c40028f6a5280b219519924dd0

PR #308, opened 2020-05-27
Allow the C matrix to be specified as UniformScaling

imciner2:
This allows for the C matrix in a statespace system to be specified as the uniform scaling operator:
```julia
julia> StateSpace(A, B, I, 0)                                                                                                           
StateSpace{Float64,Array{Float64,2}}
A = 
 0.9159352448467247     1.2085157889104425
 0.027851984520770798  -0.17026615885088972
B = 
 -0.6021632825996087
  1.147941608828048
C = 
 1.0  0.0
 0.0  1.0
D = 
 0.0
 0.0

Continuous-time state-space model


julia> StateSpace(A, B, 2*I, 0)                                                                                                         
StateSpace{Float64,Array{Float64,2}}
A = 
 0.9159352448467247     1.2085157889104425
 0.027851984520770798  -0.17026615885088972
B = 
 -0.6021632825996087
  1.147941608828048
C = 
 2.0  0.0
 0.0  2.0
D = 
 0.0
 0.0

Continuous-time state-space model
```

This is useful if you want to treat all states as outputs (such as just making a system object with only provided A and B matrices).

### DIFF
--- a/src/types/StateSpace.jl
+++ b/src/types/StateSpace.jl
@@ -46,12 +46,19 @@ end
 
 const AbstractNumOrArray = Union{Number, AbstractVecOrMat}
 
-function StateSpace{T,MT}(A::AbstractNumOrArray, B::AbstractNumOrArray, C::AbstractNumOrArray, D::AbstractNumOrArray, Ts::Real) where {T, MT <: AbstractMatrix{T}}
+function StateSpace{T,MT}(A::AbstractNumOrArray, B::AbstractNumOrArray, C::Union{UniformScaling,AbstractNumOrArray}, D::AbstractNumOrArray, Ts::Real) where {T, MT <: AbstractMatrix{T}}
+    if typeof(C) <: UniformScaling
+        C = Matrix{T}(C*I, size(A, 1), size(A, 1))
+    else
+        C = to_matrix(T, C)
+    end
+
     if D == 0
         D = fill(zero(T), size(C,1), size(B,2))
     else
         D = to_matrix(T, D)
     end
+
     return StateSpace{T,Matrix{T}}(MT(to_matrix(T, A)), MT(to_matrix(T, B)), MT(to_matrix(T, C)), MT(D), Float64(Ts))
 end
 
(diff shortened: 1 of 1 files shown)


src/types/StateSpace.jl
```diff
@@ -46,29 +46,43 @@ end
 
 const AbstractNumOrArray = Union{Number, AbstractVecOrMat}
 
-function StateSpace{T,MT}(A::AbstractNumOrArray, B::AbstractNumOrArray, C::AbstractNumOrArray, D::AbstractNumOrArray, Ts::Real) where {T, MT <: AbstractMatrix{T}}
+function StateSpace{T,MT}(A::AbstractNumOrArray, B::AbstractNumOrArray, C::Union{UniformScaling,AbstractNumOrArray}, D::AbstractNumOrArray, Ts::Real) where {T, MT <: AbstractMatrix{T}}
+    if typeof(C) <: UniformScaling
+        C = Matrix{T}(C*I, size(A, 1), size(A, 1))
```
olof3, line 51:
```suggestion
        C = Matrix{T}(C, size(A))
```

olof3:
This could be useful.

Should this also be allowed for the `B` matrix? Much less useful though.

Perhaps add some tests as well.

mfalt:
I think if we want this we should do it consistently for A and B also. In which case it should be probably solved in something like a function `to_sized_matrix`, like we do for `to_matrix`.

olof3:
It seems weird to let the `B` and `C` matrices determine the size of `A`. Also, allowing this would need a lot of code to deal with cases such that `B = I`, `C = I`, while at the same time allowing `D = 0` to give an appropriately sized zero `D` matrix.

We should perhaps only have this for the `C` matrix where it is clearly useful. It is not *that* hard to take `I(size(A))` as the input argument for the `B` matrix.

imciner2:
I don't think letting the A matrix be a UniformScaling will be a good idea. Then we get a degenerate case of `StateSpace(I, I, I, 0)` - where there is no information on the matrix sizes. Forcing the A matrix to always be user specified is the best course. As for the `B` matrix, I guess that would be possible also, but it will require more diligence depending on `D`. I also just realized that the current code actually will break for `D` of incorrect size or not specified to be 0, so I need to fix that.

> I guess to_sized_matrix is basically a Matrix{T} wrapper. I think this should wait until we figure out how we want the dispatch for StateSpace and HeteroStateSpace, see for example the coming changes with the function to_abstract_matrix in PR #256

I don't see how the `to_abstract_matrix` will affect this. The conversion of `UniformScaling` to a matrix doesn't dispatch through any functions with `AbstractArray` in their type, it uses a special version of `Matrix` provided by LinearAlgebra (the `UniformScaling` is the root of its type hierarchy: https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/uniformscaling.jl#L29, so it must be dispatched on specifically or use `Any`).

src/types/StateSpace.jl
```diff
@@ -46,29 +46,43 @@ end
 
 const AbstractNumOrArray = Union{Number, AbstractVecOrMat}
 
-function StateSpace{T,MT}(A::AbstractNumOrArray, B::AbstractNumOrArray, C::AbstractNumOrArray, D::AbstractNumOrArray, Ts::Real) where {T, MT <: AbstractMatrix{T}}
+function StateSpace{T,MT}(A::AbstractNumOrArray, B::AbstractNumOrArray, C::Union{UniformScaling,AbstractNumOrArray}, D::AbstractNumOrArray, Ts::Real) where {T, MT <: AbstractMatrix{T}}
+    if typeof(C) <: UniformScaling
```
mfalt, line 50: I prefer `C isa UniformScaling`, but that is just a matter of taste I think, and I'm not sure if we have been consistent.

src/types/StateSpace.jl
```diff
@@ -46,29 +46,43 @@ end
...
     T = promote_type(eltype(A),eltype(B),eltype(C),eltype(D))
     A = to_matrix(T, A)
     B = to_matrix(T, B)
-    C = to_matrix(T, C)
+
+    if typeof(C) <: UniformScaling
+        C = Matrix{T}(C*I, size(A, 1), size(A, 1))
+    else
+        C = to_matrix(T, C)
+    end
```
baggepinnen, line 78:
```suggestion
    if C isa UniformScaling
        C = Matrix{T}(C*I, size(A, 1), size(A, 1))
    else
        C = to_matrix(T, C)
    end
```

Since this code is called twice (in method above as well), maybe it deserves it's own function?

PR #315, opened 2020-09-11
WIP: New types for handling discrete/continuous time

olof3:


olof3:
I hope this looks alright, @mfalt , @baggepinnen 

Travis does not seem to report the result (?!), although it ran on the server and 1.0 and 1.3 passed.
Perhaps we should add 1.5 to travis as well (it passes on my machine at least)

There are a number of errors on nightly

1. 6 errors from format changes when printing types
How should we handle this, have a switch for different julia versions? Moving all print tests to a separate file would perhaps make this easier to manage

2. 1 error probably from regression in @printf,
I have posted an issues

3. 1 failure from separating kwargs using comma instead of semicolon
This is done a lot throughout the code but it only resulted in one error (in `c2d` for some reason), no problems in 1.5.1. But I guess we should change this in general?

It was a bit strange though, here is a modified mwe,

```julia
julia> function c2d_mod(G::TransferFunction; kwargs...)
            @assert (size(G) == (1,1)) "c2d(G::TransferFunction, h) not implemented for MIMO systems"
            sysd = c2d(ss(G), 0.2, kwargs...)
        end
c2d_mod (generic function with 1 method)

julia> c2d_mod(tf(1))
ERROR: BoundsError: attempt to access () at index [1]
```

It works if the assertion is removed though?!



Perhaps easier to follow if this branch is pulled and then fix the errors on master, I guess all the errors are present there anyway.

PR #316, opened 2020-09-16
Changing to the propper "Plots.jl" names to avoid warnings

KronosTheLate:
So I have only changed "color" into "seriescolor" and "x/ylabel" into "x/yguide" where I could find them, inside the call to the ```@series``` macros. This is my first pull-request and commit, and so any feedback is appreciated. I don't really know what I am doing, but I hope and believe that I fixed it right ^_^

### DIFF
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -151,7 +151,7 @@ lsimplot
                 subplot --> s2i(1,i)
                 label     --> "\$G_{$(si)}\$"
                 linestyle --> styledict[:l]
-                color --> styledict[:c]
+                seriescolor --> styledict[:c]
                 t,  y[:, i]
             end
         end
@@ -211,12 +211,12 @@ for (func, title, typ) = ((step, "Step Response", Stepplot), (impulse, "Impulse 
                     ytext = (ny > 1 && j==1) ? "Amplitude to: y($i)" : "Amplitude"
                     @series begin
                         seriestype := style
-                        xlabel --> "Time (s)"
-                        ylabel --> ytext
+                        xguide --> "Time (s)"
+                        yguide --> ytext
                         subplot --> s2i(i,j)
                         label --> "\$G_{$(si)}\$"
                         linestyle --> styledict[:l]
-                        color --> styledict[:c]
+                        seriescolor --> styledict[:c]
                         t, ydata
                     end
                 end
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -586,7 +586,7 @@ nicholsplot
             styledict = getStyleSys(sysi,length(systems))
             linestyle --> styledict[:l]
             hover --> [Printf.@sprintf("ω = %.3f", w) for w in w]
-            color --> styledict[:c]
+            seriescolor --> styledict[:c]
             angles, mag
         end
     end
@@ -622,7 +622,7 @@ sigmaplot
                 xscale --> :log10
                 yscale --> _PlotScaleFunc
                 linestyle --> styledict[:l]
-                color --> styledict[:c]
+                seriescolor --> styledict[:c]
                 w, sv[:, i]
             end
         end


baggepinnen:
Nice, thanks :)
Can you please verify that you're able to change the color of, e.g., a `bodeplot` by using the keywords `c, color`? Like so
```julia
bodeplot(sys, c=:blue)
```
We've had problems with this before, that depending on what keyword we used in the plot recipe the results would be different when the user provided these magic keywords `c, color` etc.

KronosTheLate:
I copy-pasted the file I edited in the PR into my own package (Is that a legit move? It seems to work at least, ehe...), and ran both bodeplot and impulseplot with keywords c and color, everything works fine. No errors and colors as expected

However, when I run this code
```julia
L = 1.5
R = 1
C = 0.8
num = [1.0]
den = [L*C, R*C, 1]
TF = tf(num, den)
#mag, phase, w = bode(TF)
bodeplot(TF, color=:red)
```
I get 456 lines printed into REPL (in VSCode) saying ```ERROR: syntax error```. But the plot looks great :D Maybe I should open a separate issue for that?

A third, unrelated thing: how do I know what impulse my model is subjected to in an impulseplot, and how do I edit it? I saw nothing on it in the docs for impulseplot, it only mentions the ```final time Tf and discretization time Ts```

baggepinnen:
Alright great! The `ERROR: syntax error` appears with the GR backend and I think it's due to how we provide latex strings. It's a hassle to get rid of because the behavior of GR has changed so many times in the past that every time we fix it it breaks a few months later :(

The impulse is such that it has unit area but the shortest possible time. For discrete systems this means one sample period, for continuous systems this means a Dirac delta function. You can not change this, but you can provide arbitrary inputs to the function `lsim`. In practice, it will be hard to simulate "almost" impulses for continuous time systems. The current simulation relies on mathematical properties of a true Dirac impulse to perform the simulation. See https://github.com/JuliaControl/ControlSystems.jl/blob/d183cb1dff131bc12e1e9580310d9ba596cabe9f/src/timeresp.jl#L41 where the system is rewritten and a special initial condition is set

baggepinnen (approved):

KronosTheLate:
> Alright great! The `ERROR: syntax error` appears with the GR backend and I think it's due to how we provide latex strings. It's a hassle to get rid of because the behavior of GR has changed so many times in the past that every time we fix it it breaks a few months later :(

That is a shame, the LaTeX looks great. With the ployly backend it is just un-parsed latex, so that is also not great.

> The impulse is such that it has unit area but the shortest possible time. For discrete systems this means one sample period, for continuous systems this means a Dirac delta function. You can not change this, but you can provide arbitrary inputs to the function `lsim`.

I think I will add this to the doc-string, if you are okay with me using your words? Ofc, you will probably have the final say on it anyways, so you can be a filter if I write something stupid/inaccurate. I will try to say no more than you did.

>In practice, it will be hard to simulate "almost" impulses for continuous time systems. The current simulation relies on mathematical properties of a true Dirac impulse to perform the simulation. See
> 
> https://github.com/JuliaControl/ControlSystems.jl/blob/d183cb1dff131bc12e1e9580310d9ba596cabe9f/src/timeresp.jl#L41
> 
> where the system is rewritten and a special initial condition is set

I have to admit, there is a lot I don't understand. I am a third semester engineering student who happens to prefer Julia to MatLab and Simulink if I can (due to cost, and Julia syntax/comunity/speed are all so great), so I am checking out alternatives. I had pretty much no idea what a control-system, a transfer-function or the laplace-domain were a few short weeks ago (I still don't really KNOW what the laplace-domain is, but it sure is nice for math!).  I think I have to learn more before I can take full use of your package, which really looks like it can do a lot of nice things. Good job, and thank you for your nice and full answers 😄

(Is this where I close the issue?)

baggepinnen:
> (Is this where I close the issue?)

Please don't close, it's a PR ;)

I'll merge it once all tests have passed.

PR #320, opened 2020-09-23
Leaner promotion mechanism in evalfr

baggepinnen:
The old promotion code cause some issues with special types. It might also be the case that 
```
one(T0)*one(typeof(s))
prod/prod
```
is expensive to evaluate, in which case we would really like to avoid doing computations just to figure out the type.

I don't know why the old promotion code was so complicated, but the tests pass (locally at least)  with this new one. Anyone have strong feelings for the old promotion? Blame indicates mattias as the last author
https://github.com/JuliaControl/ControlSystems.jl/blame/7fc5ba4b0ab511b699b5c280e5e0e68c52fe07c0/src/freqresp.jl

### DIFF
--- a/src/freqresp.jl
+++ b/src/freqresp.jl
@@ -39,6 +39,11 @@ function _preprocess_for_freqresp(sys::StateSpace)
 end
 
 
+function _evalfr_return_type(sys::AbstractStateSpace, s::Number)
+    temp_product = one(T0)*one(typeof(s))
+    typeof(temp_product/temp_product)
+end
+
 """
 `evalfr(sys, x)` Evaluate the transfer function of the LTI system sys
 at the complex number s=x (continuous-time) or z=x (discrete-time).
@@ -47,7 +52,7 @@ For many values of `x`, use `freqresp` instead.
 """
 function evalfr(sys::StateSpace{<:TimeEvolution,T0}, s::Number) where {T0}
     prod = one(T0)*one(typeof(s))
-    T = typeof(prod/prod)
+    T = _evalfr_return_type(sys, s)
     try
         R = s*I - sys.A
         sys.D + sys.C*((R\sys.B)::Matrix{T})  # Weird type stability issue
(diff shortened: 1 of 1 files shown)


src/freqresp.jl
```diff
@@ -46,7 +46,8 @@ at the complex number s=x (continuous-time) or z=x (discrete-time).
 For many values of `x`, use `freqresp` instead.
 """
 function evalfr(sys::StateSpace{<:TimeEvolution,T0}, s::Number) where {T0}
-    T = promote_type(T0, typeof(one(T0)*one(typeof(s))/(one(T0)*one(typeof(s)))))
+    prod = one(T0)*one(typeof(s))
+    T = typeof(prod/prod)
```
olof3, line 50: This looks much better! However `prod` doens't seem like an ideal variable name (the name of a quite common function). Perhaps it is slightly better to just call it `tmp`-

But since I guess we would like to handle structured statespace representations, and this will be more messy, it would be even better to put this functionality in a function like 

`_evalfr_return_type(sys;;AbstractStateSpace, s)`

Perhaps the cleanest approach would be to have a function that computes the inverse and returns an `Inf` matrix if it fails. I think this would avoid having to establish the return type and the try catch statement in this function?

olof3:
Is there some test case for when it breaks? Or is it just slower?

baggepinnen:
> Is there some test case for when it breaks? Or is it just slower?

I have a test case with a very complex type from another package, I did not manage to distill it down to a simple test case. 
The case when it can be slower is if the numbers are containing 10000 samples, then it's unnecessary to perform a bunch of products and divisions just to get the type.

src/freqresp.jl
baggepinnen: Yeah I don't like the try-catch since it catches other errors as well. I was quite confused for a while this morning due to exactly this.

olof3: The changes look good, we can think more about this if we get around to work on the `Hessenberg` stuff

> Perhaps the cleanest approach would be to have a function that computes the inverse and returns an Inf matrix if it fails. I think this would avoid having to establish the return type and the try catch statement in this function?

PR #336, opened 2020-09-30
Change default framestyle of pzmap-plots to zerolines

KronosTheLate:
I think that pole-zeroes maps look so so so much better with the axis centered on zero, as is provided by changing the framestyle to ```:zerolines```. I simply find origin paramount to be able to orient myself on the plot, and the lines through (0, 0) of course locate the origin visually. Plus, in pole-zero-plots, the points' position relative to the origin is one of the key points of interest ( as far as I understand the plot ). This PR changes zerolines to the default. I have added a couple of pictures blow, one with zerolines and one without, so that anyone who sees this can chime in if they agree/disagree that this change should be made.
pzmap_standard
pzmap_zerolines


olof3:
Definitely an improvement! Thanks for fixing this.

Some other suggestions if you have the time and agree.
Perhaps it would be good to also use equal scaling of the axes by default?
Shouldn't the x axis be in the middle vertically? (for systems with systems with real coefficient :))
The x's for the poles could also look better.

src/plotting.jl
```diff
@@ -724,6 +724,7 @@ pzmap
         @warn("pzmap currently only supports SISO systems. Only transfer function from u₁ to y₁ will be shown")
     end
     seriestype := :scatter
+    framestyle := :zerolines
```
baggepinnen, line 727:
```suggestion
    framestyle --> :zerolines
```
This must use the arrow Syntax to make it a default rather than to force it.

baggepinnen:
LGTM after changing to default rather than forced as the suggestion above. I'm split on the switch to equal axis, all my systems are always extremely lightly damed, but I agree that you get more usable information put of the plot if the axes have the same scale, so I'm slightly in favor of Olofs suggestion as well.

src/plotting.jl
KronosTheLate: Nice, good to know. So you fixed it now, and the PR is updated with your fix? Or do I have to do anything about it?

I gotta admit, I don't understand these plot-recipes with macros. I guess it is for defining defaults, forcing, etc in a package, whereas in a script one would just pass it all to the plot-call? A short description or pointer to documentation on why one uses things like macros, and ```-->```/```:=```  in the recipes for this package instead of just ```=```, would be nice. I currently don't get it at all...

KronosTheLate:
> Perhaps it would be good to also use equal scaling of the axes by default?
I dont think that would be the case TBH. That would be nice if the real and imaginary poles/zeroes are of a similar order of magnitude, but that is no guarantee, right? And if they are very different, it would be difficult to see the value for the smaller one on the axis. That would necessitate hover-to-see-data, like from plotly. The default GR, without the ability to hover data points, could really make this feature a pain in the butt I think.

> Shouldn't the x axis be in the middle vertically? (for systems with real coefficient :))
If all of the points are to the left, I think that there is no need to see the right side. It is nice for symmetry, but bad for seeing the x-value for the data-point. I would prefer the plot as is in the picture regarding this feature...

> The x's for the poles could also look better.
Agreed. I was thinking maybe smaller in width (half-ish) and quite the bit thicker (2-3 times)? Also maybe red poles and black zeroes, so that if they are on top of each other it is clearly visible. What changes would you make?

baggepinnen:
> The x's for the poles could also look better.
> Agreed. I was thinking maybe smaller in width (half-ish) and quite the bit thicker (2-3 times)? Also maybe red poles and black zeroes, so that if they are on top of each other it is clearly visible. What changes would you make?

One major inconvenience is that different plot backends display the same markers differently, they also handle colors slightly differently. I general, I think it's good to not deviate too much from the defaults, since the appearance has a tendency to break (i.e. change with updates of one of the plotting packages) if you do.

In any case, this PR is nice and limited in it's scope, so I would consider it good for merge. Potential changes to the marker visuals can be added in a new PR.

@KronosTheLate, don't forget to base PR:s off the updated master branch of the base package, i.e., pull ControlSystems/master into master on your fork. This way only the commits that are new appear in the PR, and not the commits from your previous PR.

olof3:
> I don't think that would be the case TBH. That would be nice if the real and imaginary poles/zeroes are of a similar order of magnitude, but that is no guarantee, right? And if they are very different, it would be difficult to see the value for the smaller one on the axis. That would necessitate hover-to-see-data, like from plotly. The default GR, without the ability to hover data points, could really make this feature a pain in the butt I think.

Yes, that is a major inconvenience of that many of the backends don't have hover-to-see-data or zooming. On the other hand it is also extremely annoying that the axes are scaled differently.  There is also the question of if pzmap is really intended to show exact pole/zero location or to give an overview of the system dynamics? With proper hover-to-see-data and zooming this would be a non-issue, but I absolutely agree with your point, so let's keep it like this for now.


> If all of the points are to the left, I think that there is no need to see the right side. It is nice for symmetry, but bad for seeing the x-value for the data-point. I would prefer the plot as is in the picture regarding this feature...

I believe I was talking about the other axis :)

KronosTheLate:
> @KronosTheLate, don't forget to base PR:s off the updated master branch of the base package, i.e., pull ControlSystems/master into master on your fork. This way only the commits that are new appear in the PR, and not the commits from your previous PR.

Right, I was wondering about that. I will keep it in mind for next time, thanks for the feedback

> I believe I was talking about the other axis :)

Oh. right. Well, I think that all the information on the negative imaginary axis is redundant, as every point is simply the complex conjugate of a point on the positive imaginary axis (for realizable systems). Therefore, the overview is actually better as is, because you have all the information and more zoomed in. I think that only showing the "top half" is therefore common practice, and I like it. But perhaps an optional argument like ```zoom=full``` vs ```zoom=half``` or something like that is a good idea? Not that I would use it, but maybe you would?

baggepinnen:
> I think that all the information on the negative imaginary axis is redundant, as every point is simply the complex conjugate of a point on the positive imaginary axis (for realizable systems). 

While this is true, most pzmaps in the literature are displayed showing both sides, so that's probably what most people are expecting (including me). I tried for a while to hide the negative imaginary axis in all plots in another package but ended up reverting since it confused me and I never got used to it.

olof3:
> > I think that all the information on the negative imaginary axis is redundant, as every point is simply the complex conjugate of a point on the positive imaginary axis (for realizable systems).
> 
> While this is true, most pzmaps in the literature are displayed showing both sides, so that's probably what most people are expecting (including me). I tried for a while to hide the negative imaginary axis in all plots in another package but ended up reverting since it confused me and I never got used to it.

I agree with all of this. Probably good to keep to the standard though, the lack of symmetry in the example plot makes it look a bit off.

PR #339, opened 2020-10-01
Get rid of the style magic in plotting

baggepinnen:
This PR removes all the magic to select colors and styles for plots, which had a number of problems
```julia
bodeplot(tf(1,[1,1]))
bodeplot!(tf(1,[2,1]))
```
would produce two plots with the same color, as the colors were chosen based on the number of input systems to the plot call.
This is instead replaced by the, now robust, default color selection in Plots. The code above now plots both phase and magnitude in the same color (by using the `group` keyword), but the color is different for the first system compared to the second.

Both calls below will also produce the desired result, the first one plots two systems with different color in the same plot, whereas the second call plots a MISO system in a 2x2 grid.
```julia
bodeplot([tf(1,[1,1]), tf(1,[2,1])])
bodeplot([tf(1,[1,1])  tf(1,[2,1])])
```


mfalt:
@JuliaControlBot test-plots

baggepinnen:
@JuliaControlBot test-plots

baggepinnen:
I think it looks good. The only possible thing to consider is if we want both curves in the sigmaplot have the same color?

mfalt:
> I think it looks good. The only possible thing to consider is if we want both curves in the sigmaplot have the same color?

Not sure what makes most sense, I guess there is no point in being able to identify each of them?

baggepinnen:
@JuliaControlBot test-plots

PR #343, opened 2020-10-05
c2d for delaysystems

olof3:
Basic functionality for when the delays are multiples of the sample time.

For the general case the situation seems a bit tricky, so we have to use some kind of approximation. Either just rounding the delays or preferably being slightly more sophisticated (I guess by extending the approach in Ch 2 of Computer Controlled Systems (Åström & Wittenmark)). Also matlab seems to use some sort of approximation.


baggepinnen (approved):
LGTM 😃

PR #346, opened 2020-10-06
Let impulse handle it's own args

### LINKED ISSUE #345: impulse/stepplot ignore keyword arguments
The problem is that these plots are defined using metaprogramming, looping over the different functions. Since the functions take different kwargs, none are provided a.t.m.

baggepinnen:
Instead of handling different function signatures in the plot recipe, this PR let's the function itself handle the function signature.
Closes #345


baggepinnen:
@mfalt How do you trigger the plot testing bot?

baggepinnen:
@JuliaControlBot test-plots

mfalt:
Testing should work now, I will just update the references, and then call the bot.

baggepinnen:
@JuliaControlBot test-plots

mfalt:
This seams to be a real bug in the plots, i think it is this line where it first breaks
```julia
tf1 = tf([1],[1,1])
tf2 = tf([1/5,2],[1,1,1])
sys = [tf1 tf2]
ts = 0:0.01:5
stepplot(sys, ts[end], ts[2]-ts[1], l=(:dash, 4))
```

baggepinnen:
I created a companion PR https://github.com/JuliaControl/ControlExamplePlots.jl/pull/12 that should resolve the test error.

mfalt:
Ok, if the plot functions are consistent with the non-plot functions, that's good. We should add a news entry before we tag.

baggepinnen:
It's technically a breaking change, maybe I should add a news entry as well as bump the version in this PR and then we can tag a new release after merging this

mfalt:
I think we can pull this once we update the documentation for stepplot and lsimplot, currently they say
```
impulseplot(sys[, tfinal[,  dt]])

  Plot step response of sys with optional final time tfinal and discretization time dt. If not
  defined, suitable values are chosen based on sys.
```
I guess it should be
```
impulseplot(sys[, Tf]; kwargs...) or impulseplot(sys[, t]; kwargs...)
```
according to 
```
y, t, x = impulse(sys[, Tf]) or y, t, x = impulse(sys[, t])

  Calculate the impulse response of system sys. If the final time Tf or time vector t is not
  provided, one is calculated based on the system pole locations.
```

mfalt:
There are more bugs:
```julia
stepplot([P,Gcl]) # Visualize the open and closed loop responses.
```
crashes in example.md

baggepinnen:
Hmm, I've been confused by how the plotting of vectors of systems is supposed to work. Can't it be handled using machinery built into Plots? What's the benefit of having a plot method accepting a vector of systems rather than just plotting each system, or broadcasting the plot function over a vector of systems?

mfalt:
> Hmm, I've been confused by how the plotting of vectors of systems is supposed to work. Can't it be handled using machinery built into Plots? What's the benefit of having a plot method accepting a vector of systems rather than just plotting each system, or broadcasting the plot function over a vector of systems?

I guess the goal is to have a convinient syntax, as in the readme. But maybe broadcast is the way to go now.

PR #348, opened 2020-10-07
Make time vector to lsim optional

baggepinnen:
This PR adds a convenience method that supplies a default time vector to `lsim` for discrete systems when a vector of inputs is supplied.


olof3:
Excellent!

mfalt:
@JuliaControlBot test-plots

PR #352, opened 2020-10-09
WIP: Add basic logic to avoid delays in series for DelayLtiSystem

albheim:
There was an issue where delay(1)*delay(1) generated a system which could not be simulated since it has a D22 matrix which is non-zero.

For the moment my goal was just to see if I could get it to work with delay(1)*delay(1) being simplified to the same as delay(2), and it seems to work. I know the code is not very nice, but wanted some input on if this is at all useful or compatible with how the rest works before I spend more time on it.

It basically just checks if there is any delay which only depends on a single other delay, and nothing else depends on that delay.

### DIFF
--- a/src/types/DelayLtiSystem.jl
+++ b/src/types/DelayLtiSystem.jl
@@ -118,7 +118,7 @@ Base.size(sys::DelayLtiSystem) = (noutputs(sys), ninputs(sys))
 # Fallbacks, TODO We should sort this out for all types, maybe after SISO/MIMO
 # {Array, Number}, Colon
 Base.getindex(sys::DelayLtiSystem, i, ::Colon) =
-    getindex(sys, index2range(i), 1:size(sys,2))
+    getindex(sys, index2range(i), 1:size(sys,2g;))
 # Colon, {Array, Number}
 Base.getindex(sys::DelayLtiSystem, ::Colon, j) =
     getindex(sys, 1:size(sys,1), index2range(j))
@@ -172,6 +172,35 @@ end
 function *(sys1::DelayLtiSystem, sys2::DelayLtiSystem)
     psys_new = sys1.P * sys2.P
     Tau_new = [sys1.Tau; sys2.Tau]
+
+    # Check for delays that are only dependent on other delays and reduce them
+    # TODO @views 
+    for i in 1:size(psys_new.C2, 1)
+        # Check if delay signal i is only update from other delay, if so call that delay j
+        if all(psys_new.C2[i, :] .== 0) && all(psys_new.D21[i, :] .== 0) && sum(psys_new.D22[i, :] .!= 0) == 1
+            j = findfirst(psys_new.D22[i, :] .!= 0)
+            # Check if delay j is not used to update anything but delay i
+            if all(psys_new.B2[:, j] .== 0) && all(psys_new.D12[:, j] .== 0) && sum(psys_new.D22[:, j] .!= 0) == 1
+                Tau_new[j] += Tau_new[i]
+                Tau_new = [Tau_new[1:i-1]; Tau_new[i+1:end]]
+
+                i = i + psys_new.ny1
+                j = j + psys_new.nu1
+                if i < j
+                    B = view(psys_new.P.B, :, [1:i-1; i+1:j-1; i; j+1:size(psys_new.P.B, 2)])
+                    D = view(psys_new.P.D, [1:i-1; i+1:size(psys_new.P.C, 1)], [1:i-1; i+1:j-1; i; j+1:size(psys_new.P.B, 2)])
+                else
+                    B = view(psys_new.P.B, :, [1:j-1; i; j+1:i-1; i+1:size(B2, 2)])
+                    D = view(psys_new.P.D, [1:i-1; i+1:size(psys_new.P.C, 1)], [1:j-1; i; j+1:i-1; i+1:size(psys_new.P.B, 2)])
+                end
+                C = view(psys_new.P.C, [1:i-1; i+1:size(psys_new.P.C, 1)], :)
+
+                P = StateSpace(psys_new.P.A, B, C, D, psys_new.P.timeevol)
+                psys_new = PartionedStateSpace{StateSpace{Continuous, Float64, Matrix{Float64}}}(P, psys_new.nu1, psys_new.ny1)
+                break
+            end
+        end
+    end
 
     DelayLtiSystem(psys_new.P, Tau_new)
 end
--- a/src/types/DelayLtiSystem.jl
+++ b/src/types/DelayLtiSystem.jl
@@ -118,7 +118,7 @@ Base.size(sys::DelayLtiSystem) = (noutputs(sys), ninputs(sys))
 # Fallbacks, TODO We should sort this out for all types, maybe after SISO/MIMO
 # {Array, Number}, Colon
 Base.getindex(sys::DelayLtiSystem, i, ::Colon) =
-    getindex(sys, index2range(i), 1:size(sys,2g;))
+    getindex(sys, index2range(i), 1:size(sys,2))
 # Colon, {Array, Number}
 Base.getindex(sys::DelayLtiSystem, ::Colon, j) =
     getindex(sys, 1:size(sys,1), index2range(j))
--- a/src/types/DelayLtiSystem.jl
+++ b/src/types/DelayLtiSystem.jl
@@ -168,41 +168,60 @@ end
 -(sys1::DelayLtiSystem, sys2::DelayLtiSystem) = +(sys1, -sys2)
 -(sys::DelayLtiSystem{T}) where {T} = *(sys, T(-1))
 
+""" 
+    reducedelays(sys::DelayLtiSystem)    
+
+Checks if `sys` has any delays in series that can be reduced to a single delay.
+Returns a new system with reductions applied.
+"""
+function reducedelays(sys::DelayLtiSystem)
+    # Should I copy?
+    B, C, D = sys.P.B, sys.P.C, sys.P.D
+    Tau = copy(sys.Tau)
+    # Check for delays that are only dependent on other delays and reduce them
+    i = sys.P.ny1
+    while i < size(C, 1)
+        i += 1
+        # Check if delay signal i is only update from other delay, if so call that delay j
+        if count(!=(0), view(C, i, :)) == 0 && count(!=(0), view(D, i, :)) == 1 
+            j = findfirst(!=(0), view(D, i, :))
+            # Check if delay j is not used to update anything but delay i
+            if j > sys.P.nu1 && count(!=(0), view(B, :, j)) == 0 && count(!=(0), view(D, :, j)) == 1 
+                scale = D[i,j] # Save this to later rescale with
+
+                i2 = i - sys.P.ny1 + sys.P.nu1 # Column of i
+                j2 = j - sys.P.nu1 + sys.P.ny1 # Row of j
+
+                # Swap i2 and j, drop what is on i2 after swap (i.e. old j column)
+                if i2 < j
+                    B = [view(B, :, [1:i2-1; i2+1:j-1])                          scale * view(B, :, i2)                          view(B, :, j+1:size(B,2))]
+                    D = [view(D, [1:i-1; i+1:size(D,1)], [1:i2-1; i2+1:j-1])     scale * view(D, [1:i-1; i+1:size(D,1)], i2)     view(D, [1:i-1; i+1:size(D, 1)], j+1:size(D,2))]
+                else
+                    B = [view(B, :, 1:j-1)                         scale * view(B, :, i2)                         view(B, :, [j+1:i2-1; i2+1:size(B,2)])]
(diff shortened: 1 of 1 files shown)


src/types/DelayLtiSystem.jl
```diff
@@ -173,6 +173,35 @@ function *(sys1::DelayLtiSystem, sys2::DelayLtiSystem)
     psys_new = sys1.P * sys2.P
     Tau_new = [sys1.Tau; sys2.Tau]
 
+    # Check for delays that are only dependent on other delays and reduce them
+    # TODO @views 
+    for i in 1:size(psys_new.C2, 1)
+        # Check if delay signal i is only update from other delay, if so call that delay j
+        if all(psys_new.C2[i, :] .== 0) && all(psys_new.D21[i, :] .== 0) && sum(psys_new.D22[i, :] .!= 0) == 1
+            j = findfirst(psys_new.D22[i, :] .!= 0)
```
baggepinnen, line 181: This returns `j = nothing` if nothing is found, if that can happen it's probaby best to check for it.

src/types/DelayLtiSystem.jl
```diff
@@ -118,7 +118,7 @@ Base.size(sys::DelayLtiSystem) = (noutputs(sys), ninputs(sys))
 # Fallbacks, TODO We should sort this out for all types, maybe after SISO/MIMO
 # {Array, Number}, Colon
 Base.getindex(sys::DelayLtiSystem, i, ::Colon) =
-    getindex(sys, index2range(i), 1:size(sys,2))
+    getindex(sys, index2range(i), 1:size(sys,2g;))
```
mfalt, line 121: Typo?

src/types/DelayLtiSystem.jl
albheim: I think it should not be possible to not have something not equal zero if we come to this place in the code since the if statement above has a clause with `sum(psys_new.D22[i, :] .!= 0) == 1` right before. 

With that said, I think I will rewrite quite a bit of this if it is actually going into the repo so it might very well not be an issue later on.

src/types/DelayLtiSystem.jl
albheim: Yup, some vim stuff. 
g; is jump to last edited spot so I assume that is what I wanted. Not sure how I missed that I inserted it...

src/types/DelayLtiSystem.jl
```diff
@@ -173,6 +173,35 @@ function *(sys1::DelayLtiSystem, sys2::DelayLtiSystem)
...
+                    B = view(psys_new.P.B, :, [1:i-1; i+1:j-1; i; j+1:size(psys_new.P.B, 2)])
+                    D = view(psys_new.P.D, [1:i-1; i+1:size(psys_new.P.C, 1)], [1:i-1; i+1:j-1; i; j+1:size(psys_new.P.B, 2)])
+                else
+                    B = view(psys_new.P.B, :, [1:j-1; i; j+1:i-1; i+1:size(B2, 2)])
+                    D = view(psys_new.P.D, [1:i-1; i+1:size(psys_new.P.C, 1)], [1:j-1; i; j+1:i-1; i+1:size(psys_new.P.B, 2)])
+                end
+                C = view(psys_new.P.C, [1:i-1; i+1:size(psys_new.P.C, 1)], :)
+
+                P = StateSpace(psys_new.P.A, B, C, D, psys_new.P.timeevol)
+                psys_new = PartionedStateSpace{StateSpace{Continuous, Float64, Matrix{Float64}}}(P, psys_new.nu1, psys_new.ny1)
```
albheim, line 199: Is it just me or should this be PartitionedStateSpace, not PartionedStateSpace?

mfalt: I think you are right, but for simplicity we should fix that in a separate PR

albheim: Agreed, just brought it up here cause it was easier.

olof3: Yes, I also realized quite recently that I had done that typo. Hadn't really noticed it before.

olof3:
Yes, something like this would be a very useful feature. How to do it in a reasonably clean and efficient way does not seem obvious. Perhaps we would like `DelayLtiSystem` s to keep track of which delays that are input and output delays and deal with it before the multiplication. I guess this would keep the representations minimal in most cases. Should we aim for finding even more  simplifications? Is it easy/efficient to handle this without keeping track of input/output delays?

This is also an issue for the `feedback` function.

My original plan was to make the implementation as basic as possible and I was hoping that it would be possible to live with occasional delays in series, but it is certainly a problem if DelayDiffEq cannot deal with it (have we checked that this feature has not been added? has it been requested/discussed? that would make it less urgent with a fix for this issue). 

Are we sure that we always would like to merge delays when possible? I think yes.

@mfalt @baggepinnen Any suggestions for where to put that document with notes on implementation of delay systems? In this repo with tex files? In the wiki?

Sometime soon it is time with an overhaul of the delay functionality. For example, much of the interconnection of delay systems could be dealt with using the somewhat recently introduced general `feedback` function. This would avoid a lot of complicated matrix manipulations, but I'm not sure how it would hurt performance.

albheim:
> Yes, something like this would be a very useful feature. How to do it in a reasonably clean and efficient way does not seem obvious. Perhaps we would like `DelayLtiSystem` to keep track of which delays that are input and output delays and deal with it before the multiplication. I guess this would keep the representations minimal in most cases. Should we aim for finding even more simplifications? Is it easy/efficient to handle this without keeping track of input/output delays?

Right now if we take for example the system `delay(1)*delay(1)*tf([1], [1, 1])*delay(1)`, my code will reduce the two first delays, but not the second one since there is no direct connection between delay and delay in the D22 matrix, instead it does one 1s delay then through the state and then through a 2s delay before output. Is it desired to keep them apart or do we want to reduce that also?

Ran `@benchmark delay(1)*delay(1)` both with the reduction and without and it made twice the allocs and took 3x more time approximately. Didn't test with larger systems.

> I believe that this issue can actually be solved in a reasonably convenient manner using that general feedback function, but I'm not sure when I will have the time.

Not sure what this is about, but if you have an idea of how it should be done better I could try to take a look at it.

src/types/DelayLtiSystem.jl
```diff
@@ -168,41 +168,60 @@ end
...
         end
     end
+    P = StateSpace(copy(sys.P.A), B, C, D, timeevol(sys))
+    P = PartionedStateSpace(P, sys.P.nu1, sys.P.ny1)
+    DelayLtiSystem(P, Tau)
+end
 
-    DelayLtiSystem(psys_new.P, Tau_new)
+function *(sys1::DelayLtiSystem, sys2::DelayLtiSystem)
+    psys_new = sys1.P * sys2.P
```
mfalt, line 221: @olof3 We should probably document what this means? It says connecting in series, but it is just connecting the first output of sys2 to the first input to sys1.
I.e. for delay systems  sys2.y to sys1.u right, and letting the inputs be `[sys1.u; sys1.d; sys2.d]` and the outputs `[sys1.y; sys1.d; sys2.d]`.
This should probably not be called `*` or `series`, but rather `series_1_1` (as in 1 to 1) or something like that.

olof3: When the partitioned statespaces are part of lower LFTs (which is sort of implicit in a delay systems context), then it is not ambiguous, but that is certainly clear.

I think it could be good to add a general multiplication function that could be utilized here.

I think the right solution is a major overhaul of partitioned statespaces and related features, but that will require significant work and thought.

A temporary comment about that partitioned statespaces are considered as parts of lower LFTs could be added if we can't live with this until it gets fixed. For now, `series_1_1` is not too bad either, but I'm not sure what a decent future solution would look like.

olof3:
> > I believe that this issue can actually be solved in a reasonably convenient manner using that general feedback function, but I'm not sure when I will have the time.
> 
> Not sure what this is about, but if you have an idea of how it should be done better I could try to take a look at it.

I think the right approach is to deal with delays in series before multiplying the partitioned statespaces. But I think that would be  sort of a major overhaul to get everything in place. I also think that you would need to be a bit careful about instances where you  would like the signal in between two delays as an output.


> Right now if we take for example the system `delay(1)*delay(1)*tf([1], [1, 1])*delay(1)`, my code will reduce the two first delays, but not the second one since there is no direct connection between delay and delay in the D22 matrix, instead it does one 1s delay then through the state and then through a 2s delay before output. Is it desired to keep them apart or do we want to reduce that also?

In conclusion
I don't think that we shouldn't worry about those, it probably take quite a bit of effort and the states will no longer correspond to the same things.

> There might be other approaches that I am not thinking of.
Edit: It might not be too hard to implement manualy solving the problem in steps of tau but we would probably lose a lot of the nice magic in DDE.jl with interpolations and the ability to choose solvers that preserves a certain order of the solution.

Some sort of specialized version for linearized systems could be possible (although not too trivial). Introducing discrete-time delay systems with delay buffers would probably be easier and good for other reasons as well.

olof3:
> > I believe that this issue can actually be solved in a reasonably convenient manner using that general feedback function, but I'm not sure when I will have the time.
> 
> Not sure what this is about, but if you have an idea of how it should be done better I could try to take a look at it.

I think the right approach is to deal with delays in series before multiplying the partitioned statespaces. But it would be sort of a major overhaul to get everything in place. I guess one would also need to be a bit careful about instances where you  would like the signal in between two delays as an output.

I guess that the proposed approach is not the way we would eventually like to do this, but feel free to pursue it a bit further. Anyway, a lot of thanks for bringing up the issue, I think the current discussion is very valuable, and we should discuss the way forward in more detail. I would be more happy to see more work going into the last thing I talk about in this comment and handling `c2d` for delay systems where the delays are not multiples of the sample times (which is sort of a prerequisite for that to work in general).


> Right now if we take for example the system `delay(1)*delay(1)*tf([1], [1, 1])*delay(1)`, my code will reduce the two first delays, but not the second one since there is no direct connection between delay and delay in the D22 matrix, instead it does one 1s delay then through the state and then through a 2s delay before output. Is it desired to keep them apart or do we want to reduce that also?

I don't think that we shouldn't worry about those (at least not for now), it probably take quite a bit of effort and the states will no longer correspond to the same things.

> There might be other approaches that I am not thinking of.
Edit: It might not be too hard to implement manualy solving the problem in steps of tau but we would probably lose a lot of the nice magic in DDE.jl with interpolations and the ability to choose solvers that preserves a certain order of the solution.

Some sort of specialized version for linearized systems could be possible (although not too trivial).

Introducing discrete-time delay systems with delay buffers would probably be easier and is something that I would like to see for other reasons as well. That would also solve that simulation issue.

albheim:
> > Not sure what this is about, but if you have an idea of how it should be done better I could try to take a look at it.
> 
> I think the right approach is to deal with delays in series before multiplying the partitioned statespaces. But it would be sort of a major overhaul to get everything in place. I guess one would also need to be a bit careful about instances where you would like the signal in between two delays as an output.

I see the value of maybe just analyzing input and output delays and removing them in the combination step instead of doing it after combination, though this might need different analysis depending on how they are combined. But maybe it is only multiplication and feedback that can introduce new chained delays? If so it seems one could assume each system is already free of chained delays and only check for input and output delays and reduce them in the combination step for both of them without too much difficulty. 

I could try to look at this if you want. But it also seems @olof3 has some idea of a bigger rework so I'm not sure if that is preferrable?

albheim:
Noticed that in the PartitionedStateSpace multiplication routine we use the `getproperty` (e.g. `s1.C1`) that seems to creates a copy of the part of `s1.C` corresponding to C1, and we call it multiple times. Would this not create an unnecessary amount of allocations? Should we not just create one variable `s1C1 = view(s1.P.C, 1:s1.ny1, : )` and use that throughout the calculation? 

I tested this and it seems the view approach is not faster (at least not for reasonably small systems) but just doing `s1C1 = s1.P.C1` in the start and use that seems to give fewer allocations and slightly better time. Not sure if it is worth the change, but thought I would just point it out. 

Is there a reason why we don't just have partitioned state space keep track of its own matrices since it seems we never really want to access B, C or D but always want the partitioned view?

olof3:
> Noticed that in the PartitionedStateSpace multiplication routine we use the getproperty (e.g. s1.C1) that seems to creates a copy of the part of s1.C corresponding to C1, and we call it multiple times. Would this not create an unnecessary amount of allocations? Should we not just create one variable s1C1 = view(s1.P.C, 1:s1.ny1, : ) and use that throughout the calculation?

> I tested this and it seems the view approach is not faster (at least not for reasonably small systems) but just doing s1C1 = s1.P.C1 in the start and use that seems to give fewer allocations and slightly better time. Not sure if it is worth the change, but thought I would just point it out.

The way I did this is certainly not optimal. What you suggest is pretty much what I did in the general feedback function, and that was certainly better performance wise. I see little point in changing this now, since I'm not sure that we should keep the `PartitionedStateSpace` in its current form

 > Is there a reason why we don't just have partitioned state space keep track of its own matrices since it seems we never really want to access B, C or D but always want the partitioned view?

Perhaps. But there would be an awful lot of fields in that type, and this would  make some things a bit more messy. But this option definitely has some advantages. Again, the future of `PartitionedStateSpace` is a bit uncertain.

albheim:
I'm closing this since it is solved in a much nicer and more complete way by #377

PR #356, opened 2020-10-10
Only support julia v1.5

baggepinnen:
This PR changes the compatibility of Julia to v1.5, instead of the previous 1.0. The motivation for this is that it takes too much developer effort to maintain support for old julia versions.

- testing a package takes time, especially since most people do not have anything but the latest version installed locally.
- opening the build logs on CI takes time, figuring out what went wrong on an older release takes a lot of time.
- fixing things that don’t work on an old version of Julia takes a ton of time. Especially if one of your dependencies has been changed and is not compatible with earlier Julia.

This would mean that we can also get rid of some ugly version-specific code in this package, and that tests will go a bit faster. This also gives us access to things like
- Thread support
- Easier forwarding of keyword arguments with same name
- 5-arg `mul!`
- `eachrow, eachcol` etc.

On the whole, I think most people would benefit from a faster development cycle rather than new features being added for those who linger on julia v1.0. This also appears to be the mindset of the rest of the ecosystem, most packages drop support for older versions without hesitation. Only well backed packages with lots of maintainers strive for compat with 1.0. 
The following have all ditched 1.0 on the main branch
- [DifferentialEquations.jl](https://github.com/SciML/DifferentialEquations.jl/blob/master/Project.toml)
- [Flux](https://github.com/FluxML/Flux.jl/blob/master/Project.toml)
- [Plots](https://github.com/JuliaPlots/Plots.jl/blob/master/Project.toml)

Failures on old Julia versions but not on latest:
- https://travis-ci.com/github/JuliaControl/ControlSystems.jl/builds/189161877
- https://travis-ci.com/github/JuliaControl/ControlSystems.jl/builds/184390032
- https://travis-ci.com/github/JuliaControl/ControlSystems.jl/builds/184390299
- https://github.com/JuliaControl/ControlSystems.jl/pull/364/checks?check_run_id=1238242213
- https://github.com/JuliaControl/ControlSystems.jl/pull/408#issuecomment-744310939

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: julia
 julia:
-  - 1.0
   - 1
   - nightly
 matrix:
--- a/Project.toml
+++ b/Project.toml
@@ -29,7 +29,7 @@ MacroTools = "0.5"
 OrdinaryDiffEq = "5.2"
 Plots = "0.24, 0.25, 0.26, 0.27, 0.28, 0.29, 1.0"
 Polynomials = "1.0"
-julia = "1.0"
+julia = "1.5"
 
 [extras]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -31,25 +31,12 @@ to_abstract_matrix(A::Vector) = reshape(A, length(A), 1)
 to_abstract_matrix(A::Number) = fill(A, 1, 1)
 
 # Do no sorting of eigenvalues
-@static if VERSION > v"1.2.0-DEV.0"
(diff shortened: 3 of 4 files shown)


KronosTheLate:
If you want opinions, here is mine:

I dont care much for supporting 1.0, especially at the cost of development-time. Also, if other important packages are dropping 1.0 support, users should be on the updated version of julia anyways.

imciner2:
Considering how much they have diverged, this makes sense, and there has been talk about 1.6 becoming the next LTS release so a new LTS probably isn't that far away (1.6 would go into LTS when they branch 1.7 is what I have heard).

albheim:
Since #470 we now support 1.6 and up, so I assume we could now merge the other changes in this PR that removes some version specific code?

baggepinnen:
Yep, sounds good

PR #357, opened 2020-10-10
Add a PR formatting bot

baggepinnen:
This PR adds a bot that runs JuliaFormatter on each PR, borrowed from 
https://github.com/JuliaReinforcementLearning/ReinforcementLearningZoo.jl/blob/master/.github/workflows/format_pr.yml

It's up for discussion whether or not we want this, but it might be a nice idea?

### DIFF
--- a/.github/workflows/format_pr.yml
+++ b/.github/workflows/format_pr.yml
@@ -1,0 +1,37 @@
+name: format-pr
+
+on:
+  schedule:
+    - cron: '0 * * * *'
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


mfalt:
Exactly what does it do? Format all the code when a PR is pulled?
I'm a bit worried that it might accidentally create a huge diff if they change the formatting slightly.

baggepinnen:
It submits a PR with formatted code after a PR is pulled, and probable creates a large diff which I agree is not so nice. Perhaps a better strategy is to push a commit that performs the formatting to the PR branch, in that way the formatting change would be squashed together with the other commits in the PR.

baggepinnen:
I tried this for one of my repos and it was extremely annoying

PR #359, opened 2020-10-10
Submit coverage to codecov instead of coveralls

baggepinnen:
Let's see if I can get codecov to comment something on this PR


mfalt:
Looks good, I guess we can just update README with the new links and pull this?

baggepinnen:
Will add a slack badge when we've figured out how to best access the correct channel through a link

PR #360, opened 2020-10-10
remove lsima simulating an adaptive controller

baggepinnen:
These functions were never exported and are not used by anything in the toolbox.


mfalt (approved):
There are no matches on google either, so probably no one uses it.

PR #364, opened 2020-10-11
Implement forward euler for c2d and d2c

baggepinnen:
Forward Euler is given by
```
x' = x + h*Ax
Ad = (I + h*Ac)
```


baggepinnen:
The rest of the tests only fail on julia v1.0  now

PR #365, opened 2020-10-11
Create TagBot.yml

baggepinnen:
It's the current recommendation to use this workflow instead of the old JuliaTagBot. This bot produces automatic changelogs which is quite nice


mfalt (approved):

baggepinnen:
I uninstalled the previous tagbot

PR #366, opened 2020-10-11
Correct legend in README plot

baggepinnen:


mfalt:
We should probably just remove the TransferFunction[...] notation also. It was needed in some early version of the package to avoid creating a mimo system

PR #368, opened 2020-10-13
Make default frequency vector span larger

baggepinnen:
I have long thought that the default frequency vector spans too few frequencies to make, e.g., bodeplots easy to parse. This PR approximately extends the frequency vector by one decade in each direction. I assume the result will be visible in the plots posted by the plot bot.

### DIFF
--- a/src/freqresp.jl
+++ b/src/freqresp.jl
@@ -164,8 +164,8 @@ function _bounds_and_features(sys::LTISystem, plot::Val)
     fzp = sort!(fzp)
     # Determine the bounds on the frequency vector
     if !isempty(fzp)
-        w1 = floor(fzp[1] - 0.2)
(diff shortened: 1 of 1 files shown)


baggepinnen:
@JuliaControlBot test-plots

baggepinnen:
I like the new one better, anyone who disagrees?

KronosTheLate:
I like the new one a whole lot better. I have actually wondered about trying to fix this myself, so I am very happy with this change. 

If the range is too large, one can always make it smaller by changing xlims. Possibly you (we? Idk if my few contributions allow my self-inclusion into "the people who decide changes in this package") could extend the range by yet another decade in each direction, but set the default xlims to the value you set in this PR, as the plots look good now. That way, one could make the ranger broader than the default when needed.

PR #372, opened 2020-10-16
WIP: Added DAE version of DDE system

mfalt:
This is an implementation of lsim for `DelayLTISystem` where the differential equation is solved as an Delay-DAE instead of a DDE.
It works, in particular also when `D22 /= 0`, but the numerical performance is pretty bad. Posted here for reference.


mfalt:
For reference on the quality:
The system 
`sys_known = feedback(delay(1)*2.0/s, 1)`
and 
`sys_known = feedback(2.0/s*delay(1), 1)`
was simulated and compared to the old method.
y1 is the correct result here:
lsim
y1 and y2 are the corresponding errors, quite significant
lsimError

mfalt:
Closed by https://github.com/JuliaControl/ControlSystems.jl/pull/377

PR #373, opened 2020-10-18
WIP: Delay lsim

mfalt:
This is a new version of https://github.com/JuliaControl/ControlSystems.jl/pull/372 where I manually save the delayed signal and solve the system using the standard Tsit5() ODE solver.
This is still WIP, impulse seems to fail, and the y signal needs to be handled better.
We should probably also do some interpolation of my custom history logger.

Edit: the speed is partially slow because of a bug in `Deque` that I fixed: https://github.com/JuliaCollections/DataStructures.jl/pull/689


olof3:
If you were looking for more tests, you could try `step(feedback(1, exp(-s)))`, although it could be a bit challenging. It certainly is for Matlab's solver.

mfalt:
The accuracy was significantly better by reverting back to DelayDiffEq (but still formulating as an ODE) since we can use the discontinuity tracking. There are only two problems remaining:
* We can't simulate systems without internal states (e.g. `delay(1)`)
* Accuracy could be improved by tracking discontinuities in the manual history function. We currently lose most of the accuracy when the solver steps arbitrarily close `t_k-eps` to a known discontinuity `t_k`, since we interpolate from that point to the discontinuity. If we saved (we already know thanks to DDE.jl) that the point `t_k` is a discontinuity, then we could either extrapolate from old values `t_{k-2},t_{k-1}` or do a ZOH from `t_{k-1}`.

mfalt:
>     * Accuracy could be improved by tracking discontinuities in the manual history function. 

This is implemented now, the only remaining thing that is blocking is lsim of static systems.
We could also improve the search in history function to speed it up a but, but I don*t thiink this is a bottleneck.

olof3:
> the only remaining thing that is blocking is lsim of static systems.

You mean without "finite states" as you said in your previous comment? I guess it should be easy in principle, it's just that it isn't done yet?

src/delay_history.jl
```diff
@@ -0,0 +1,168 @@
...
+            return v1
+        else
+            # Extrapolate linearly from two previous points
+            return (v0*(t1 - t) + v1*(t - t0))/(t1-t0)
+        end
+    else
+        # Interpolate between current and next
+        return (v1*(t2 - t) + v2*(t - t1))/(t2-t1)
+    end
+    return v1 # Can't happen
```
olof3, line 102: Isn't it better to throw an error if this cant happen? Seems like a weird programming style, or is it  to avoid type instability?

src/delay_history.jl
```diff
@@ -0,0 +1,168 @@
...
+    buff::Vector{Deque{TimedVal{ValT,TimeT}}}
+    τ::Vector{TimeT}
+    disc::Vector{ValT}
+    function DDEBuffer(::Type{ValT}, τ::AbstractVector{TimeT}) where {ValT,TimeT}
+        buff = [Deque{TimedVal{ValT,TimeT}}() for i in 1:length(τ)]
+        new{ValT,TimeT}(buff, copy(τ), ValT[])
+    end
+end
+
+# get the element at the back
```
olof3, line 17: I don't think that these kind of comments add anything to the understanding of what these methods do. The code itself is as clear as it gets. Keeping these kinds of comments adds clutter and the risk of that the comment potentially becomes inconsistent and confusing if the code is changed. 

Of course comments in general are good. Would be nice with a comment that summarize how `DDEBuffer`, `HistoryFunction`, etc fits together.

src/delay_history.jl
```diff
@@ -0,0 +1,168 @@
+struct TimedVal{ValT, TimeT}
+    t::TimeT
+    val::ValT
+end
```
olof3, line 4: What is this type used for? A comment would be good. Is it a timestamped value? Then `TimestampedVal` would add a lot of clarity (and the comment could perhaps be skipped).

src/delay_history.jl
```diff
@@ -0,0 +1,168 @@
...
+
+# TODO: Create index of recently accessed times for quicker search
+struct DDEBuffer{ValT, TimeT}
+    buff::Vector{Deque{TimedVal{ValT,TimeT}}}
+    τ::Vector{TimeT}
+    disc::Vector{ValT}
```
olof3, line 10: `disc` seems a bit ambiguous, would prefer `discontinuities` even though it is a bit awkwardly long.

src/delay_history.jl
olof3, line 17: I don't think that these kind of comments add anything to the understanding of what these methods do. The code itself is as clear as it gets. Keeping these kinds of comments adds clutter and the risk of that the comment potentially becomes inconsistent and confusing if the code is changed. 

Of course comments in general are good. Would be nice with a comment that summarizes how `DDEBuffer`, `HistoryFunction`, etc fits together.

src/delay_history.jl
```diff
@@ -0,0 +1,168 @@
...
+        # If extrapolating forward
+        @warn "Interpolation inaccurate at t=$t"
+        return v1
+    elseif t2 <= t1
+        @warn "Interpolation found two equal time steps at t=$t"
+        return v1
+    elseif t2 in b.disc
+        # Next time point is a discontinuity
+        if v0 === nothing
+            # Not no provious point exists
```
olof3, line 87: Some typos.

src/delay_history.jl
```diff
@@ -0,0 +1,168 @@
...
+    `h = HistoryFunction(::Type{ValT}, f::Function, tinit, τ; eps=sqrt(ValT)) where ValT`
+
+    Returns object `h` that is able to store and return a history of old values.
+    It will store up to `length(τ)` values.
+
+    h(t,i) will return the stored value with index `i` at a previous time `t`.
+    If `t < tinit` then h(t,i) = f(t,i).
+
+    `h[t,i] = v` Saves the value `v` for index `i` at time `t.
+    It will also erase old values at index `i` for times before `t-τ[i]`
```
olof3, line 125: erase -> drop ?

src/delay_systems.jl
```diff
@@ -90,93 +99,151 @@ function lsim(sys::DelayLtiSystem{T,S}, u, t::AbstractArray{<:Real}; x0=fill(zer
...
+    u!(uout, t)
 
-    u!(uout, t)     # uout = u(t)
+    # hout = d(t-Tau)
+    for k=1:length(Tau)
+        hout[k] = h(t-Tau[k], k)
+    end
 
-    #dx .= A*x + B1*ut
+    #dx(t) .= A*x(t) + B1*u(t) +B2*d(t-tau)
```
olof3, line 117: This looks a bit off.. I guess i would expect dx/dt, or dotx, but perhaps this is standard notation?

src/delay_history.jl
```diff
@@ -0,0 +1,168 @@
...
+
+# add an element to the back
+function Base.push!(b::DDEBuffer, t, val, indx::Int)
+    if !isempty(b,indx) && last(b, indx).t > t
+        throw(AssertionError("Can only push times after the buffer end"))
+    end
+    push!(b.buff[indx], TimedVal(t, val))
+end
+
+# Get last value before time t
```
olof3, line 40: Is this comment misplaced? It seems like the function does more than this?

olof3 (commented):
Nice work @mfalt ! A massive commit, was somewhat difficult to parse the content and get the full picture, just some minor comments for improved clarity.

olof3:
Is the main motivation for this PR to handle the type of systems in #352, or does it have other advantages?

src/delay_history.jl
mfalt: Yes, its old

baggepinnen:
I had a look through the code but don't really feel qualified to review it properly. It somehow feels like parts of the code in this PR could be upstreamed to DelayDiffEq? If at all possible, it would be nice to have the DiffEq people maintain this code if it could be generally useful.

mfalt:
Closed by https://github.com/JuliaControl/ControlSystems.jl/pull/377

PR #376, opened 2020-10-22
implement conv and remove DSP

albheim:
Remove dependency on DSP and implement naive conv internally.


Project.toml
```diff
@@ -33,9 +31,10 @@ julia = "1.0"
...
 [extras]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 GR = "28b8d3ca-fb5f-59d9-8090-bfdbd6d07a71"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+DSP = "717857b8-e6f2-59f4-9121-6e50c889abd2"
 
 [targets]
 test = ["Test", "Documenter", "GR", "StaticArrays"]
```
mfalt, line 40: If DSP is needed as part of test environment, it needs to be here too

albheim: Will fix

mfalt:
Are there any differences between this implementation and the one in DSP?

albheim:
> Are there any differences between this implementation and the one in DSP?

I took the implementation from polynomials.jl multiplication method. I did look at the DSP one but it was not very easy to read since it is supposed to be very general and for any number of dimensions, but my quick glance made me think that they do something pretty similar at least.

baggepinnen:
I actually have an upcoming PR that makes quite extensive use of DSP.jl to perform factorizations, and I have also implemented conversions between our types and theirs so we can use DSP for filter design etc. I could put this in another repo, but it might as well be in Controlsystems.jl?

albheim:
> I actually have an upcoming PR that makes quite extensive use of DSP.jl to perform factorizations, and I have also implemented conversions between our types and theirs so we can use DSP for filter design etc. I could put this in another repo, but it might as well be in Controlsystems.jl?

Sounds good to me. The reason for this PR was that we only used `conv` from DSP and only for rather simple cases. Since DSP seemed to be one of the packages that added a bit to our load time I thought we could get rid of it with our own simple implementation of `conv`. But if you use it in more stuff now I have nothing against keeping it.

If no one object I think we can close this.

baggepinnen:
Roughly how much did DSP add to the startup time? I can see the point of keeping startup time low so maybe more people should weigh in. I almost always have DSP loaded because I need something from there, so for me it's a non-issue. But other people might not be doing as much filtering etc.?

albheim:
> Roughly how much did DSP add to the startup time? I can see the point of keeping startup time low so maybe more people should weigh in. I almost always have DSP loaded because I need something from there, so for me it's a non-issue. But other people might not be doing as much filtering etc.?

Wasn't too bad actually, first using after change was 65 and 77 seconds (for no DSP vs DSP) while second using took around 35 and 37 seconds. So as long as we feel it adds something useful it does not feel like a too heavy dependancy to me.

baggepinnen:
Alright sweet. I'm on Julia 1.6 and it takes me about 3s to load the package, so it gets better by itself in due time.

PR #377, opened 2020-10-22
DelaySystems lsim as DDE

mfalt:
This is finally a version of simulating delay systems that is both accurate and clean. The idea is to let DifferetialEquations.jl keep track of `y(t)` and `d(t)` by integrating them into variables `Y(t),D(t)`. The values `d(t-\tau)` can then be retrieved as a derivative of the built in history function. And the derivative of `Y(t)` is saved while it is computed. The only downside I can see is that we probably loose an order in the solution, i.e. a 5th order solver could probably only be accurate to 4th order, but this is still better than the alternatives.

Potentially features that we could implement is
* allowing the user to supply the history `d(t)` for times before 0
* Allowing the user to access the full solution sol, and logging d(t).

The current is better than the previous implementation, and comparable to the ODE version https://github.com/JuliaControl/ControlSystems.jl/pull/373 and better tan the DAE version https://github.com/JuliaControl/ControlSystems.jl/pull/372


Project.toml
```diff
@@ -7,7 +7,9 @@ version = "0.8.0"
 [deps]
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
 DSP = "717857b8-e6f2-59f4-9121-6e50c889abd2"
+DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
```
mfalt, line 10: This should be removed

src/ControlSystems.jl
```diff
@@ -105,6 +105,9 @@ import LinearAlgebra: BlasFloat
 export lyap # Make sure LinearAlgebra.lyap is available
 import Printf, Colors
 import DSP: conv
+import DataStructures: Deque # For DDE history simulation
```
mfalt, line 108: Remove

src/delay_systems.jl
```diff
@@ -81,7 +81,12 @@ end
 
     Returns: times `t`, and `y` and `x` at those times.
 """
-function lsim(sys::DelayLtiSystem{T,S}, u, t::AbstractArray{<:Real}; x0=fill(zero(T), nstates(sys)), alg=MethodOfSteps(Tsit5()), kwargs...) where {T,S}
+function lsim(sys::DelayLtiSystem{T,S}, u, t::AbstractArray{<:Real};
+        x0=fill(zero(T), nstates(sys)),
+        alg=DelayDiffEq.MethodOfSteps(Tsit5()),
+        abstol=1e-6, reltol=1e-6,
+        kwargs...) where {T,S}
```
albheim, line 88: Update docstring to reflect the added options?

src/delay_systems.jl
```diff
@@ -90,93 +95,124 @@ function lsim(sys::DelayLtiSystem{T,S}, u, t::AbstractArray{<:Real}; x0=fill(zer
...
+    A, B1, B2, C1, C2, D11, D12, D21, D22, y, Tau, u!, uout, hout, tmpx, tmpy1, tmpy2, tmpd1, tmpnyx, tsave = p
 
-    u!(uout, t)     # uout = u(t)
+    nx = size(A,1)
+    nd = length(Tau)
+    ny = size(C1,1)
+    
+    dx = view(du, 1:nx)
+    dy = view(du, (nx+1):(nx+ny))
+    dd = view(du, (nx+ny+1):(nx+ny+nd))
```
albheim, line 112: Given that this is dx, dY and dD I think it would be clearer to either name them that or name them dx, y and d.

src/delay_systems.jl
```diff
@@ -90,93 +95,124 @@ function lsim(sys::DelayLtiSystem{T,S}, u, t::AbstractArray{<:Real}; x0=fill(zer
...
+    x = view(u, 1:nx)
 
-    #dx .= A*x + B1*ut
+    # uout = u(t)
+    u!(uout, t)
+
+    # hout = d(t-Tau)
+    for k=1:length(Tau)
+        # Get the derivative of the history for d at indices corresponding to Tau
+        hout[k] = h(p, t-Tau[k], Val{1}, idxs=(nx+ny+k))
```
albheim, line 121: I think it should be "hout = D(t-Tau)" and "history of D at"

src/delay_systems.jl
```diff
@@ -90,93 +95,124 @@ function lsim(sys::DelayLtiSystem{T,S}, u, t::AbstractArray{<:Real}; x0=fill(zer
...
-    _lsim(sys, u!, t, x0, alg; kwargs...)
+
+    _lsim(sys, u!, t, x0, alg; abstol=abstol, reltol=reltol, kwargs...)
 end
 
-function dde_param(dx, x, h!, p, t)
-    A, B1, B2, C2, D21, Tau, u!, uout, hout, tmp = p
+# We actually simulate the integral of y and d here.
+function dde_param(du, u, h, p, t)
+    A, B1, B2, C1, C2, D11, D12, D21, D22, y, Tau, u!, uout, hout, tmpx, tmpy1, tmpy2, tmpd1, tmpnyx, tsave = p
```
albheim, line 104: What is tmpnyx? Seems to not be used, or am I blind?

src/delay_systems.jl
albheim: And same for some other places. Also y -> Y and d -> D?

src/delay_systems.jl
```diff
@@ -90,93 +95,124 @@ function lsim(sys::DelayLtiSystem{T,S}, u, t::AbstractArray{<:Real}; x0=fill(zer
...
-
-    @views for k=1:length(Tau)     # Add each of the delayed signals
-        u!(uout, t-Tau[k])      # uout = u(t-tau[k])
-        h!(hout, p, t-Tau[k])
-        dk_delayed = dot(C2[k,:], hout) + dot(D21[k,:], uout)
-        dx .+= B2[:, k] .* dk_delayed
+    mul!(tmpx, B1, uout)
+    dx .+= tmpx
+    mul!(tmpx, B2, hout)
+    dx .+= tmpx
```
olof3, line 129:
```suggestion
    dx .+= mul!(tmpx, B1, uout)
    dx .+= mul!(tmpx, B2, hout)
```
Perhaps more readable? Or use 5-arg mul, that would save us a lot of tmp variables.

src/delay_systems.jl
```diff
@@ -90,93 +95,124 @@ function lsim(sys::DelayLtiSystem{T,S}, u, t::AbstractArray{<:Real}; x0=fill(zer
...
         (out, t) -> (out .= u(t))
     end
-    _lsim(sys, u!, t, x0, alg; kwargs...)
+
+    _lsim(sys, u!, t, x0, alg; abstol=abstol, reltol=reltol, kwargs...)
 end
 
-function dde_param(dx, x, h!, p, t)
-    A, B1, B2, C2, D21, Tau, u!, uout, hout, tmp = p
+# We actually simulate the integral of y and d here.
```
olof3, line 102: Add comment with something like
`#generic parametrized dde used for simulating DelayLtiSystem`
(if this is indeed true)

src/delay_systems.jl
```diff
@@ -90,93 +95,124 @@ function lsim(sys::DelayLtiSystem{T,S}, u, t::AbstractArray{<:Real}; x0=fill(zer
...
+    ny = size(C1,1)
+    
+    dx = view(du, 1:nx)
+    dy = view(du, (nx+1):(nx+ny))
+    dd = view(du, (nx+ny+1):(nx+ny+nd))
+    x = view(u, 1:nx)
 
-    #dx .= A*x + B1*ut
+    # uout = u(t)
+    u!(uout, t)
```
olof3, line 116: It's quite unfortunate that DifferentialEquations prefer to use `u` for the state. I guess at least one reminder of this would be in order, or if we could use `v` or some other letter for this state?

As it is now, it could be somewhat confusing.

src/delay_systems.jl
albheim: I think the 5-arg mul seems nice and fast
```
using LinearAlgebra
using BenchmarkTools

A = randn(10, 10)
B = randn(10, 1)
C = randn(10, 3)

x = randn(10, 1)
u = randn(1, 1)
h = randn(3, 1)

dx = zero(x)
tmp = similar(x)

function f1(A, B, C, x, u, h, tmp)
    mul!(dx, A, x)
    mul!(tmp, B, u)
    dx .+= tmp
    mul!(tmp, C, h)
    dx .+= tmp
end
@benchmark f1($A, $B, $C, $x, $u, $h, $tmp) # ~1 microsecond

function f2(A, B, C, x, u, h, tmp)
    mul!(dx, A, x)
    dx .+= mul!(tmp, B, u)
    dx .+= mul!(tmp, C, h)
end
@benchmark f2($A, $B, $C, $x, $u, $h, $tmp) # ~1 microsecond

function f3(A, B, C, x, u, h)
    mul!(dx, A, x)
    mul!(dx, B, u, 1, 1)
    mul!(dx, C, h, 1, 1)
end
@benchmark f3($A, $B, $C, $x, $u, $h) # ~650 nanoseconds
```

src/delay_systems.jl
mfalt: Yes, althouh in this function I use x for states and u for control signal to make it slightly more clear.

albheim:
How is it going with this, it seems pretty done? Maybe some comments to look at, but would be nice to get it merged.

Issue #369, opened 2020-10-14
Using zpk with complex poles requires specific order for conjugates


albheim:
Is this documented somewhere? 

Cause I wanted to use zpk to create a system and had some complex values that I threw in in some random order and got this error about not having matching complex conjugates. Looking at the code it seems that the `roots2real_poly_factors` method requires the pole with positive imaginary part to come first among the two and the other right after. I don't really see the reason why we require an order at all? It should not be very tricky to just keep a set of non-matched complex poles and check through them for matches for every new complex pole we hit, right? Or at least we should document it in the docstring for zpk so people understand why they get this error.

mfalt:
The reason behind this is that it is not trivial to do the matching when there are numerical errors. Say for example poles at
`(1+x*im),(1-x*im),1+0*im` for some very small `x`. If you are not very careful, under numerical errors, you might match `(1+x*im)`and `1+0*im`. But now you are left with `(1-x*im)` which is not close enough to the imaginary axis to be considered "real", and you therefore end up with a complex valued system, or an error, even when the system you were manipulating was real.
The good thing is that for real matrices, eigenvalue solvers like LAPACK will reduce their problems to 2x2 matrices, and therefore always know which eigenvalues belong together. The result is that we can rely on this specific structure internally.

This became a problem when julia changed the behaviour of the eigenvalue solver in v1.2, to do some random ordering by default. You can therefore find calls to `eigvalsnosort` (found in `utilities.jl`), which asks julia not to sort them, and we therefore get the order as supplied by LAPACK.

I do agree however, that we should either document this requirement, or sort the poles that the user supplies when creating a system. Especially if the user tries to create a TransferFunction{Continous, SisoZpk{<:Real,<:Complex}}, for example in
https://github.com/JuliaControl/ControlSystems.jl/blob/4d1efa16ba8a35263203c2678846f06af7803023/src/types/SisoTfTypes/SisoZpk.jl#L117

olof3:
I think that the error message from `zpk` is already reasonable clear about what the cause and the fix is, but it could definitely be made more clear and documented further.

```julia
julia> zpk([], [-1 + im, 1, -1 - im], 1)
ERROR: AssertionError: zpk model should be real-valued, but poles do not come in conjugate pairs.
```
[...]

PR #385, opened 2020-10-30
Order zpk poles in constructor

albheim:
Add `order_conjugates!` to pair up and order poles/zeros in the order LAPACK uses (positive imaginary part first followed by conjugate) and call it in constructor for SisoZpk type if there are complex poles/zeros.

Ordering only cares about pairing up exact conjugates and putting them in the desired order, and leaves other things as it is. It also skips a complex number if there is no conjugate, which will then be caught by `check_real` which is called right after. 

From #369 I though that there could be numerical problems that forced us to assume conjugates might not match perfectly, but I believe after discussing with @mfalt that this was a misunderstanding from my side and that we can assume the correct conjugate always exists exactly in our internal transformations. And assuming that the user supplies exact conjugates is reasonable to me so I implemented this under the assumption that there will be exact matches, else it will error in `check_real`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -37,4 +37,4 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 DSP = "717857b8-e6f2-59f4-9121-6e50c889abd2"
 
 [targets]
-test = ["Test", "Documenter", "GR", "StaticArrays"]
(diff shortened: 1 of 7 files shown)


albheim:
Woops, did this based on my conv branch... 

Is the easiest to just start a new branch locally based on master and remake the few changes I had?

albheim:
Made #386 instead

PR #386, opened 2020-10-30
Add function to order conjugate pairs in zpk constructor

albheim:
Add `order_conjugates!` to pair up and order poles/zeros in the order LAPACK uses (positive imaginary part first followed by conjugate) and call it in constructor for SisoZpk type if there are complex poles/zeros.

Ordering only cares about pairing up exact conjugates and putting them in the desired order, and leaves other things as it is. It also skips a complex number if there is no conjugate, which will then be caught by `check_real` which is called right after. 

From #369 I though that there could be numerical problems that forced us to assume conjugates might not match perfectly, but I believe after discussing with @mfalt that this was a misunderstanding from my side and that we can assume the correct conjugate always exists exactly in our internal transformations. And assuming that the user supplies exact conjugates is reasonable to me so I implemented this under the assumption that there will be exact matches, else it will error in `check_real`.

### DIFF
--- a/src/types/SisoTfTypes/SisoZpk.jl
+++ b/src/types/SisoTfTypes/SisoZpk.jl
@@ -103,7 +103,9 @@ function minreal(sys::SisoZpk{T,TR}, eps::Real) where {T, TR}
     SisoZpk{T, TR}(newZ, newP, sys.k)
 end
 
-"""Reorder the vector x of complex numbers so that complex conjugates come after each other, with the one with positive imaginary part first. Returns true if the conjugates can be paired and otherwise false."""
+""" Reorder the vector x of complex numbers so that complex conjugates come after each other, 
+    with the one with positive imaginary part first. Returns true if the conjugates can be 
+    paired and otherwise false."""
 function pairup_conjugates!(x::AbstractVector)
     i = 0
     while i < length(x)
--- a/src/types/SisoTfTypes/SisoZpk.jl
+++ b/src/types/SisoTfTypes/SisoZpk.jl
@@ -12,10 +12,8 @@ struct SisoZpk{T,TR<:Number} <: SisoTf{T}
             z = TR[]
         end
         if TR <: Complex && T <: Real
-            order_conjugates!(z)
-            @assert check_real(z) "zpk model should be real-valued, but zeros do not come in conjugate pairs."
-            order_conjugates!(p)
-            @assert check_real(p) "zpk model should be real-valued, but poles do not come in conjugate pairs."
+            @assert pairup_conjugates!(z) "zpk model should be real-valued, but zeros do not come in conjugate pairs."
+            @assert pairup_conjugates!(p) "zpk model should be real-valued, but poles do not come in conjugate pairs."
         end
         new{T,TR}(z, p, k)
     end
@@ -104,35 +102,14 @@ function minreal(sys::SisoZpk{T,TR}, eps::Real) where {T, TR}
     SisoZpk{T, TR}(newZ, newP, sys.k)
 end
 
-# FIXME: Perhaps move to some other file with auxilliary functions,
-# the name could also be imporoved. Perhaps this functionality can be found in some other package.
-""" If TR is Complex and T is Real, check that every pole is matched to its conjugate
-    this assumes that the compelx poles are ordered as they are output by the LAPACK
-    routines that return complex-conjugated values, i.e., (x+iy) is followed by (x-iy)"""
-function check_real(r_vec::AbstractVector{<:Complex})
-    k = 1
-    while k <= length(r_vec)
-        if isreal(r_vec[k])
-            # Do nothing
-        else
-            if k == length(r_vec) # Element is complex and there is no more elements to match up with
-                return false
-            elseif conj(r_vec[k]) == r_vec[k+1] # Element is complex and succeeding element is its conjugate
-                k=k+1
-            else
-                return false
-            end
-        end
-        k += 1
-    end
-    return true
-end
-
-function order_conjugates!(x)
+""" If TR is Complex and T is Real, pair up every pole with its conjugate as they are ordered 
+    by the LAPACK routines that return complex-conjugated values, i.e., (x+iy) is followed by (x-iy)"""
+function pairup_conjugates!(x)
     i = 0
     while i < length(x)
         i += 1
         imag(x[i]) == 0 && continue
+        foundconj = false
         for j in i+1:length(x)
             if x[i] == conj(x[j])
                 if imag(x[i]) > imag(x[j]) # Swap i+1 <-> j
@@ -145,11 +122,16 @@ function order_conjugates!(x)
                     x[j] = x[i+1]
                     x[i+1] = tmp
                 end
-                i += 1
+                foundconj = true
                 break
             end
         end
+        if !foundconj 
+            return false
+        end
+        i += 1 # Since it is a pair and the conjugate was found
     end
+    return true
 end
 
 function evalfr(f::SisoZpk{T1,TR}, s::T2) where {T1<:Number, TR<:Number, T2<:Number}
--- a/src/types/SisoTfTypes/SisoZpk.jl
+++ b/src/types/SisoTfTypes/SisoZpk.jl
@@ -109,22 +109,19 @@ function pairup_conjugates!(x::AbstractVector)
     while i < length(x)
         i += 1
         imag(x[i]) == 0 && continue
-        foundconj = false
-        for j in i+1:length(x)
-            if x[i] == conj(x[j])
-                if imag(x[i]) > imag(x[j]) # Swap i+1 <-> j
-                    tmp = x[i+1]
-                    x[i+1] = x[j]
-                    x[j] = tmp
-                else # Rotate i -> i+1 -> j -> i
-                    tmp = x[i]
-                    x[i] = x[j]
-                    x[j] = x[i+1]
-                    x[i+1] = tmp
-                end
-                foundconj = true
-                break
-            end
+
+        # Attempt to find a matching conjugate to x[i]
+        j = findnext(==(conj(x[i])), x, i+1)
+        j === nothing && return false
+
+        tmp = x[j]
+        x[j] = x[i+1]
+        # Make sure that the complex number with positive imaginary part comes first
+        if imag(x[i]) > 0
+            x[i+1] = tmp
+        else
+            x[i+1] = x[i]
+            x[i] = tmp
         end
         if !foundconj 
             return false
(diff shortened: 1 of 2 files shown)


src/types/SisoTfTypes/SisoZpk.jl
```diff
@@ -12,7 +12,9 @@ struct SisoZpk{T,TR<:Number} <: SisoTf{T}
             z = TR[]
         end
         if TR <: Complex && T <: Real
+            order_conjugates!(z)
             @assert check_real(z) "zpk model should be real-valued, but zeros do not come in conjugate pairs."
+            order_conjugates!(p)
             @assert check_real(p) "zpk model should be real-valued, but poles do not come in conjugate pairs."
```
albheim, line 18: I believe `check_real` is only used here so maybe it would be nicer to just make that check when ordering them?

albheim:
I found some weird behavior I didn't understand in https://github.com/JuliaControl/ControlSystems.jl/blob/311013eef4dd3384872abfca1c04dabb213c97ed/src/types/zpk.jl#L37 which seems to call the SisoZpk constructor twice. Not really connected to this, but thought I would just ask if anyone had a clue why this would be? The type created is exactly the same as the one the array has, so no conversion or anything should be needed?

src/types/SisoTfTypes/SisoZpk.jl
```diff
@@ -133,7 +128,29 @@ function check_real(r_vec::AbstractVector{<:Complex})
     return true
 end
 
-
+function order_conjugates!(x)
```
albheim, line 131: Currently it modifies the vectors which maybe is not very good since the constructor does not have bang notation? Not sure what the convention is there.

mfalt: We generally let the inputs be references to the user inputted arrays, with the implicit implication that the user might screw things up of they change the vectors afterwards. But I agree, we should probably do a copy here.

src/types/SisoTfTypes/SisoZpk.jl
mfalt: I think it's better to be explicit. But the name of check_real could probably be improved. (I don't think it's necessary to change for this PR)

mfalt (changes requested):

src/types/SisoTfTypes/SisoZpk.jl
olof3: Yes, i think the check should be done when "ordering" them, but I think it would be better to call the function something like `matchup_conjugates`, or `pairup_conjugates`. Whlie it currently does order them, I'd say that the main task is to pair them up. If the name is changed to something like this, it is natural that it throws an error if there is a conjugate that can't be paired up.

olof3:
Great that you're looking into this. This PR definitely addresses the problem that you discovered in that other issue.

I think it could be a good time to think about if the current way is the way we would like to do things (although, I'm quite happy with the current approach). 

* (current approach) Ensure that conjugates come after each other, with the conjugate with positive imaginary part first. This could be a problem for symbolic variables, but this is my only concern. Not sure how much effort or how realistic it is to achieve even half-reasonable support for this?
* Ensure that the conjugates come after each other, but don't worry about which comes first. This doesn't make it any harder to convert to transfer functions, but gives a slightly less ordered impression.
* Just check that the conjugates match up. Then there will be a little bit of more work when converting to a transfer function, but I believe that the impact is primarily there (and an even less orderly print).

albheim:
> I found some weird behavior I didn't understand in
> 
> https://github.com/JuliaControl/ControlSystems.jl/blob/311013eef4dd3384872abfca1c04dabb213c97ed/src/types/zpk.jl#L37
> 
> which seems to call the SisoZpk constructor twice. Not really connected to this, but thought I would just ask if anyone had a clue why this would be? The type created is exactly the same as the one the array has, so no conversion or anything should be needed?

I created a specific issue for this after talking with @mfalt, see #387

albheim:
> I think it could be a good time to think about if the current way is the way we would like to do things (although, I'm quite happy with the current approach).

I also think the current approach works well enough. It does not take much time to pair up the roots, and especially if they are already paired it is basically a single for loop over the roots. It makes for easier/nicer printing which I feel is nice.

Now I only did pair up the roots since that is what crashed the code some times, but I was also wondering if the roots from LAPACK were sorted according to any additional criteria, have not tried to look this up myself. Anyone of you know this?

olof3:
> I was also wondering if the roots from LAPACK were sorted according to any additional criteria, have not tried to look this up myself. Anyone of you know this?

From what I've seen, the only guarantee is that they are only paired up, with the one with positive real imaginary part first. Not 100% sure though.

src/types/SisoTfTypes/SisoZpk.jl
albheim: I changed the name to `pairup_conjugates` and also implemented a check in the pairing loop. For me it is either way with the `check_real`, I feel it is a bit unnecessary to loop through the poles twice, but that probably is a very small fraction of the time so if you think the code is clearer with it I can go with that.

src/types/SisoTfTypes/SisoZpk.jl
albheim: I added `z, p = copy(z), copy(p)` in the case where we reorder the vectors.

src/types/SisoTfTypes/SisoZpk.jl
```diff
@@ -109,32 +103,38 @@ function minreal(sys::SisoZpk{T,TR}, eps::Real) where {T, TR}
...
-            # Do nothing
-        else
-            if k == length(r_vec) # Element is complex and there is no more elements to match up with
-                return false
-            elseif conj(r_vec[k]) == r_vec[k+1] # Element is complex and succeeding element is its conjugate
-                k=k+1
-            else
-                return false
+""" If TR is Complex and T is Real, pair up every pole with its conjugate as they are ordered 
+    by the LAPACK routines that return complex-conjugated values, i.e., (x+iy) is followed by (x-iy)"""
```
olof3, line 107:
```suggestion
"""Reorder the vector x of complex numbers so that complex conjugates come after each other, with the one with positive imaginary part first. Returns true if the conjugates can be paired and otherwise false."""
```
Good with a comment, but what is TR and T? Also, might not necessarily be poles.

src/types/SisoTfTypes/SisoZpk.jl
```diff
@@ -109,32 +103,38 @@ function minreal(sys::SisoZpk{T,TR}, eps::Real) where {T, TR}
...
+function pairup_conjugates!(x)
```
olof3, line 108:
```suggestion
function pairup_conjugates!(x::AbstractVector)
```

src/types/SisoTfTypes/SisoZpk.jl
```diff
@@ -6,19 +6,20 @@ struct SisoZpk{T,TR<:Number} <: SisoTf{T}
...
+    function SisoZpk{T,TR}(z::Vector{TR}, p::Vector{TR}, k::T) where {T<:Number, TR<:Number}
         if k == zero(T)
             p = TR[]
             z = TR[]
         end
         if TR <: Complex && T <: Real
-            @assert check_real(z) "zpk model should be real-valued, but zeros do not come in conjugate pairs."
-            @assert check_real(p) "zpk model should be real-valued, but poles do not come in conjugate pairs."
+            z, p = copy(z), copy(p)
+            @assert pairup_conjugates!(z) "zpk model should be real-valued, but zeros do not come in conjugate pairs."
```
olof3, line 16: Should we just let the error be thrown in `pairup_conjugates!` (similarly to in `common_timeevol`), or is this the way we want to go? This way we can let the user know if the problem is with the zeros or the poles. @mfalt

src/types/SisoTfTypes/SisoZpk.jl
```diff
@@ -109,32 +103,38 @@ function minreal(sys::SisoZpk{T,TR}, eps::Real) where {T, TR}
...
+                else # Rotate i -> i+1 -> j -> i
+                    tmp = x[i]
+                    x[i] = x[j]
+                    x[j] = x[i+1]
+                    x[i+1] = tmp
+                end
+                foundconj = true
+                break
             end
         end
```
olof3, line 129:
```suggestion

        # Attempt to find a matching conjugate to x[i]
        j = findnext(==(conj(x[i])), x, i+1)
        j === nothing && return false

        tmp = x[j]
        x[j] = x[i+1]
        # Make sure that the complex number with positive imaginary part comes first
        if imag(x[i]) > 0
            x[i+1] = tmp
        else
            x[i+1] = x[i]
            x[i] = tmp
        end
```
Perhaps a bit more readable in certain respects. Hopefully it works :)

src/types/SisoTfTypes/SisoZpk.jl
```diff
@@ -109,32 +103,38 @@ function minreal(sys::SisoZpk{T,TR}, eps::Real) where {T, TR}
...
-        k += 1
+        if !foundconj 
+            return false
+        end
```
olof3, line 132:
```suggestion
```

test/test_zpk.jl
```diff
@@ -38,6 +38,9 @@ z = zpk("z", 0.005)
 @test zpk([], [], 1.0) == zpk(Float64[], Float64[], 1.0)
 @test zpk([], [1.0+im,1.0-im], 1.0) == zpk(ComplexF64[], [1.0+im,1.0-im], 1.0)
 
+# Test constructors with different conjugate ordering
+@test zpk([], [1.0-im,1.0+im], 1.0) == zpk(ComplexF64[], [1.0+im,1.0-im], 1.0)
+@test zpk([], [1.0+im,1.0,1.0-im], 1.0) == zpk(ComplexF64[], [1.0+im,1.0-im,1.0], 1.0)
```
olof3, line 43: One or two more tests would be much appreciated, perhaps with poles that don't match up, and perhaps some double poles.

olof3 (commented):
The idea looks alright, some minor suggestions for readability.

src/types/SisoTfTypes/SisoZpk.jl
albheim: Was thinking about it when I did this, I originally wanted to move it to `pairup_conjugates!` which felt like nicer code, but then the error message would not be as informative so I thought this was better after all. But I'm open for suggestions.

src/types/SisoTfTypes/SisoZpk.jl
albheim: Agreed, way nicer.

src/types/SisoTfTypes/SisoZpk.jl
albheim: Yeah, mostly just stole the comment from `checkreal` to have something, but yours is much clearer.

test/test_zpk.jl
albheim: I'll try to fix some more.

mfalt:
I think it looks good, and i prefer to keep the check in the constructor. It is explicit and clear in the code that this is assumed.

olof3:
LGTM

olof3:
re-triggered the tests, and it seems like this can be merged

Issue #333, opened 2020-09-28
More autovec functions


baggepinnen:
We should probably add the following also:
freqrespv, numvecv, denvecv, numpolyv, denpolyv.
Otherwise LGTM.

_Originally posted by @mfalt in https://github.com/JuliaControl/ControlSystems.jl/pull/209#discussion_r494935224_

PR #388, opened 2020-11-02
Update autovec and add more functions

albheim:
Opened this to start working on adding the additional functions mentioned in #333, also found out about `Core.@__doc__` which is a nicer way of attaching docstrings to macro generated functions than how I was dealing with it so will update that also.

### DIFF
--- a/src/freqresp.jl
+++ b/src/freqresp.jl
@@ -100,7 +100,7 @@ end
 Compute the magnitude and phase parts of the frequency response of system `sys`
 at frequencies `w`
 
-`mag` and `phase` has size `(length(w), ny, nu)`""" bode
+`mag` and `phase` has size `(length(w), ny, nu)`""" 
 @autovec (1, 2) function bode(sys::LTISystem, w::AbstractVector)
     resp = freqresp(sys, w)
     return abs.(resp), rad2deg.(unwrap!(angle.(resp),1)), w
@@ -112,7 +112,7 @@ end
 Compute the real and imaginary parts of the frequency response of system `sys`
 at frequencies `w`
 
-`re` and `im` has size `(length(w), ny, nu)`""" nyquist
+`re` and `im` has size `(length(w), ny, nu)`""" 
 @autovec (1, 2) function nyquist(sys::LTISystem, w::AbstractVector)
     resp = freqresp(sys, w)
     return real(resp), imag(resp), w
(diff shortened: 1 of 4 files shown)


src/freqresp.jl
```diff
@@ -5,7 +5,7 @@ Evaluate the frequency response of a linear system
 `w -> C*((iw*im -A)^-1)*B + D`
 
 of system `sys` over the frequency vector `w`."""
-function freqresp(sys::LTISystem, w_vec::AbstractVector{<:Real})
+@autovec (1) function freqresp(sys::LTISystem, w_vec::AbstractVector{<:Real})
```
albheim, line 8: It seems that adding autovec to freqresp is not as simple as this. The macro assumes that the return value of the function is a tuple and that the index sent in indexes into this tuple. 

We could pretty easily create the special case where we send no index (`@autovec () function ...`) and make that mean to `vec` the entire returned value. This would be rather easy to fix and the only way I can come up with right now. Maybe not the most intuitive, but could okay since it is an internal function?

src/freqresp.jl
```diff
@@ -124,7 +124,7 @@ end
 Compute the singular values of the frequency response of system `sys` at
```
mfalt, line 124:
```suggestion
Compute the singular values `sv` of the frequency response of system `sys` at
```

mfalt (changes requested):
Is this ready to pull @albheim ? Do we have any remaining functions we should add autovec to?

albheim:
> Is this ready to pull @albheim ? Do we have any remaining functions we should add autovec to?

I think I was waiting for feedback on using `@autovec () f()=...` to mean that the entire returned value should be autovecd. So far I only added freqresp, and from #333 we still have numvecv, denvecv, numpolyv and denpolyv. They also seem to be a single returned value so it would be using the same notation in that case.

albheim:
It was decided that remaining functions from #333 could wait since they introduced some problems in that includes in `ControlSystems.jl` became circular (could rather easily be solved) and that they did not work with how `@autovec` was currently designed. I have some idea of how we could reconstruct `@autovec` to be able to handle more general input instead of integer indices, but I think that will not be a priority for now. 

So with that I think this PR can be considered done.

baggepinnen:
Mattias requested changes here, but I can't seem to figure out if that has been addressed or not?

albheim:
> Mattias requested changes here, but I can't seem to figure out if that has been addressed or not?

I don't remember exactly what it was but I believe I fixed what was requested.

baggepinnen:
I retriggered the tests to make sure, if they pass we'll pull

PR #393, opened 2020-11-13
removed multiplication sign in tf print

mfalt:
With a recent change in Polynomials.jl we are now able to print transferfunctions without the multiplication sign. I think this was how we used to have it, and we had some "broken" tests for this.
See diff for examples. 

Edit: It would be nice to also change so that we print `s+2.0` instead of `1.0s+2.0` but I don't see how that is possible without overriding their internal `showone(::Type{T}) where T` in Polynomials.jl

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -28,7 +28,7 @@ LaTeXStrings = "1.0"
 MacroTools = "0.5"
 OrdinaryDiffEq = "5.2"
 Plots = "0.24, 0.25, 0.26, 0.27, 0.28, 0.29, 1.0"
-Polynomials = "1.0"
+Polynomials = "1.1.10"
 julia = "1.0"
 
 [extras]
--- a/docs/src/examples/example.md
+++ b/docs/src/examples/example.md
@@ -107,9 +107,9 @@ P  = tf(B,A)
 # output
 
 TransferFunction{Continuous,ControlSystems.SisoRational{Float64}}
-         1.0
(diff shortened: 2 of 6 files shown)


baggepinnen (approved):

olof3:
I had some discussions with the maintainer of Polynomials which was the reason for the change. The maintainer of Polynomials is very helpful and open to suggestions, so I'm pretty sure that more changes could be made.

For example
> s+2.0 instead of 1.0s+2.0 b

was on the table, but there were not many other people weighing in on the discussions, so I don't think that this change made it (perhaps it did though)

We also discussed the number of decimals to print. I think both of us were a bit confused about what the compactness level of prints to REPL should be like. My feeling is that the default behavior for printing Polynomials to the repl uses too many digits (cf. printing row vectors).

I think we could what we want if you guys also weight in on what you think.

KronosTheLate:
> I think we could what we want if you guys also weight in on what you think.

Could you link the discussion for where we should chime in?

mfalt:
The reduction of decimals is implemented in Polynomails and could pretty easily be implemented here if we want. They have an implementation for not printing the "1.0" but it can only be modified per type, e.g `Float64` and not per call.

olof3:
> but it can only be modified per type, e.g Float64 and not per call.

That's not too great. I am pretty sure it would get fixed pretty quickly if you opened an issue.

Issue #397, opened 2020-11-19
Simulator does not allow non-zero D matrix


albheim:
I noticed that when I create a continuous system with equal number of poles and zeros I can't use `step` since it is complaining that a non-zero D-matrix is not supported. 

The error comes from simulator.jl here
https://github.com/JuliaControl/ControlSystems.jl/blob/e9837ee6da0111d7703d4131313d6ff52557cf90/src/simulators.jl#L39-L46

Though it seems that in `lsim`, which is the function calling `Simulator`, it only expects the states to be calculated and calculates the output given the D-matrix by itself. Seems like these two are not entirely synced in how they work, `Simulator` exposes the `y` function which is not used and `lsim` seems like it does not have a problem with the D-matrix but `Simulator` has (which I still don't really see the reason why it is a problem).
https://github.com/JuliaControl/ControlSystems.jl/blob/e9837ee6da0111d7703d4131313d6ff52557cf90/src/timeresp.jl#L183-L192

The assertion that D should be zero comes from a #118 in a commit named "Fix stack overflow error" by @baggepinnen, though when testing to remove that assertion it seems that running
```julia
using ControlSystems
sys = ss(-1, 2, 3, 4)
step(sys)
```
works fine and gives the values I expect when plotting. Running all tests also seems to work without the assertion. So I guess I'm wondering if @baggepinnen has any memory of why it was once needed and if it is still needed for some case that is not tested? Or could we just remove it?

baggepinnen:
I do not have any memory left of this unfortunately, but if you do not get said stack-overflow error I'm all for removing the restriction :)

albheim:
Okay, will put together a PR for that then. 

Also wondering about `Simulator` in general, is it used for any external things? It seems `lsim` is the only place we use it internally, and it seems `lsim` is only using parts of it (`y` function is ignored). 

Was thinking that either it seems it could be dropped and `lsim` can contain those few extra rows, or maybe we should use the functionality exposed by `Simulator` and make sure the `y` function can be used properly with `lsim`?

baggepinnen:
There were once several different `AbstractSimulator` such as for gain scheduling and output feedback etc., but they have since been removed. `Simulator` still allows you to modify any part of the linear dynamics and thus simulate a party nonlinear system, e.g., Hammerstein/Weiner systems etc.
There are also some users of `Simulators` out there, but I couldn't find a discussion on it now so it has probably come up on slack :/

[...]

PR #398, opened 2020-11-19
Fix Simulator blocking non-zero D

albheim:
Remove assertion that blocks non-zero D matrices in `Simulator` type used in `lsim` according to #397

### DIFF
--- a/src/simulators.jl
+++ b/src/simulators.jl
@@ -38,7 +38,6 @@ plot(t, s.y(sol, t)[:], lab="Open loop step response")
 """
 function Simulator(P::AbstractStateSpace, u::F = (x,t) -> 0) where F
     @assert iscontinuous(P) "Simulator only supports continuous-time system. See function `lsim` for simulation of discrete-time systems."
-    @assert all(P.D .== 0) "Can not simulate systems with direct term D != 0"
     f = (dx,x,p,t) -> dx .= P.A*x .+ P.B*u(x,t)
     y(x,t) = P.C*x #.+ P.D*u(x,t)
     y(sol::ODESolution,t) = P.C*sol(t)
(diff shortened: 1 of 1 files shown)


src/simulators.jl
```diff
@@ -38,7 +38,6 @@ plot(t, s.y(sol, t)[:], lab="Open loop step response")
 """
 function Simulator(P::AbstractStateSpace, u::F = (x,t) -> 0) where F
     @assert iscontinuous(P) "Simulator only supports continuous-time system. See function `lsim` for simulation of discrete-time systems."
-    @assert all(P.D .== 0) "Can not simulate systems with direct term D != 0"
     f = (dx,x,p,t) -> dx .= P.A*x .+ P.B*u(x,t)
     y(x,t) = P.C*x #.+ P.D*u(x,t)
```
baggepinnen, line 42:
```suggestion
    y(x,t) = P.C*x .+ P.D*u(x,t)
```
I guess you have to add the contribution of the input to the output as well

albheim: So this is another thing I mentioned in #397, it works as is since `lsim` does not use this `y` function to calculate `y` but instead does it by itself. It feels like `lsim` should either use this (after fixing it) or maybe remove `y` here. I was not sure of the preferred way so I started with only the assertion to test if it ran in the CI pipeline.

I'm not sure I understand the benefit of the `Simulator` type at all. Since it is only used in `lsim` it seems like it would be easier and less code to just write out the simulation there? But either solution is fine for me, wanted some input before I started to make more stuff.

baggepinnen: It should probably still be there, since we put the `y` function in the `Simulator` struct, so the user might call it later and then expects it to contain `Du` if `D != 0`. There is one more method of `y` added below that is also affected.

albheim: Agreed, not sure why I didn't accept the suggestion before.

PR #400, opened 2020-11-23
Field-matching constructors. Fixes #399

### LINKED ISSUE #399: Add constructors with nu, ny, nx for easy rebuilding
Some useful functions that need to recurse through arbitrarily nested `struct`s and rebuild them with different values--like `@set` and `@set!` from Setfield.jl or `replace_particles` (along with its derived functions `mean_object` and `nominal`) from MonteCarloMeasurements.jl--require constructors that match the positions of the type's fields. Since `TransferFunction` and `StateSpace` have fields for their number of inputs/outputs[/states], but no constructors that take them in, it creates problems for these types of functions. For example, these don't currently work:
```julia
julia> using ControlSystems, MonteCarloMeasurements, Setfield

julia> a = zpk([-1±0.1], [im, -im], 5);

julia> @set! a.matrix[1].k = 8
ERROR: MethodError: no method matching TransferFunction(::Array{ControlSystems.SisoZpk{Particles{Float64,2000},Complex{Particles{Float64,2000}}},2}, ::Continuous, ::Int64, ::Int64)
...

julia> mean_object(a)
┌ Error: Failed to create a `TransferFunction` by calling it with its fields in order. For this to work, `TransferFunction` must have a constructor that accepts all fields in the order they appear in the struct and 
accept that the fields that contained particles are replaced by 0. Try defining a meaningful constructor that accepts arguments with the type signature
...
```
but adding
```julia
StateSpace(A, B, C, D, timeevol, nx, nu, ny) = StateSpace(A, B, C, D, timeevol)
```
makes it so they work:
```julia
julia> @set! a.matrix[1].k = 8
TransferFunction{Continuous,ControlSystems.SisoZpk{Particles{Float64,2000},Complex{Particles{Float64,2000}}}}
   1.0s + 1.0 ± 0.1
8.0----------------
     1.0s^2 + 1.0

julia> mean_object(a)
TransferFunction{Continuous,ControlSystems.SisoZpk{Float64,Complex{Float64}}}
    1.0s + 1.0
8.0------------
   1.0s^2 + 1.0

```
This and the similar method for `TransferFunction`, are all that's needed for these types of functions to work. I can make a pull request with these, if it would be useful.

jonniedie:
I just went ahead and made a PR to implement #399 because it was a simple enough change.

### DIFF
--- a/src/types/StateSpace.jl
+++ b/src/types/StateSpace.jl
@@ -49,6 +49,8 @@ end
 function StateSpace(A::MT, B::MT, C::MT, D::MT) where {T, MT <: AbstractMatrix{T}}
     StateSpace(A, B, C, D, Continuous())
 end
+# Necessary for functions that rebuild structs
+StateSpace(A, B, C, D, timeevol, nx, nu, ny) = StateSpace(A, B, C, D, timeevol)
 
 """ If D=0 then convert to correct size and type, else, create 1x1 matrix"""
 function fix_D_matrix(T::Type,B,C,D)
--- a/src/types/StateSpace.jl
+++ b/src/types/StateSpace.jl
@@ -50,7 +50,16 @@ function StateSpace(A::MT, B::MT, C::MT, D::MT) where {T, MT <: AbstractMatrix{T
     StateSpace(A, B, C, D, Continuous())
 end
 # Necessary for functions that rebuild structs
-StateSpace(A, B, C, D, timeevol, nx, nu, ny) = StateSpace(A, B, C, D, timeevol)
+function StateSpace(A, B, C, D, timeevol, nx, nu, ny)
+    if size(A, 1) != nx
+        error("The number of rows of A ($(size(A,1))) is not equal to nx ($nx)")
+    elseif size(B, 2) != nu
+        error("The number of columns of B ($(size(B,2))) is not equal to nu ($nu)")
+    elseif size(C, 1) != ny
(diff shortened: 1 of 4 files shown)


src/types/StateSpace.jl
```diff
@@ -49,6 +49,8 @@ end
 function StateSpace(A::MT, B::MT, C::MT, D::MT) where {T, MT <: AbstractMatrix{T}}
     StateSpace(A, B, C, D, Continuous())
 end
+# Necessary for functions that rebuild structs
+StateSpace(A, B, C, D, timeevol, nx, nu, ny) = StateSpace(A, B, C, D, timeevol)
```
baggepinnen, line 53: Perhaps a check should be made so that the sizes provided  are consistent with the actual sizes of the matrices? Since the user will now be exposed to these constructors, it makes sense to not accept inconsistent values.

baggepinnen (changes requested):
I'm in favor of this, the lack of those constructors bother me on an almost daily basis. Apart from not accepting invalid sizes, I think it looks good.

baggepinnen (approved):

jonniedie:
Oops, looks like I had a typo in my test

olof3:
Shouldn't we just remove the fields and rely on `getproperty` and avoid adding extra code, see #399.

baggepinnen:
This should be fixed now that we no longer have the extra fields (#411)
I hope it solves your original problem (#399)

PR #401, opened 2020-11-23
Remove xvfb as dependancy for plotting tests

albheim:
I noticed that PkgEval does not seem happy with ControlSystems, and after checking with @mfalt he thought this might be because there is tests that require plotting which will not run on a headless server if just running `test ControlSystems`.

ControlSystems CI pipeline seems to use `xvfb` to emulate a virtual x environment to allow for some of the plot tests to run.

I found this [discourse thread](https://discourse.julialang.org/t/deactivate-plot-display-to-avoid-need-for-x-server/19359) discussing plotting on headless servers and thought I would check what the difference was using nul instad of 100 for GKSwstype.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ addons:
   apt:
     packages:
       - libgtk-3-dev
-      - xvfb
       - dvipng
       - texlive
 cache:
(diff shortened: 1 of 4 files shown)


mfalt:
If tests pass with this (and Plots are able to build in makedocs) I think it's a great fix. If the second doesn't work we could probably just add the xvfb to the makedocs part.

albheim:
I was also wondering if there is a reason for having these settings spread out over three files? Would it not be enough to set it in runtests.jl and we could skip the rest?

If people want to run make.jl or test_plot.jl by itself I think it is fine to assume they can know enough to set the plotting environment themselves?

And the one in .travis.yml seems like it is not really necessary at all if it is no matter set in the files where it matters anyway?

mfalt:
> I was also wondering if there is a reason for having these settings spread out over three files? Would it not be enough to set it in runtests.jl and we could skip the rest?
> 
> If people want to run make.jl or test_plot.jl by itself I think it is fine to assume they can know enough to set the plotting environment themselves?
> 
> And the one in .travis.yml seems like it is not really necessary at all if it is no matter set in the files where it matters anyway?
> 
Even though I know how to do it, it is nice to be able to run the scipts locally whithout editing them. But in general it is needed in test_plots.jl so you can run 
`]test ControlSystems`
locally, even on headless-server. In docs/make.jl so you can run `makedocs(modules=[ControlSystems])`.
Maybe it can be removed in `travis.yml` if it is included in those files.

> What is `PLOT_TEST` even used for? Could only find it in three places in the repo, and that was the three places it was set.
This is read by Plots.jl. For example it is used in Plots.jl tests: https://github.com/JuliaPlots/Plots.jl/blob/2774b155b4c2b70db3fefd80bd623c34afed992f/test/runtests.jl#L100
It seems to be related to keeping the performance constant when there is no display information: https://github.com/JuliaPlots/Plots.jl/pull/1318/files

albheim:
> Even though I know how to do it, it is nice to be able to run the scipts locally whithout editing them. But in general it is needed in test_plots.jl so you can run
> `]test ControlSystems`
> locally, even on headless-server. In docs/make.jl so you can run `makedocs(modules=[ControlSystems])`.
> Maybe it can be removed in `travis.yml` if it is included in those files.

Okay, I was thinking that you might want to run it with other settings if you are not on a headless server but I realise that we probably prefer consistency in test/docs as the default and if you really want to run it in some other way you can dig in there and change it.

> This is read by Plots.jl. For example it is used in Plots.jl tests: https://github.com/JuliaPlots/Plots.jl/blob/2774b155b4c2b70db3fefd80bd623c34afed992f/test/runtests.jl#L100
> It seems to be related to keeping the performance constant when there is no display information: https://github.com/JuliaPlots/Plots.jl/pull/1318/files
> Edit: I added it for some reason, but I don't remember exactly.

Okay, seems good as it is then.

albheim:
I also tried to run the tests without our travis settings on my headless workstation and it seems to work fine so hopefully this could solve the PkgEval problem.

albheim:
This was solved in #408

Issue #384, opened 2020-10-30
Allow place to handle multi input systems


albheim:
Currently `place` only handles SISO systems and it might be nice to try to implement something similar as matlab has where they find a robust set of poles for multi input systems according to [Robust Pole Assignment in Linear State Feedback](https://perso.uclouvain.be/paul.vandooren/publications/KautskyNV85.pdf). Have not managed to understand the algorithm but could try to take a look at it if it is of interest.

baggepinnen:
In anticipation of a better method, the following does an okay job
```julia
using ControlSystems, Optim

function generic_place(A, B::AbstractMatrix, desired_poles; max_tries = 20)
    function loss(K)
        e = eigvals(A-B*K)
        sum(abs2.(e-desired_poles))
    end    
    desired_poles = sort(complex.(desired_poles), by=LinearAlgebra.eigsortby)
    local res
    for i = 1:max_tries
        K = randn(reverse(size(B)))
        res = Optim.optimize(
            loss,
            K,
            BFGS(),
            Optim.Options(
                store_trace       = false,
                show_trace        = false,
                show_every        = 1,
                iterations        = 10000,
                allow_f_increases = false,
                time_limit        = 100,
                x_tol             = 0,
                f_tol             = 0,
                g_tol             = 1e-12,
                f_calls_limit     = 0,
                g_calls_limit     = 0,
                ),
                )
        (res.f_converged || res.g_converged || res.x_converged) && break
    end
    res.minimizer, res
end
            
A = randn(3,3)
B = randn(3,1)

desired_poles = [-1,-2,-3]
PK = place(A,B,desired_poles)
OK, res = generic_place(A,B,desired_poles)

@test PK ≈ OK rtol=1e-2
```
Of course, you may add additional penalties to the loss function, encouraging use of potential redundancy to reduce nosie influence etc.

baggepinnen:
Closed by #854

PR #402, opened 2020-11-27
WIP: Robust mimo place

albheim:
Trying to implement a robust mimo pole placement according to #384

### DIFF
--- a/src/synthesis.jl
+++ b/src/synthesis.jl
@@ -119,8 +119,12 @@ function place(A, B, p)
     n != size(B,1) && error("A and B must have same number of rows")
     if size(B,2) == 1
         acker(A,B,p)
+    elseif size(B, 2) <= size(A, 1)
+        placemimo(A, B, p)
+    elseif size(B, 2) == size(A, 1)
+        println("This should be implemented")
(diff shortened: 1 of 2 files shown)


albheim:
Currently it seems like placing real poles work fine, but there is some problem with complex poles which I don't understand.

Testing for random matrices and poles with 5-10 states it seems that `acker` performs a bit faster as expected, taking around 50-100 microseconds while `placemimo` takes 150-1000 microseconds. We have a slightly better accuracy for `placemimo` though, usually coming up one order of magnitude better, though both are around 1e-8 in the normed difference to the real poles.

Will test a bit more once I get complex poles working.

albheim:
> Currently it seems like placing real poles work fine, but there is some problem with complex poles which I don't understand.

Found that the theory in the paper says complex poles closed under conjugate works with this method, but then the algorithm quickly mentions that this specific version only works for real poles...

I will try to figure out how to get it to work for complex numbers also, but not sure when I will find that motivation. If we want it to work for real poles I could quickly remove the tests for complex and add an error if it's called with mimo and complex.

PR #403, opened 2020-11-28
allow specification of Tf instead of time vector in lsim

baggepinnen:


mfalt:
What is the warning codecov complains about here? We should have at least one test for this case.

I guess Julia is still able to infer the type of `t` here?

mfalt:
Also, sys.Ts is not what we want to use for time vector, at least not for continuous systems.

PR #406, opened 2020-12-04
Add dlqr method for StateSpace input

adrhill:
This method was already documented in the `dlqr` docstring but not implemented:
https://github.com/JuliaControl/ControlSystems.jl/blob/6c54a259980fc3ac784b45db395edfcd9ea59300/src/synthesis.jl#L67

### DIFF
--- a/src/synthesis.jl
+++ b/src/synthesis.jl
@@ -100,6 +100,11 @@ function dlqr(A, B, Q, R)
     return K
 end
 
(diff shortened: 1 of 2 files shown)


baggepinnen (approved):
LGTM. Later on we can probably remove the `d` versions of these methods since now the `Continuous/Discrete` is part of the type and this can be solved with dispatch rather than separate functions.

Issue #407, opened 2020-12-04
Migrate from Travis to Github Actions


mfalt:
Travis no longer provides free CI for open source packages, see https://discourse.julialang.org/t/travis-ci-changes-pricing-plans/49528

I have tried to apply for open source credits, but depending on if they approve it and how often we will have to re-apply, it might not be a great long-time solution, so we consider changing to github actions which is still free for open source projects.
If someone wants to try with a PR, there are some code snippets that can probably be used here
https://discourse.julialang.org/t/easy-workflow-file-for-setting-up-github-actions-ci-for-your-julia-package/49765

Edited: I have applied for credits

baggepinnen:
I use github CI for most of my repos nowadays, it's very smooth and on the whole better than Travis, so I welcome the change 🙂

imciner2:
:+1: this would definitely be a good move. GitHub actions has come along way and I think is now the better choice. It also allows for all 3 platforms (Windows, MacOS and Linux) to be tested at once, so there is no more need to have separate CI services for different platforms.

I have a few workflows I use for Julia packages that I can put into an MR if you are interested.

mfalt:
@imciner2 that would be great. I would have done it myself if I want busy finishing my thesis right now.

PS: I got a response from Travis staying that they don't even have the system for OSS and education projects in place yet, so I don't think we should depend on them in the future.

albheim:
Sounds like a good idea to me.

I took a look at the link you posted @mfalt and saw that their sample CI seems to have test and doctest split into 2 different parts, where test is run on 1, 1.x and nightly while doctest is only run on 1. Do we prefer to split it up like this also and remove the call to doctest from runtests.jl? I'm not sure if we have anything specific in the documentation that we want to run on all versions?

albheim:
My PR is just me testing out github actions, and might not lead to something working. So if @imciner2 had some nice workflows I did not mean to steal the issue :laughing:

baggepinnen:
Done :)

PR #408, opened 2020-12-07
Switch to github actions

albheim:
Was interested in testing out github actions so just tried to replace travis with the github actions ci file suggested in discourse linked in top comment in #407


test/runtests.jl
```diff
@@ -39,10 +38,4 @@ my_tests = [
     _t0 = time()
     run_tests(my_tests)
     println("Ran all code tests in $(round(time()-_t0, digits=2)) seconds")
-
-
-    println("Test Doctests")
-    _t0 = time()
-    doctest(ControlSystems)
-    println("Ran Doctests in $(round(time()-_t0, digits=2)) seconds")
```
albheim, line 47: Seems that not having doctest in with the other tests so that it is run when checking coverage reduce coverage quite a bit.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+          - '1' # Leave this line unchanged. '1' will automatically expand to the latest stable 1.x release of Julia.
+          - 'nightly'
+        os:
+          - ubuntu-latest
+        arch:
+          - x64
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install apt deps
+        run: sudo apt-get install libgtk-3-dev dvipng texlive
```
albheim, line 28: Are all these needed for both test and doctest?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: julia-actions/julia-runtest@v1
+      - uses: julia-actions/julia-processcoverage@v1
+      - uses: codecov/codecov-action@v1
+        with:
+          file: lcov.info
```
albheim, line 48: Do we want to run coverage for both 1.0 and 1? What does that even mean, does it average over the two?

imciner2: IMO coverage is probably only needed on one version, unless there starts to be a lot of version-specific code introduced (which I don't forsee being an issue in this package).

test/runtests.jl
imciner2: Can you pinpoint which parts are getting reduced coverage when the doctests are removed? I don't think we should be relying on the doctests for the testing coverage, since I believe doctests are more designed to ensure the documentation examples don't go stale rather than the code is functionally correct.

imciner2:
I think it would be nicer if the different CI parts (e.g. code tests vs. documentation generation) were split out into separate workflow files.

test/runtests.jl
mfalt: I agree, but historically we have had problems with where there docs didn't build properly or were outdated without us noticing, that is why I added it to the tests as well. If GitHub actions allows us to visualize and stop a merge when doctest fails I have no problem removing them from the tests, if we also add tests to cover the same functionality.

.github/workflows/ci.yml
mfalt: I guess they are related to plots, which i think are run in both.

albheim: I tried to remove them from test, and it seems to have worked. So I'll leave them out for now at least, though doctest will still have them.

test/runtests.jl
albheim: > Can you pinpoint which parts are getting reduced coverage when the doctests are removed? I don't think we should be relying on the doctests for the testing coverage, since I believe doctests are more designed to ensure the documentation examples don't go stale rather than the code is functionally correct.

In the [coverage report](https://codecov.io/gh/JuliaControl/ControlSystems.jl/pull/408/changes) you can see exactly which lines lost coverage. Most of it seems to be from pid_design.jl and discrete.jl

albheim: > I agree, but historically we have had problems with where there docs didn't build properly or were outdated without us noticing, that is why I added it to the tests as well. If GitHub actions allows us to visualize and stop a merge when doctest fails I have no problem removing them from the tests, if we also add tests to cover the same functionality.

We can have a step in the workflow that is running `doctest` and running `make.jl` for the docs, and we can have it be required I believe.

> Edit: A nice feature with having the doctest in tests if that wont forget to run the doctest when developing locally. I don't know about you, but I rarely run them manually when developing.

I'd never run doctest locally if it was not in, though I'm not sure I'd want to most of the time. It feels like the tests are what I'd use to see if I'm close to something working, and the running the PR can catch if something broke with the docs, a little less time running locally for me :smile: 

But if you feel it's warranted to run doctest all the time that is also fine by me, you have more experience with the workflows for this package.

baggepinnen:
@albheim how's it going with this PR? It's still marked as WIP, is it getting close to merge? I have a huge backlog of commits to submit but right now there's no CI to show me how many bugs I've introduced :P

albheim:
> @albheim how's it going with this PR? It's still marked as WIP, is it getting close to merge? I have a huge backlog of commits to submit but right now there's no CI to show me how many bugs I've introduced :P

Ohh, yeah it is not really working as intended yet. Have not had time over the weekend. Might have some time tomorrow, or you are very welcome to take a look.

It seems like tests are working for 1 and 1.0 for now, but nightly tries for 6 hrs before stopping, have not looked at that yet but probably want to stop it earlier. Docs also have some problems in that it can fail doctest but still run without showing it, so was talking to Mattias about moving doctest into the normal test (as it has been before).

test/runtests.jl
baggepinnen: The integration of github actions is (non surprising) much better than it ever was for Travis, and a commit or PR will display the red cross instead of the green checkbox whenever any of the worflows failed. It should thus be fairly easy to detect the docs build failing even if it's not part of the tests. There are some noise left due to coverage sometimes reporting a failure for no good reason, but the docs failure remains easily visible on hover over the red cross.

albheim:
It seems like nightly gets stuck when installing deps and building (and does the same locally for me) and will run for 6 hrs as defualt, I changed the runtime to be max 30 minutes since it seems like successful builds are usually under 20 minutes. 

Also updated StaticArrays dependency to only be compatible with 1.0 (only used in testing/documenting) since it has changed to format and thus prints differently which crashed the doctest between julia 1 and 1.0.

baggepinnen:
We do not want to hold back StaticArrays.jl to an old version here, if Julia 1.0 is giving you problems I say we ditch it in favor of 1.3

.github/workflows/docs.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: |
+          julia --project=docs -e '
+            using Pkg
+            Pkg.develop(PackageSpec(path=pwd()))
+            Pkg.instantiate()'
+      - name: Generate and test documentation
+        run: julia --project=docs docs/make.jl
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }}
```
albheim, line 35: https://github.com/JuliaDocs/Documenter.jl/issues/1177

albheim:
> We do not want to hold back StaticArrays.jl to an old version here, if Julia 1.0 is giving you problems I say we ditch it in favor of 1.3

Yeah, I'm a bit confused about this one cause I can't seem to get the same results locally. But I agree that we should not hold back based on that, just testing around a bit to understand what is the problem.

albheim:
Set up a local julia v1.0.5 env and tests does not work (after I have updated docs to match tests for StaticArrays v1.0.1). Trying to manually set StaticArrays version to 1.0.1 gives this 

```julia
(v1.0) pkg> add StaticArrays@v1.0.1
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package OrdinaryDiffEq [1dea7af3]:
 OrdinaryDiffEq [1dea7af3] log:
 ├─possible versions are: [4.0.0, 4.1.0, 4.2.0, 4.3.0, 4.4.0-4.4.1, 4.5.0, 4.6.0, 4.7.0-4.7.1, 4.8.0-4.8.1, 4.9.0, 4.10.0, 4.11.0-4.11.1, 4.12.0-4.12.4, 4.13.0, 4.14.0, 4.15.0-4.15.1, 4.16.0-4.16.5, 4.17.0-4.17.2, 4.18.0-4.18.3, 4.19.0, 4.20.0, 4.21.0-4.21.1, 5.0.0, 5.1.0-5.1.4, 5.2.0-5.2.1, 5.3.0, 5.4.0-5.4.1, 5.5.0, 5.6.0-5.6.1, 5.7.0-5.7.1, 5.8.0-5.8.1, 5.9.0, 5.10.0, 5.11.0-5.11.1, 5.12.0, 5.13.0, 5.14.0, 5.15.0-5.15.1, 5.16.0, 5.17.0-5.17.2, 5.18.0, 5.19.0, 5.20.0-5.20.1, 5.21.0, 5.22.0, 5.23.0, 5.24.0, 5.25.0, 5.26.0-5.26.8, 5.27.0-5.27.1, 5.28.0-5.28.1, 5.29.0, 5.30.0, 5.31.0, 5.32.0-5.32.2, 5.33.0, 5.34.0-5.34.1, 5.35.0-5.35.5, 5.36.0-5.36.1, 5.37.0, 5.38.0-5.38.3, 5.39.0-5.39.1, 5.40.0, 5.41.0, 5.42.0-5.42.10, 5.43.0, 5.44.0, 5.45.0-5.45.1, 5.46.0] or uninstalled
 ├─restricted to versions 5.2.0-5 by ControlSystems [a6e380b2], leaving only versions [5.2.0-5.2.1, 5.3.0, 5.4.0-5.4.1, 5.5.0, 5.6.0-5.6.1, 5.7.0-5.7.1, 5.8.0-5.8.1, 5.9.0, 5.10.0, 5.11.0-5.11.1, 5.12.0, 5.13.0, 5.14.0, 5.15.0-5.15.1, 5.16.0, 5.17.0-5.17.2, 5.18.0, 5.19.0, 5.20.0-5.20.1, 5.21.0, 5.22.0, 5.23.0, 5.24.0, 5.25.0, 5.26.0-5.26.8, 5.27.0-5.27.1, 5.28.0-5.28.1, 5.29.0, 5.30.0, 5.31.0, 5.32.0-5.32.2, 5.33.0, 5.34.0-5.34.1, 5.35.0-5.35.5, 5.36.0-5.36.1, 5.37.0, 5.38.0-5.38.3, 5.39.0-5.39.1, 5.40.0, 5.41.0, 5.42.0-5.42.10, 5.43.0, 5.44.0, 5.45.0-5.45.1, 5.46.0]
 │ └─ControlSystems [a6e380b2] log:
 │   ├─possible versions are: 0.8.0 or uninstalled
 │   └─ControlSystems [a6e380b2] is fixed to version 0.8.0
 ├─restricted by julia compatibility requirements to versions: [4.0.0, 4.1.0, 4.2.0, 4.3.0, 4.4.0-4.4.1, 4.5.0, 4.6.0, 4.7.0-4.7.1, 4.8.0-4.8.1, 4.9.0, 4.10.0, 4.11.0-4.11.1, 4.12.0-4.12.4, 4.13.0, 4.14.0, 4.15.0-4.15.1, 4.16.0-4.16.5, 4.17.0-4.17.2, 4.18.0-4.18.3, 4.19.0, 4.20.0, 4.21.0-4.21.1, 5.0.0, 5.1.0-5.1.4, 5.2.0-5.2.1, 5.3.0, 5.4.0-5.4.1, 5.5.0, 5.6.0-5.6.1, 5.7.0-5.7.1, 5.8.0-5.8.1, 5.9.0, 5.10.0, 5.11.0-5.11.1, 5.12.0, 5.13.0, 5.14.0, 5.15.0-5.15.1, 5.16.0, 5.17.0-5.17.2, 5.18.0, 5.19.0, 5.20.0-5.20.1, 5.21.0, 5.22.0, 5.23.0, 5.24.0, 5.25.0, 5.26.0-5.26.8, 5.27.0-5.27.1, 5.28.0-5.28.1, 5.29.0, 5.30.0, 5.31.0] or uninstalled, leaving only versions: [5.2.0-5.2.1, 5.3.0, 5.4.0-5.4.1, 5.5.0, 5.6.0-5.6.1, 5.7.0-5.7.1, 5.8.0-5.8.1, 5.9.0, 5.10.0, 5.11.0-5.11.1, 5.12.0, 5.13.0, 5.14.0, 5.15.0-5.15.1, 5.16.0, 5.17.0-5.17.2, 5.18.0, 5.19.0, 5.20.0-5.20.1, 5.21.0, 5.22.0, 5.23.0, 5.24.0, 5.25.0, 5.26.0-5.26.8, 5.27.0-5.27.1, 5.28.0-5.28.1, 5.29.0, 5.30.0, 5.31.0]
 └─restricted by compatibility requirements with StaticArrays [90137ffa] to versions: 5.46.0 or uninstalled — no versions left
   └─StaticArrays [90137ffa] log:
     ├─possible versions are: [0.8.0-0.8.3, 0.9.0-0.9.2, 0.10.0, 0.10.2-0.10.3, 0.11.0-0.11.1, 0.12.0-0.12.5, 1.0.0-1.0.1] or uninstalled
     └─restricted to versions 1.0.1 by an explicit requirement, leaving only versions 1.0.1
```

albheim:
> Set up a local julia v1.0.5 env and tests does not work (after I have updated docs to match tests for StaticArrays v1.0.1). Trying to manually set StaticArrays version to 1.0.1 gives this
> 

Seems like the same thing holds for 1.3

albheim:
As it is now I set the compatability to StaticArrays 0.12 because then it could run for both julia 1 and 1.0 versions. This was more in case we wanted to have something that worked now for your incoming patches @baggepinnen, but not how I'd like to see it solved in the long run. I did not manage to get it to work with julia 1 and 1.3 either cause then there was a dependancy on DataStructures.jl that did not work with StaticArrays 1.0. Didn't look at why DataStructures even get into the dependancies but will see if I can figure that out later.

baggepinnen:
I added this to https://github.com/JuliaControl/ControlSystems.jl/pull/356 as yet another instance of time waste due to trying to support old julia versions

albheim:
> I added this to #356 as yet another instance of time waste due to trying to support old julia versions

Agreed, I'm all for not spending too much time on keeping up support for old versions. 

One thing here is that StaticArrays is only used on testing/doctesting so it seems like not too big a deal to keep it back solely based on ControlSystems, but if you want to use more packages than ControlSystems there might be other deps on StaticArrays which might then be held back which is not optimal.

olof3:
> I added this to #356 as yet another instance of time waste due to trying to support old julia versions

I also think that the better option would be to drop the support for the older versions.  especially given the time it takes to maintain. Even if it feels quite unsatisfactory. I guess there's not much to do about it.

We have some conditionals in the code depending on the version, if they makes things work better in old versions, I think we could keep them around for a bit longer, but if everything is completely broken anyway, we could get rid of them.

baggepinnen:
> One thing here is that StaticArrays is only used on testing/doctesting so it seems like not too big a deal to keep it back solely based on ControlSystems, but if you want to use more packages than ControlSystems there might be other deps on StaticArrays which might then be held back which is not optimal.

This is the key, if we limit the compat, a user installing ControlSystems will not be able to use the new versions of StaticArrays. A pragmatic option would be to simply write other tests that do not use StaticArrays, but it's nice to have the tests as using StaitcArrays is the goto solution for increasing performance in a large number of cases.

albheim:
Btw, when I moved the travis setup to github actions I dropped the xvfb dependancy since that seemed more annoying to set up and practically implemented what is suggested in #401. If we would like to save that for another time I can remove it from this PR, and otherwise we should remember to close that when merging this.

albheim:
> > One thing here is that StaticArrays is only used on testing/doctesting so it seems like not too big a deal to keep it back solely based on ControlSystems, but if you want to use more packages than ControlSystems there might be other deps on StaticArrays which might then be held back which is not optimal.
> 
> This is the key, if we limit the compat, a user installing ControlSystems will not be able to use the new versions of StaticArrays. A pragmatic option would be to simply write other tests that do not use StaticArrays, but it's nice to have the tests as using StaitcArrays is the goto solution for increasing performance in a large number of cases.

After discussing with @mfalt I'm now running the doctest in the same workflow as building the docs (nicer anyway imo) so they no longer run with tests. This allows us to not set a compat in the main project.toml since we have the separate project.toml for docs. So I have locked StaticArrays to 0.12 in the docs project.toml which fixes the problem with doctest, while in the main project.toml it is not set to any compat. 

What I also noticed is that we only seem to use it at all in doctest so maybe just remove it completely from the main project.toml?

test/runtests.jl
albheim: We had some weird behaviour where makedocs in docs/make.jl failed the doctest (it runs doctest internally, and it printed the errors we had based on the version problems) but managed to do the build and the workflow counted as passed. This was not expected to me and is not documented as far as I can see in Documenter.jl, so for now I run doctest first by itself once and then run docs/make.jl where makedocs runs doctest a second time (this could be disabled for the makedocs).

docs/src/man/introduction.md
```diff
@@ -14,6 +14,8 @@ DocTestSetup = quote
     T = P/(1+P)
     plotsDir = joinpath(dirname(pathof(ControlSystems)), "..", "docs", "build", "plots")
     mkpath(plotsDir)
+    ENV["PLOTS_TEST"] = "true"
+    ENV["GKSwstype"] = "nul"
```
albheim, line 18: Moved this here so we had it working automatically for both makedocs and doctest. As it was before we both had to set this in docs/make.jl and in docs.yml before calling doctest, now both are handled here.

albheim:
> After discussing with @mfalt I'm now running the doctest in the same workflow as building the docs (nicer anyway imo) so they no longer run with tests. This allows us to not set a compat in the main project.toml since we have the separate project.toml for docs. So I have locked StaticArrays to 0.12 in the docs project.toml which fixes the problem with doctest, while in the main project.toml it is not set to any compat.

With this said I still think we can up the supported version, but maybe leave that for #356 so we can get this PR merged?

docs/make.jl
```diff
@@ -1,9 +1,9 @@
 # Set plot globals
 ENV["PLOTS_TEST"] = "true"
-ENV["GKSwstype"] = "100"
+ENV["GKSwstype"] = "nul"
```
albheim, line 3: As suggested in #401 to remove dependency on xvfb. Did it here so I would not have to figure out how to get that working correctly in gh actions just to then have to undo it. If we encounter problem with generating images for the plots we can patch it later.

baggepinnen:
Check seem to be passing, time to pull?

albheim:
> Check seem to be passing, time to pull?

I'm okay with the current state of it.

baggepinnen:
Alright, I'll pull and will deal with eventual consequences tomorrow 😊

Issue #419, opened 2021-01-09
Stable system with unstable step response


KronosTheLate:
So I had this issue a while back, and decided to see if it presists. It does.

My transfer function `TF` is a simple RLC circuit, where the output is the voltage over the capacitor. The bodeplot looks right, and the PZmap and margin plots confirm that it is a stable system. But the step response is exponential. Am I misunderstanding something, or should this be a simple oscillatory response for a short time, followed by a steady-state reference value?

For other examples previously tested, the value went towards a finite output, but it had spikes and irregulatiries forever. I can not create a single example that has a steady output when given a step, other than `TF=tf(1, 1)`. Screenshots of the issue found below:
![image](https://user-images.githubusercontent.com/61620837/104095705-b0398b80-5298-11eb-813e-bf4bf8b4a873.png)
![image](https://user-images.githubusercontent.com/61620837/104095709-b62f6c80-5298-11eb-9317-ff05565e7b76.png)
![image](https://user-images.githubusercontent.com/61620837/104095716-bcbde400-5298-11eb-9251-c978125a2659.png)
![image](https://user-images.githubusercontent.com/61620837/104095723-c6474c00-5298-11eb-8415-fc3e867d51ec.png)

baggepinnen:
I can reproduce your results since you didn't post any code, but I would check the result of converting the tf to a statespace model. Maybe there has been numerical issues in the conversion due to the small coefficients.

KronosTheLate:
All the code is seen in the screenshots, but I have copied it below for copy/pasting:
```
begin
	using ControlSystems, Plots, PlutoUI
	plotlyjs()
	default(:size, (670, 600))
	s = tf("s")
	frac(num, den) = /(num, den)
	oplus(args...) = 1.0/sum(x->1.0/x, args)
	md"Preamble hideout"
end
R, L, C = 1, 4.7e-6, 1.5e-8
TF = frac(1, s^2*L*C+s*R*C+1)
marginplot(TF, ticks=:default)
pzmap(TF)
stepplot(TF)
```

olof3:
Transfer function representations are inherently numerically sensitive and I guess it is generally advisable to stay away from them expect for educational purposes.

[...]

Issue #420, opened 2021-01-10
Function `_default_dt` always returning the same value


albheim:
This code seems to always return 0.07 for a stable continuous system as input and the villain is row 271 where we take the minimum of a bunch of absolut values and 0, where 0 is always less or equal to an absolute value. It seems to be some simple heuristic to get a time step time based on the poles, and everything but the zero seems reasonable to me. 

Looking back in history I see that this row used to be the same but without the zero, which was then added 5 years ago by @mfalt in https://github.com/JuliaControl/ControlSystems.jl/commit/f53b48904de64e97cb2a575fcc5b78706acb3679. Instead of just removing it I thought I would check if you remembered anything, I assume there was some reason behind the change just that it went wrong? Or am I completely missing something here?

https://github.com/JuliaControl/ControlSystems.jl/blob/6c54a259980fc3ac784b45db395edfcd9ea59300/src/timeresp.jl#L264-L277

mfalt:
`minimum` throws an error on an empty collection I'm pretty sure, that's why an extra number is added. One solution is probably to change 0 to Inf in the condition, and then `if r == 0 || r == Inf` since we still need some default when there is a pole at 0.

albheim:
> `minimum` throws an error on an empty collection I'm pretty sure, that's why an extra number is added. One solution is probably to change 0 to Inf in the condition, and then `if r == 0 || r == Inf` since we still need some default when there is a pole at 0.

Ahh, that makes sense. I can put together a PR for that. Should I add some test for that? I mean we probably do run tests that use `_default_dt` but apparently we don't test for this at least.

mfalt:
Yes, a test would be great, something i should have added when i fixed it.

albheim:
> `minimum` throws an error on an empty collection I'm pretty sure, that's why an extra number is added. One solution is probably to change 0 to Inf in the condition, and then `if r == 0 || r == Inf` since we still need some default when there is a pole at 0.

As I mention in the PR, do we really need the `r == 0` case though? Since we know `isstable(sys) == true` in the else clause we know that the real part of all poles is less than 0.

PR #421, opened 2021-01-10
Fix _default_dt

albheim:
See #420

### DIFF
--- a/src/timeresp.jl
+++ b/src/timeresp.jl
@@ -268,8 +268,8 @@ function _default_dt(sys::LTISystem)
         return 0.05
     else
         ps = pole(sys)
-        r = minimum([abs.(real.(ps));0])
-        if r == 0.0
+        r = minimum([abs.(real.(ps)); Inf])
+        if r == 0.0 || r == Inf
             r = 1.0
         end
         return 0.07/r
--- a/src/timeresp.jl
+++ b/src/timeresp.jl
@@ -256,7 +256,7 @@ end
 function _default_time_vector(sys::LTISystem, tfinal::Real=-1)
     dt = _default_dt(sys)
     if tfinal == -1
-        tfinal = 100*dt
+        tfinal = 200dt
     end
     return 0:dt:tfinal
 end
@@ -264,15 +264,12 @@ end
 function _default_dt(sys::LTISystem)
     if isdiscrete(sys)
         return sys.Ts
-    elseif !isstable(sys)
+    elseif all(iszero, pole(sys)) # Static or pure integrators
         return 0.05
     else
-        ps = pole(sys)
-        r = minimum([abs.(real.(ps)); Inf])
-        if r == Inf
-            r = 1.0
-        end
-        return 0.07/r
+        ω0_max = maximum(abs.(pole(sys)))
+        dt = round(1/(12*ω0_max), sigdigits=2)
+        return dt
     end
 end
 
--- a/src/timeresp.jl
+++ b/src/timeresp.jl
@@ -269,7 +269,7 @@ function _default_dt(sys::LTISystem)
     else
         ps = pole(sys)
         r = minimum([abs.(real.(ps)); Inf])
-        if r == 0.0 || r == Inf
+        if r == Inf
             r = 1.0
         end
         return 0.07/r
--- a/test/test_timeresp.jl
+++ b/test/test_timeresp.jl
@@ -179,14 +179,16 @@ y, t2, x = step(G, 2)
 
 
 # Test _default_dt 
-sysstab1 = tf(1, [1, 0.5])
-sysstab2 = tf(1, [1, 0])
+sysstab = tf(1, [1, 0.5])
+sysstatic = zpk([], [], 1.0)
 sysunstab = tf(1, [1, -1])
+sysint = tf(1, [1, 0])
 sysd = tf(1, [1, 1], 0.01)
 
-@test ControlSystems._default_dt(sysstab1) == 0.14
-@test ControlSystems._default_dt(sysstab2) == 0.07
-@test ControlSystems._default_dt(sysinstab) == 0.05
+@test ControlSystems._default_dt(sysstab) == 0.14
+@test ControlSystems._default_dt(sysstatic) == 0.07
+@test ControlSystems._default_dt(sysunstab) == 0.05
+@test ControlSystems._default_dt(sysint) == 0.05
 @test ControlSystems._default_dt(sysd) == 0.01
 
 
--- a/test/test_timeresp.jl
+++ b/test/test_timeresp.jl
@@ -185,9 +185,9 @@ sysunstab = tf(1, [1, -1])
 sysint = tf(1, [1, 0])
 sysd = tf(1, [1, 1], 0.01)
 
-@test ControlSystems._default_dt(sysstab) == 0.14
-@test ControlSystems._default_dt(sysstatic) == 0.07
-@test ControlSystems._default_dt(sysunstab) == 0.05
+@test ControlSystems._default_dt(sysstab) == 0.17
+@test ControlSystems._default_dt(sysstatic) == 0.05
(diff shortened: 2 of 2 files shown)


mfalt:
One test for Inf case, e.g static system would be good

mfalt (approved):

src/timeresp.jl
```diff
@@ -268,8 +268,8 @@ function _default_dt(sys::LTISystem)
         return 0.05
     else
         ps = pole(sys)
-        r = minimum([abs.(real.(ps));0])
-        if r == 0.0
+        r = minimum([abs.(real.(ps)); Inf])
+        if r == Inf
```
albheim, line 272: It seems that checking `r == 0.0` is superfluous since we know that `isstable(sys)` is true in the else clause and this means that no real part of any pole of the systems is greater or equal to zero.

olof3:
Okay, perhaps now is a good opportunity to actually fix `_default_dt` to behave acceptably.

The current heuristic is incorrect and gives severely misleading plots for, e.g., resonant systems, also for unstable systems the behavior is dubious.

Compare `impulseplot(sys)`  and `impulseplot(c2d(sys, 1e-3), 5)` where
```julia
sys = DemoSystems.resonant(ω0=100, ζ=0.01)
```

My feeling is that we should take 
```julia
ω0_max = maximum(abs.(pole(sys)))
dt = round(1/ω0_max/10, sigdigits=2)
```
It seems nice to round.
I seem to recall that 8--10 is the recommend minimum of samples per time-constant, but a few more wouldn't hurt, say 15.
I think it is worse with too few samples, rather than too many. IMO this is just a heuristic to get a first look at the dynamics. 
Perhaps we should introduce `const SAMPLES_PER_TIMECONSTANT = 10` (even if just locally) to stay away from magic numbers.

I don't see why this approach shouldn't be used for unstable systems as well?

For static systems and systems with only integrators, `dt = 0.05` or `dt = 0.1` seems alright, and for this we could check `all(iszero, ps)`, and thus we don't have to worry about an empty or all-zero pole vector for the nominal case?

murrayrm:
If it is useful, you could take a look at the `python-control` functionality [here](https://github.com/python-control/python-control/blob/a82b520e5acb95ae37ebfc2610a3b5065cee7927/control/timeresp.py#L836), which was implemented by Ilhan Polat, with modifications by Sawyer Fuller.

olof3:
> you could take a look at the python-control functionality here
Many thanks for the pointer, definitely a few things that could be useful.

olof3:
@baggepinnen @mfalt I was a bit surprised that step and impulse calls the `DifferentialEquations` solver instead of the internal discrete-timer solver. Is this the intended approach? If we are using `DifferentialEquations` then I guess we don't need to worry about `dt` (and perhaps not even the final time)?

baggepinnen:
The internal solver does not handle the case when u is a function. If u is a vector, we might as well use the discrete solver.

albheim:
Just want to say that I might not have a lot of time over in the coming weeks to work more on this, so if anyone is keen on doing more improvements you are very welcome to take over this PR. Otherwise I would suggest merging this and opening an issue for discussing what should be done more that this, just so we can fix this bug that should affect quite a lot of code.

olof3:
Sure, sounds good. But should it be maximum rather than minimum? I guess we want `dt` to be short enough to capture the fast dynamics. I think that would be good to fix while you're at it.

Essentially these two lines that make it a little bit more readable

```
ω0_max = maximum(abs.(pole(sys)))
dt = round(1/ω0_max/12, sigdigits=2)
```

albheim:
> Sure, sounds good. But shouldn't it be `maximum` rather than `minimum`? I guess we want `dt` to be short enough to capture the fast dynamics. I think it would be good to fix that while you're at it.
> 
> Essentially these two lines that make it a little bit more readable
> 
> ```
> ω0_max = maximum(abs.(pole(sys)))
> dt = round(1/ω0_max/12, sigdigits=2)
> ```

Given how `_default_time_vector` is implemented (default is `0:dt:100dt`) it seems like using `minimum` is a way to make it capture the range of time while the system is still moving, though I'm not saying I think this is a good way. I just tried to fix the error where it always returned the same value.

A better way would probably be to choose `dt` based on the fast poles while `tfinal` is chosen based on the slow ones, but as they mention in the `python-control` implementation just doing this naively might not be so good.

I don't really have a strong opinion, so if you think it would be good to change this already now I'm fine with that. Just thought I would mention that the `_default_time_vector` might return some very short timespans then.

olof3:
> The reason for the Inf in the minimum is to be able to handle static systems which will otherwise throw an error, so you would have that problem here also. Though it might be nicer to handle that as its own case instead of this hack with adding a check for Inf.

I think I mentioned that in the previous post, but I didn't copy it again.

> Given how _default_time_vector is implemented (default is 0:dt:100dt) it seems like using minimum is a way to make it capture the range of time while the system is still moving, though I'm not saying I think this is a good way. I just tried to fix the error where it always returned the same value.

Yes this seems quite questionable. Will your fix also settle #419?

IMO opinion it would be better to go for the fastest time constant (obviously this is also very problematic (just as you say), and something smarter should really be done, perhaps using a solver with adaptive step length selection). However I think that would still be better than what we currently have which gives completely wrong plots. 

Anyway, feel free to fix what you came and open a new issue for further discussions.

Some thoughts.
* It seems like a hard problem and I guess we should be happy with something that works most of the time.
* For plotting it seems like it would be acceptable to use a relatively large number of samples (many more than a hundred, which seems to be the current default).
* I guess the `tfinal` should preferably not be decided a priori, but rather from using some kind of stopping criterion, perhaps some combination of rate of change and convergence to G(0).

albheim:
> I think I mentioned that in the previous post, but I didn't copy it again.

True, I think that looks reasonable.
 
> IMO opinion it would be better to go for the fastest time constant (obviously this is also very problematic (just as you say), and something smarter should really be done, perhaps using a solver with adaptive step length selection). However I think that would still be better than with what we currently have (which might give completely wrong plots).

Sure, I'll try with that and push it if it works nice enough.

> * For plotting it seems like it would be acceptable to use a relatively large number of samples (many more than a hundred, which seems to be the current default).

Okay, I'll try to increase the default a bit.

> * I guess the `tfinal` should preferably not be decided a priori, but rather from using some kind of stopping criterion, perhaps some combination of rate of change and convergence to G(0).

This sounds like a nice solution if it is possible to get this to work nicely with DifferentialEquations.jl, which I have no clue if it is. But might be worth looking into.

olof3:
Looks good. Seems like it fixes #419. Although there will be many situations where also this approach will be very unsatisfactory.

@baggepinnen @mfalt, any thoughts?

albheim:
> Looks good. Seems like it fixes #419. Although there will be many situations where also this approach will be very unsatisfactory.
> 
> @baggepinnen @mfalt, any thoughts?

Does #419 work for you now? I just tested and it does not error, but it does not seem to give the result I expect.

olof3:
I think that you are right I was happy that it gave some kind of more reasonable output than the last time. Did you also get a few oscillation periods of increasing amplitude?

albheim:
> I think that you are right that it's still wrong, I was just happy that it gave some kind of more reasonable output than the last time. Did you also get a few oscillation periods of increasing amplitude?

Yes, exactly.

I did look at matlabs state space transformation of the transfer function and it is rather different than our. Trying with their state space instead it gives similar results as matlab and something that seems more a step response for a fast oscillatory system.

```
A = [-2.128e5 -3.382e6; 4.194e6 0]
B = [2048; 0]
C = [0 1651]
matlabsys = ss(A, B, C, 0)
stepplot(matlabsys)
y, t, x = step(matlabsys, range(0, stop=5e-5, length=300))
plot(t, y)
```

olof3:
I continued the discussion in #419. I think this is good enough to be merged.

albheim:
> I continued the discussion in #419. I think this is good enough to be merged.

Yes, I feel this is good enough for now.

PR #434, opened 2021-01-17
QOL improvements for plotting

baggepinnen:
This PR implements some minor fixes to make plotting nicer. In particular, the weird gain circles in `nyquist` are replaced for a circle indicating `Ms = 2` (default, but can be changed by kwarg).

### DIFF
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -748,7 +748,7 @@ pzmap
                 real(z),imag(z)
             end
         end
-        if !isempty(p[1])
+        if !isempty(p)
             @series begin
                 group --> i
                 markershape --> :xcross
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -377,7 +377,7 @@ optionally provided.
 `gaincircle` plots the circle corresponding to |T(iω)| = 1, where `T` is
 the complementary sensitivity function.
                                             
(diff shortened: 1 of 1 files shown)


baggepinnen:
@JuliaControlBot test-plots

baggepinnen:
Anyone have anything against the cosmetic changes to `nyquistplot`? The old appearance can be recovered with `Ms=1, framestyle = :semi`, but I think the defaults here are a bit more useful.

[framestyle docs](http://docs.juliaplots.org/latest/generated/gr/#gr-ref34 )

olof3:
Definitely improvements

Some thoughts: 

`Ms = 2` is considered quite high in many circumstances, but it is hard to give a sensible default value. At least 2 is a lot more meaningful than 1, as well as nice and even :). It could be considered to leave it out as default or have a few different Ms circles (not sure how feasible it is to label them).

I would vote for not plotting the circle around the origin as default. It is useful if you would like to eyeball the phase margin, but as one in most circumstances "should" be considering the maximum sensitivity, it is a distraction most of the time.

baggepinnen:
New default Ms=1.5, with option to supply a vector of values. Value appears on hover in plotly, couldn't get annotations to work properly. T gain circle removed.

baggepinnen:
@JuliaControlBot test-plots

olof3:
Great!

I realize now that the critical point isn't indicated, a nicely sized red cross would be nice.

There seems to be some inconsistencies in the plot test? Titles and circles.

It feels a little bit weird with Nyquist plots like these for MIMO systems. The MIMO nyquist criterion is a different thing. But perhaps this could occasionally be useful (not that the MIMO nyquist criterion is overly useful either).

mfalt:
@JuliaControlBot test-plots

PR #437, opened 2021-01-27
Removed the matrix type parameter from StateSpace, closes #413

### LINKED ISSUE #413: Get rid of the Matrix type parameter in StateSpace
Now that we have `HeteroStateSpace` we could get rid of the Matrix type parameter in `StateSpace` which has been quite painful to deal through promotion etc, while it is too inflexible to actually add much value. I think we have thought about this before, and sort of concluded to go ahead on this? @mfalt @baggepinnen

olof3:
Now that we have HeteroStateSpace, there is no need for the matrix parameter in the StateSpace type. It turned out to be that it wasn't that useful since one common matrix type didn't give enough flexibility. This makes the type signature for the 99% of the cases that use StateSpace a lot nicer.

Closes #413

### DIFF
--- a/src/types/conversion.jl
+++ b/src/types/conversion.jl
@@ -19,37 +19,34 @@
 # Base.convert(::Type{<:TransferFunction{<:SisoRational}}, b::Number) = tf(b)
 # Base.convert(::Type{<:TransferFunction{<:SisoZpk}}, b::Number) = zpk(b)
 #
-Base.convert(::Type{TransferFunction{TE,SisoZpk{T1, TR1}}}, b::AbstractMatrix{T2}) where {TE, T1, TR1, T2<:Number} =
-    zpk(T1.(b), undef_sampletime(TE))
-Base.convert(::Type{TransferFunction{TE,SisoRational{T1}}}, b::AbstractMatrix{T2}) where {TE, T1, T2<:Number} =
-    tf(T1.(b), undef_sampletime(TE))
+Base.convert(::Type{TransferFunction{TE,SisoZpk{T1, TR1}}}, D::AbstractMatrix{T2}) where {TE, T1, TR1, T2<:Number} =
+    zpk(convert(Matrix{T1}, D), undef_sampletime(TE))
+Base.convert(::Type{TransferFunction{TE,SisoRational{T1}}}, D::AbstractMatrix{T2}) where {TE, T1, T2<:Number} =
+    tf(Matrix{T1}(D), undef_sampletime(TE))
 
-function convert(::Type{StateSpace{TE,T,MT}}, D::AbstractMatrix{<:Number}) where {TE,T, MT}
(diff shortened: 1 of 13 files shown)


src/types/conversion.jl
```diff
@@ -97,7 +94,7 @@ function Base.convert(::Type{StateSpace{TE,T,MT}}, G::TransferFunction) where {T
     # could be much cleaner.
     #T = Base.promote_op(/, T0, T0)
 
-    Ac = Bc = Cc = Dc = A = B = C = D = Array{T}(undef, 0, 0)
+    Ac = Bc = Cc = Dc = A = B = C = D = Matrix{T}(undef, 0, 0)
```
baggepinnen, line 97: Could probably declare `local Ac , Bc , Cc , Dc , A , B , C , D`
instead of creating empty matrices (unrelated to this PR so feel free to not fix it here if you don't feel like it).

baggepinnen (commented):
Looks good to me. There are a lot of changed lines (conversion/promotion etc.) that are not covered by the tests. These were not tested before either so it's not really a problem with this PR per se, but are we confident that we are not introducing new conversion/promotion bugs with these changes?

olof3:
I added a few more tests. It seems like the promote_rules for system and number aren't really used.

It also seems like the existing tests for promotion are a bit flawed, e.g.,
```
@test promote(ss(1), ss(1.)) == (ss(1.), ss(1.)
````
as this would be true even if no promotion was taking place. But that would be for another issue.

**Comment by albheim** [comment]
> It also seems like the existing tests for promotion are a bit flawed, e.g.,
> 
> ```
> @test promote(ss(1), ss(1.)) == (ss(1.), ss(1.)
> ```
> 
> as this would be true even if no promotion was taking place. But that would be for another issue.

`ss(1)` and `ss(1.)` gives me different types, one based on Int64 and one on Float64, so I think promotion is needed then?

**Comment by olof3** [comment]
> ss(1) and ss(1.) gives me different types, one based on Int64 and one on Float64, so I think promotion is needed then?

The issue is that also
```
@test (ss(1), ss(1.)) == (ss(1.), ss(1.))
```
i.e., the test passes even if not promotion would take place.

albheim:
> > ss(1) and ss(1.) gives me different types, one based on Int64 and one on Float64, so I think promotion is needed then?
> 
> The issue is that
> 
> ```
> (ss(1), ss(1.)) == (ss(1.), ss(1.))
> ```
> 
> returns true, i.e., the test passes even if not promotion would take place.

Ahh, makes sense.

olof3:
> Ahh, makes sense. Maybe should test if type is equal instead of testing value then?
Yes, or both. Perhaps even better if test the methods that rely on the promote rules. But that would be for another issue.

@baggepinnen I think this should be good enough to be pulled.

Issue #441, opened 2021-02-03
c2d does not maintain type


albheim:
Noticed that when using `c2d` on a `zpk` system it changes type in the returned system. Seems like it could be nice to have the type stay the same.

```julia
julia> G = tf(1.0, [1, 1])
TransferFunction{Continuous, ControlSystems.SisoRational{Float64}}
   1.0
----------
1.0s + 1.0

Continuous-time transfer function model

julia> c2d(G, 0.1)
TransferFunction{Discrete{Float64}, ControlSystems.SisoRational{Float64}}
   0.09516258196404037
-------------------------
1.0z - 0.9048374180359594

Sample Time: 0.1 (seconds)
Discrete-time transfer function model
```

olof3:
Absolutely, this is something that should be fixed.

I guess the easy way is just to add a conversion, but I guess we really should use [pole--zero mapping](https://en.wikipedia.org/wiki/Matched_Z-transform_method). Seems to be pretty straightforward to implement. I guess there are other conversion methods  that I'm not aware of, but I think this is a very common one.

albheim:
Gave a conversion a quick try and got a bit confused at how to handle types correctly. Have not tried the pole zero mapping yet but seems like it should be rather easy. But should not this be an option for the `method` keyword in `c2d` rather than being the automatic option? Or maybe it could be the default for a `zpk` system, but also be available as a keyword for other system representations?

albheim:
Looked at the pole zero matching and implemented it for zpk types, and it seems to work. The thing I am a bit unsure about is how to nicely handle the dispatch when this is added. Before everything would be discretized through state space and there is the `c2d_x0map` that contains all the different transformations, and transferfunctions are just transformed and sent here. If we now have pz matching it seems like this would mess up this structure a bit and might need some more work than I was up for right now (i.e. we dont want pz matching to be in c2d_x0map since then we will transform zpk->ss->zpk in the process.

PR #444, opened 2021-02-15
Maintain type for zpk in c2d

albheim:
Simple type conversion as mentioned in #441

### DIFF
--- a/src/discrete.jl
+++ b/src/discrete.jl
@@ -227,11 +227,11 @@ function c2d_poly2poly(p,h)
 end
 
 
-function c2d(G::TransferFunction{<:Continuous, T}, h, args...; kwargs...) where {T<:SisoTf}
+function c2d(G::TransferFunction{<:Continuous, T}, h, args...; kwargs...) where {T}
     @assert issiso(G) "c2d(G::TransferFunction, h) not implemented for MIMO systems"
     sys = ss(G)
     sysd = c2d(sys, h, args...; kwargs...)
-    return convert(TransferFunction{typeof(sysd.timeevol), Base.typename(T)}, sysd)
+    return convert(TransferFunction{typeof(sysd.timeevol), Base.typename(T).wrapper}, sysd)
 end
 
 """
--- a/src/discrete.jl
+++ b/src/discrete.jl
@@ -228,7 +228,7 @@ end
 
 
 function c2d(G::TransferFunction{<:Continuous, T}, h, args...; kwargs...) where {T}
-    #@assert issiso(G) "c2d(G::TransferFunction, h) not implemented for MIMO systems"
+    @assert issiso(G) "c2d(G::TransferFunction, h) not implemented for MIMO systems"
     sys = ss(G)
     sysd = c2d(sys, h, args...; kwargs...)
     return convert(TransferFunction{typeof(sysd.timeevol), Base.typename(T).wrapper}, sysd)
--- a/src/discrete.jl
+++ b/src/discrete.jl
@@ -228,10 +228,13 @@ end
 
 
 function c2d(G::TransferFunction{<:Continuous, T}, h, args...; kwargs...) where {T}
-    @assert issiso(G) "c2d(G::TransferFunction, h) not implemented for MIMO systems"
+    @assert issiso(G) "c2d(G::TransferFunction, h) not implemented for MIMO systems" 
     sys = ss(G)
     sysd = c2d(sys, h, args...; kwargs...)
-    return convert(TransferFunction{typeof(sysd.timeevol), Base.typename(T).wrapper}, sysd)
+    # Extract the basic type (SisoRational/SisoZpk) without the parametric values
+    # This allows the type promotion to be handled by convert
+    BT = Base.typename(T).wrapper 
+    return convert(TransferFunction{typeof(sysd.timeevol), BT}, sysd)
 end
 
 """
--- a/src/discrete.jl
+++ b/src/discrete.jl
@@ -227,12 +227,11 @@ function c2d_poly2poly(p,h)
 end
 
 
-function c2d(G::TransferFunction{<:Continuous}, h, args...; kwargs...)
-    ny, nu = size(G)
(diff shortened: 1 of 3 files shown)


src/discrete.jl
```diff
@@ -227,12 +227,11 @@ function c2d_poly2poly(p,h)
...
 
-function c2d(G::TransferFunction{<:Continuous}, h, args...; kwargs...)
-    ny, nu = size(G)
-    @assert (ny + nu == 2) "c2d(G::TransferFunction, h) not implemented for MIMO systems"
+function c2d(G::TransferFunction{TE, T}, h, args...; kwargs...) where {TE<:Continuous, T<:SisoTf}
+    @assert issiso(G) "c2d(G::TransferFunction, h) not implemented for MIMO systems"
     sys = ss(G)
     sysd = c2d(sys, h, args...; kwargs...)
-    return convert(TransferFunction, sysd)
+    return convert(TransferFunction{typeof(Discrete(h)), T}, sysd)
```
albheim, line 234: Not sure this is the way to do it, for example if `T` is based on `Int` there are some problems? Can I say that I want to convert to only the first part of `T`, i.e. the `SisoZpk` or `SisoRational` and allow for automatically finding the numeric base?

olof3: I guess you can do `convert(TransferFunction{typeof(Discrete(h)), SisoZpk}, sysd)`, etc, and then either have two functions (or an if statement in this function) (perhaps you already knew this). I guess that there is a better way, but I think it would be good to avoid something that is too tricky.

Perhaps slightly better with `typeof(Discrete(h)` -> `typeof(sysd.timeevol)`

Couldn't we just use `zpk` / `tf` to avoid fiddling with the types? Although it in general feels good to be explicit with the types.

albheim: I was thinking of something along those lines as an option, to have the input type as a dispatch. And this might be good if we want to specialize on how to handle different methods depending on input type also (like handling pole matching for zpk should not go through a state space). 

But for now I felt like there should be a nice way of doing it in one function and tried to figure that out. Maybe not that important.

albheim: Found `Base.typename` which seems to extract only the base type without the parametric types.

Also switched to `typeof(sysd.timeevol)`, seems nicer.

albheim: Seems like it is not doing exactly what I thought, here is a discussion about doing somethign similar https://discourse.julialang.org/t/stripping-parameter-from-parametric-types/8293/13

src/discrete.jl
```diff
@@ -227,12 +227,11 @@ function c2d_poly2poly(p,h)
 end
 
 
-function c2d(G::TransferFunction{<:Continuous}, h, args...; kwargs...)
-    ny, nu = size(G)
-    @assert (ny + nu == 2) "c2d(G::TransferFunction, h) not implemented for MIMO systems"
+function c2d(G::TransferFunction{<:Continuous, T}, h, args...; kwargs...) where {T}
```
albheim, line 230: Is it better to say explicitly that `T <: SisoTf` here or is it better to leave it blank? Does it make any difference in this case?

src/discrete.jl
```diff
@@ -227,12 +227,11 @@ function c2d_poly2poly(p,h)
...
+    @assert issiso(G) "c2d(G::TransferFunction, h) not implemented for MIMO systems"
     sys = ss(G)
     sysd = c2d(sys, h, args...; kwargs...)
-    return convert(TransferFunction, sysd)
+    return convert(TransferFunction{typeof(sysd.timeevol), Base.typename(T).wrapper}, sysd)
```
albheim, line 234: This will only get the base type of `T`, e.g. `SisoZpk`, without the parametric values. This allows the already existing `convert` methods to figure out the correct numeric types.

src/discrete.jl
```diff
@@ -227,12 +227,11 @@ function c2d_poly2poly(p,h)
 end
 
 
-function c2d(G::TransferFunction{<:Continuous}, h, args...; kwargs...)
-    ny, nu = size(G)
-    @assert (ny + nu == 2) "c2d(G::TransferFunction, h) not implemented for MIMO systems"
+function c2d(G::TransferFunction{<:Continuous, T}, h, args...; kwargs...) where {T}
+    @assert issiso(G) "c2d(G::TransferFunction, h) not implemented for MIMO systems"
```
albheim, line 231: Is this really needed? Tried commenting it out and running with some random MIMO system
```
sys = ss(...)
stepplot([sys, c2d(zpk(sys), 0.1)])
```
and seemed to work fine.

olof3: Yeah, I was wondering about the same thing. Good that you tried it out. Add a some tests and drop it.

> ss(zpk(ss(...))) on a MIMO system made it explode in states a bit, so maybe that is the reason for this?

Perhaps. Anyway, one should try to stay away from all these transfer function representations in most circumstances.

src/discrete.jl
olof3: That looks somewhat tricky, I don't think I would not have been able to decode that. An option could be `T <: SisoZpk ? SizoZpk : SisoRational`, which I think is less elegant but a lot more readable.

The whole issue would go away if we have separate conversions methods for TransferFunction{TE,SisoZpk} and TransferFunction{TE,SisoRational}. Perhaps we like to have that anyway, in order to cover pole--zero matching as well as other c2d conversions in the best way possible? I haven't thought about it much though.

src/discrete.jl
albheim: Saw there was a broken test to check that type of a zpk run through c2d should be a zpk still. Just uncommented that for now.

albheim: I didn't really feel like coming up with tests for the mimo cases right now so added back the assertion for the time being.

src/discrete.jl
albheim: Added explanation for now

src/discrete.jl
albheim: It seems we can get some state explosion, so maybe that is the reason to not allow automatic conversion between tf mimo and state space.

PR #446, opened 2021-02-16
Change all sample time to Ts

### LINKED ISSUE #445: Inconsistent internal notation for sampling time
Saw there was a mix of `h` and `Ts` mainly in the discrete.jl but some other places also. Nothing major but nice with consistency in my opinion, I can go through and change if there is no specific reason not to. I assume `Ts` is the one we want since that seemed to be used more?

albheim:
Fixes #445

### DIFF
--- a/src/pid_design.jl
+++ b/src/pid_design.jl
@@ -150,21 +150,21 @@ rlocus
 end
 
 """
-    laglink(a, M; h=0)
+    laglink(a, M; Ts=0)
 
 Returns a phase retarding link, the rule of thumb `a = 0.1ωc` guarantees less than 6 degrees phase margin loss. The bode curve will go from `M`, bend down at `a/M` and level out at 1 for frequencies > `a`
 """
-function laglink(a, M; h=0)
+function laglink(a, M; Ts=0)
     numerator = [1/a, 1]
     denominator = [M/a, 1]
     gain = M
     G = tf(gain*numerator,denominator)
-    return  h <= 0 ? G : c2d(G,h)
+    return  Ts <= 0 ? G : c2d(G,Ts)
 end
 
 
 """
-    leadlink(b, N, K; h=0)
+    leadlink(b, N, K; Ts=0)
 
 Returns a phase advancing link, the top of the phase curve is located at `ω = b√(N)` where the link amplification is `K√(N)` The bode curve will go from `K`, bend up at `b` and level out at `KN` for frequencies > `bN`
 
@@ -174,17 +174,17 @@ Values of `N < 1` will give a phase retarding link.
 
 See also `leadlinkat` `laglink`
 """
-function leadlink(b, N, K; h=0)
+function leadlink(b, N, K; Ts=0)
     numerator = [1/b, 1]
     denominator = [1/(b*N), 1]
     gain = K
     G = tf(gain*numerator,denominator)
-    return  h <= 0 ? G : c2d(G,h)
+    return  Ts <= 0 ? G : c2d(G,Ts)
 
 end
 
 """
-    leadlinkat(ω, N, K; h=0)
+    leadlinkat(ω, N, K; Ts=0)
 
 Returns a phase advancing link, the top of the phase curve is located at `ω` where the link amplification is `K√(N)` The bode curve will go from `K`, bend up at `ω/√(N)` and level out at `KN` for frequencies > `ω√(N)`
 
@@ -194,9 +194,9 @@ Values of `N < 1` will give a phase retarding link.
 
 See also `leadlink` `laglink`
 """
-function leadlinkat(ω, N, K; h=0)
+function leadlinkat(ω, N, K; Ts=0)
     b = ω / sqrt(N)
-    return leadlink(b,N,K,h=h)
+    return leadlink(b,N,K,Ts=Ts)
 end
 
 """
--- a/src/pid_design.jl
+++ b/src/pid_design.jl
@@ -154,7 +154,12 @@ end
 
 Returns a phase retarding link, the rule of thumb `a = 0.1ωc` guarantees less than 6 degrees phase margin loss. The bode curve will go from `M`, bend down at `a/M` and level out at 1 for frequencies > `a`
 """
-function laglink(a, M; Ts=0)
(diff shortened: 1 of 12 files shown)


baggepinnen (approved):
LGTM
This changes the API to some methods and is thus a breaking change. We should add a news entry explaining the change

mfalt:
> LGTM
> This changes the API to some methods and is thus a breaking change. We should add a news entry explaining the change

I agree, but it looks simple enough to add a deprecated version of the lead and lag method with the old kwargs. I think that would be nice to avoid any unexpected breakage for a few version.

src/simulators.jl
```diff
@@ -23,9 +23,9 @@ Used to simulate continuous-time systems. See function `?solve` for additional i
 # Usage:
 ```
 using OrdinaryDiffEq
-h              = 0.1
+Ts             = 0.1
 Tf             = 20
```
olof3, line 27: In this case `Ts` is not a sample time, but the spacing of a time vector, so imo it should be `dt`, not sure if there are other instances of this. Also, a more clear-cut case of problematic inconsistency is `Tf` vs `tfinal`, where I am greatly in favor of the latter.

src/pid_design.jl
```diff
@@ -150,21 +150,21 @@ rlocus
...
 Returns a phase retarding link, the rule of thumb `a = 0.1ωc` guarantees less than 6 degrees phase margin loss. The bode curve will go from `M`, bend down at `a/M` and level out at 1 for frequencies > `a`
 """
-function laglink(a, M; h=0)
+function laglink(a, M; Ts=0)
     numerator = [1/a, 1]
     denominator = [M/a, 1]
     gain = M
     G = tf(gain*numerator,denominator)
-    return  h <= 0 ? G : c2d(G,h)
+    return  Ts <= 0 ? G : c2d(G,Ts)
```
olof3, line 162: I guess nothing is better than 0 as a default value to signal continuous time.

albheim: Yeah, I was thinking about this also but seemed like a small enough thing to not really care about. But maybe one could dispatch on whether Ts is present or not and handle it as in other places with Continuous/Discrete types if we want to handle non-positive Ts as we do in other places?

src/simulators.jl
albheim: Agreed. Has there also been a previous discussion on the `Tf`vs `tfinal`? I would say that `Tf` seems more consistent with `Ts` so just for that reason I might prefer that. But I don't really have a strong opinion.

albheim:
> I agree, but it looks simple enough to add a deprecated version of the lead and lag method with the old kwargs. I think that it would be nice to avoid any unexpected breakages using that for a few versions.

How would I go about this. Tried to have a look at the `@deprecate` and figured I should add
```julia
@deprecate laglink(a, M; h=0) laglink(a, M; Ts=h)
```
in `ControlSystems.jl` while keeping both versions of the methods. I realised that the methods overwrite each other and I assume this is since they only differ in keyword arguments. How do you deprecate on keyword args in a nice way?

baggepinnen:
> but it looks simple enough to add a deprecated version of the lead and lag method with the old kwargs. I think that it would be nice to avoid any unexpected breakages using that for a few versions.

Deprecation warnings are off by default now, so users will never see them unless they explicitly start julia with `--depwarn=yes`. This makes them somewhat less useful than they used to be.

albheim:
Okay, looking at the deprecate code it seems like it will overwrite if both method signatures are the same. 

https://github.com/JuliaLang/julia/blob/9c31a1b7e35334b0d3e209e1aec4cb79a7f8d507/base/deprecated.jl#L68-L73

I think you could easily do something ugly like having both `Ts` and `h` as kwargs and just do some checking of which of them is not zero there and call `Base.depwarn` if `h` is used.

src/simulators.jl
olof3: Lower case t for final time is standard in the literature, and consistent with how we denote time vectors. `Tf` clahses with `tf`

src/pid_design.jl
olof3: >  if we want to handle non-positive Ts as we do in other places

If we do, then I don't think we should.

albheim: I'm not sure what I wrote there, I assume I just wanted to say "dispatch on Ts present or not and use Continuous/Discrete types as we do in other places".

src/simulators.jl
albheim: Then that sounds good to me. So I saw you writing `tf`, `t_f` and `tfinal`, just wanted to make sure that `tfinal` is what you preferred?

Should I make a quick run through the code and update those also for this PR?

olof3: Yes, `tfinal`. That would be very much appreciated.

albheim: Had a quick look and it seems like currently `Tf` is used quite a lot more than `tfinal`. But I ran through them and now it should be consistent at least.

albheim:
As @olof3 mentioned there was also an inconsistency between `Tf` and `tfinal` where `Tf` was currently used more but was not preferred for him. I have no strong opinion, I think `Tf` can be nice since it is more similar to `Ts` but `tfinal` seems like a nicer variable name overall so for now I went with his suggestion and updated that as well. 

I also had a go at adding some deprecations as @mfalt suggested, not sure it is implemented very nicely but I tested them and if you have `--depwarn yes` it shows the message when using `h`. They are not in the ControlSystems.jl file as the other deprecations seems to be, but I added a comment there at least so you can find them from there.

olof3:
> I think Tf can be nice since it is more similar to Ts

I would argue that this is not necessarily a good thing and rather a reason to consider something else.

Similarity with the function `tf`, uppercase for a scalar real variable, and inconsistency with the literature are aspects in favor of not using Tf.

albheim:
> Similarity with the function `tf`, uppercase for a scalar real variable, and inconsistency with the literature are aspects in favor of not using Tf.

Yeah, sounds reasonable. This is pretty done in my eyes unless someone has other opinions on the change?

olof3:
@baggepinnen @mfalt Time to pull this?

Will be a good improvement in terms of consistency. Even though `h` looked more natural to me in some of the cases :)

PR #449, opened 2021-02-19
New method for feedback2dof

baggepinnen:


baggepinnen:
Should maybe add a block diagram in the docstring
https://asciiflow.com/
seems useful for drawing them

src/connections.jl
```diff
@@ -318,6 +318,19 @@ end
...
+Return the closed-loop system with process `P0`, controller `C` and feedforward filter `F` from reference to control signal (by-passing `C`).
+"""
+function feedback2dof(P0::TransferFunction, C::TransferFunction, F::TransferFunction)
+    !issiso(P0) && error("Feedback not implemented for MIMO systems")
+    N,D = numpoly(P0)[], denpoly(P0)[]
+    P,Q = numpoly(C)[], denpoly(C)[]
+    W,R = numpoly(F)[], denpoly(F)[]
+    den = (Q*D + N*P)*R
+    tf(Q*N*W + N*P*R, den)
+end
```
olof3, line 333: I'm having a little bit of a hard time to figure this one out.

What is the transfer function that is being calculated? Would be good to write that explicitly in the docstring, II would have guessed `inv(1 + PC) * F`, but I'm not sure that is correct?

Is the purpose of the function to set up the desired transfer functions without excessive poles/zeros that cancel each other out?

I would perhaps prefer denoting numerators/denominators with e.g., F_den, F_num, or numF, denF, or something like that, now, there are too many letters to keep track of, then `P` can be used as an input argument.

I guess that this is only for Continuous time, that could perhaps be indicated in the type parameters.

I should be able to give some more comments, once I understand this a little bit better.

baggepinnen: Thanks for the comments, it should hopefully be much improved now. The idea is to get the desired transfer function with better numerical properties than the naive approach. There could potentially be more of these methods added, i.e., for the common case 
```
  +------+     +------+      +------+
r |      |  -  |      |      |      |
--|   F  +----->   C  +------>   P  +--+-->
  |      |  |  |      |      |      |  |
  +------+  |  +------+      +------+  |
            |                          |
            +--------------------------+

```

src/connections.jl
```diff
@@ -318,6 +318,34 @@ end
...
+      +--------------------------------+
+"""
+function feedback2dof(P::TransferFunction{T}, C::TransferFunction{T}, F::TransferFunction{T}) where T
+    !issiso(P) && error("Feedback not implemented for MIMO systems")
+    Pn,Pd = numpoly(P)[], denpoly(P)[]
+    Cn,Cd = numpoly(C)[], denpoly(C)[]
+    Fn,Fd = numpoly(F)[], denpoly(F)[]
+    den = (Cd*Pd + Pn*Cn)*Fd
+    isdiscrete(P) ? tf(Cd*Pn*Fn + Pn*Cn*Fd, den, P.Ts) : tf(Cd*Pn*Fn + Pn*Cn*Fd, den)
+end
```
olof3, line 348: I hope the following works, I think it's how @mfalt and I tried to use it for other types of connections. It handles promotion of sample times and checks for sample-time mismatch.
```suggestion
function feedback2dof(P::TransferFunction{TE}, C::TransferFunction{TE}, F::TransferFunction{TE}) where TE
    !issiso(P) && error("Feedback not implemented for MIMO systems")
    timeevol = common_timeevol(P, C, F)
    
    Pn,Pd = numpoly(P)[], denpoly(P)[]
    Cn,Cd = numpoly(C)[], denpoly(C)[]
    Fn,Fd = numpoly(F)[], denpoly(F)[]
    den = (Cd*Pd + Pn*Cn)*Fd
    tf(Cd*Pn*Fn + Pn*Cn*Fd, den, timeevol)
end
```

src/connections.jl
olof3: Great. Yes, I was just going to comment on that the figure above is what I primarily associate with "2DOF controller design".

I think it looks good otherwise.

PR #455, opened 2021-03-16
Loosen `c2d` system type from `StateSpace` to `AbstractStateSpace`

imciner2:
Using the abstract `AbstractStateSpace` type allows for `HeteroStateSpace` to use the `c2d` function as well. The `d2c` function is already using the abstract type, so this makes the two functions symmetric.


baggepinnen (approved):
LGTM
Thanks for the PR :)

mfalt (approved):

PR #457, opened 2021-03-21
Implement `luenberger`

Aocci:
Implements a wrapper around `place` which allows for a more convenient user workflow when performing pole-placement for observer design.


baggepinnen (approved):

mfalt:
Looks good to me. Thanks for the PR!

I would probably have skipped the text about how the method is implemented, or changes it to "This method is equivalent to calling place(A', C', p)'", but that's a minor detail. I noticed some other issues in this file, but opened that as a separate issue https://github.com/JuliaControl/ControlSystems.jl/issues/458

Issue #462, opened 2021-03-30
`similarity_transform(sys, T)` throws an error when transformation matrix `T` is complex


tfoliva:
Using Version 0.9.2.

Trying to perform a similarity transformation with a complex matrix returns 
```
ERROR: InexactError: Float64(-5.159983634948731 - 4116.475358187984im)
Stacktrace:
 [1] Real at ./complex.jl:37 [inlined]
 [2] convert at ./number.jl:7 [inlined]
 [3] setindex! at ./array.jl:847 [inlined]
 [4] _unsafe_copyto!(::Array{Float64,2}, ::Int64, ::Array{Complex{Float64},2}, ::Int64, ::Int64) at ./array.jl:257
 [5] unsafe_copyto! at ./array.jl:311 [inlined]
 [6] _copyto_impl! at ./array.jl:335 [inlined]
 [7] copyto! at ./array.jl:321 [inlined]
 [8] copyto! at ./array.jl:347 [inlined]
 [9] copyto_axcheck! at ./abstractarray.jl:946 [inlined]
 [10] Array{Float64,2}(::Array{Complex{Float64},2}) at ./array.jl:562
 [11] convert at ./array.jl:554 [inlined]
 ```

I needed to convert a system to modal canonical form, so I implemented a similar function which performs the eigendecomposition, (similar to MATLAB's `prescale` I guess). I can make a pull request. Furthermore, are there plans to implement this system transformations (controlable canonical form, observable canonical form, modal canonical form)?

baggepinnen:
Hey and thanks for the issue. Feel free to make such PR, that'd be great. The implementation effort going into this package is mostly driven by need, so unless someone needs it and implements it, it's unfortunately unlikely to happen.

PR #464, opened 2021-04-01
feat: add `prescale(sys::StateSpace)` and corresponding test.

tfoliva:
Adds a `prescale` function as mentioned in #462


test/test_matrix_comps.jl
```diff
@@ -70,6 +70,12 @@ syst = similarity_transform(sys, Tr)
 @test sys.B ≈ Tr*syst.B
 @test sys.C*Tr ≈ syst.C
 
+nsys, T = prescale(sys)
+@test isdiagonal(nsys.A)
```
tfoliva: should be `isdiag`

tfoliva (commented):
Fixes the `isdiag` test

tfoliva:
I noticed that somehow the `test_matrix_comps` file does not recognize the change I made to `matrix_comps` when defining the new function `prescale`. I got the same behavior locally at my machine, It seems that the tests import the official ControlSystems package version from the Julia registry. How can I perform the tests considering the changes I made?

baggepinnen:
You have forgotten to export the name of the new function, that's why it's not found in the tests. Where does the name come from? Is it a standard name?

tfoliva (commented):
exported previously written functions

tfoliva:
> You have forgotten to export the name of the new function, that's why it's not found in the tests. Where does the name come from? Is it a standard name?

Now I have exported the function. As for the name: I saw it was already used in MATLAB and GNU Octave's Control toolbox. Also Python-control lists a function with the same name in a sort of "to-do list" of MATLAB vs. Python-control compatibility.

On the other hand, I am also working on `canon` function to decompose a state space model in its canonical forms (diagonal, controlable and observable) and the diagonal form would eventually overcome the need for the prescale function. 

However, the `prescale` function seems to be called on lots of other functions in the aforementioned control libraries due to the fact that it returns a numerically robust representation of the system. I think here, it could have a similar role perhaps.

baggepinnen:
Alright great! I'm not sure why the doc tests are failing, the output looks correct to me. In either case, the failures are unrelated to this PR

src/matrix_comps.jl
```diff
@@ -583,6 +583,28 @@ function similarity_transform(sys::ST, T) where ST <: AbstractStateSpace
...
+Such that `Ã` is diagonal.
+Returns a new scaled state-space object and the associated transformation
+matrix.
+"""
+function prescale(sys::StateSpace)
+    d, S = eigen(sys.A)
+    A = Diagonal(d)
+    B = S\sys.B
+    C = sys.C*S
+    normalized_sys = iscontinuous(sys) ? ss(A, B, C, sys.D) : ss(A, B, C, sys.D, sys.Ts)
```
mfalt, line 604: 'StateSpace(A,B,C,sys.c,sys.timeevol)' would have been an approach that avoid the conditional here. We should probably also extend it to AbstractStateSpace, but there is a separate issue for that.

olof3:
>Now I have exported the function. As for the name: I saw it was already used in MATLAB and GNU Octave's Control toolbox. Also Python-control lists a function with the same name in a sort of "to-do list" of MATLAB vs. Python-control compatibility.

The function `prescale` implemented in this PR does not seem to do the same thing as the Matlab function with the same name? E.g., prescale on a random system in Matlab does not give me a system with a diagonal A matrix.

I think the issue in #462 was that `similarity_transform` was broken, that would have been good to fix.

> On the other hand, I am also working on canon function to decompose a state space model in its canonical forms (diagonal, controlable and observable) and the diagonal form would eventually overcome the need for the prescale function.

Something like this could occasionally be useful though.

tfoliva:
> The function `prescale` implemented in this PR does not seem to do the same thing as the Matlab function with the same name? E.g., prescale on a random system in Matlab does not give me a system with a diagonal A matrix.

It has been a long time since I don't use MATLAB, so I did not check if it returned a diagonal matrix. The name was taken as inspiration but it could be changed.

The idea behind the `prescale` function here was to take a badly conditioned system and transform it to diagonal canonical form which is more numerically robust. Which it does in fact do.

For the naming part, when the `canon` function is ready it could be called with a 'diagonal' option which would do the same thing as the `prescale` function and thus the latter function could be deleted.

tfoliva:
I will look deeper into what `prescale` actually does in this other packages. Apparently as @olof3 said MATLAB uses a different scaling procedure, so I will try to find which one.

baggepinnen:
Bump, is the current version of `prescale` something we'd like to keep for v1.0? It produces complex-coefficient systems in general, which is probably not what most people want?

PR #465, opened 2021-04-04
Update `prescale` function

tfoliva:
Prescale function now actually balances the system matrices A, B and C instead of performing a similarity transformation into diagonal form.


src/matrix_comps.jl
```diff
@@ -585,24 +585,33 @@ end
...
 matrix.
 """
-function prescale(sys::StateSpace)
-    d, S = eigen(sys.A)
-    A = Diagonal(d)
-    B = S\sys.B
-    C = sys.C*S
-    normalized_sys = iscontinuous(sys) ? ss(A, B, C, sys.D) : ss(A, B, C, sys.D, sys.Ts)
-    return normalized_sys, S
+function prescale(sys::StateSpace) where ST <: AbstractStateSpace
```
tfoliva, line 600:
```suggestion
function prescale(sys::ST) where ST <: AbstractStateSpace
```

src/matrix_comps.jl
```diff
@@ -585,24 +585,31 @@ end
...
+function prescale(sys::ST) where ST <: AbstractStateSpace
+    n, m = size(sys.B)
+    p = size(sys.C, 1)
+
+    # create an augmented system to balance the whole system at once
+    metasys = [sys.A sys.B; sys.C zeros(p, m)]
+    S, P, B = balance(metasys)
+
+    # reconstruct A, B and C by taking slices of the balanced metasystem
+    bal_metasys = P\B*P  # undo permutation
```
olof3, line 608: Would it be possible to use `perm=false` to avoid having to undo the permutation?

src/matrix_comps.jl
```diff
@@ -585,24 +585,31 @@ end
...
+    M = [A B;
+         C 0]
 ```
-Such that `Ã` is diagonal.
-Returns a new scaled state-space object and the associated transformation
-matrix.
+such that their column and row norms are closer.
+This results in a system with overall better numerical conditioning.
+
+Returns a new scaled state-space object.
```
olof3, line 597: Do you have any reference to where this method is discussed?

src/matrix_comps.jl
```diff
@@ -585,24 +585,31 @@ end
...
-    B = S\sys.B
-    C = sys.C*S
-    normalized_sys = iscontinuous(sys) ? ss(A, B, C, sys.D) : ss(A, B, C, sys.D, sys.Ts)
-    return normalized_sys, S
+function prescale(sys::ST) where ST <: AbstractStateSpace
+    n, m = size(sys.B)
+    p = size(sys.C, 1)
+
+    # create an augmented system to balance the whole system at once
+    metasys = [sys.A sys.B; sys.C zeros(p, m)]
```
olof3, line 604: You are using two terms here, both augmented and meta system. I seem to remember that it is instead typically called a packed representation? I think that is the term used in the Essentials of Robust Control by Zhou & Doyle,

I think it should be okay to just call the packed system representation `M` as you do in the tests.

test/test_matrix_comps.jl
```diff
@@ -70,11 +70,13 @@ syst = similarity_transform(sys, Tr)
...
-@test T*nsys.A ≈ sys.A*T
-@test T*nsys.B ≈ sys.B
-@test nsys.C ≈ sys.C*T
+nsys = prescale(sys)
+@test pole(nsys) ≈ pole(sys)
+n, m = size(sys.B)
+p = size(sys.C, 1)
+M = [nsys.A nsys.B; nsys.C zeros(p, m)]
+M2 = [sys.A sys.B; sys.C zeros(p, m)]
+@test cond(M) <= cond(M2)
```
olof3, line 79: I think it would be reasonable with quite a few more tests. For example to see that the system dynamics isn't changed by the prescaling. E.g. by comparing the frequency response of a MIMO systems before and after.

PR #467, opened 2021-04-05
More LQG updates

baggepinnen:
This PR adds tests for LQG and makes sure it conforms to Glad Ljung

### DIFF
--- a/src/model_augmentation.jl
+++ b/src/model_augmentation.jl
@@ -39,9 +39,15 @@ function add_low_frequency_disturbance(sys::AbstractStateSpace{Continuous}, Ai::
     add_disturbance(sys, fill(-ϵ, 1, 1), Cd)
 end
 
-function add_low_frequency_disturbance(sys::AbstractStateSpace{Continuous}; ϵ=0)
-    Cd = sys.B
-    add_disturbance(sys, fill(-ϵ, 1, 1), Cd)
+function add_low_frequency_disturbance(sys::AbstractStateSpace{Continuous}; ϵ=0, measurement=false)
+    nx,nu,ny = sys.nx,sys.nu,sys.ny
+    if measurement
+        Cd = I(nu)
+        add_measurement_disturbance(sys, -ϵ*I(nu), Cd)
+    else
+        Cd = sys.B
+        add_disturbance(sys, -ϵ*I(nu), Cd)
+    end
 end
 
 function add_resonant_disturbance(sys::AbstractStateSpace{Continuous}, ω, ζ, Ai::Integer; measurement=false)
--- a/src/model_augmentation.jl
+++ b/src/model_augmentation.jl
@@ -1,0 +1,73 @@
+"""
+    add_disturbance(sys::AbstractStateSpace{Continuous}, Ad::AbstractMatrix, Cd::AbstractMatrix)
+
+See CCS pp. 144
+
+# Arguments:
+- `sys`: System to augment
+- `Ad`: The dynamics of the disturbance
+- `Cd`: How the disturbance states affect the states of `sys`. This matrix as the shape (sys.nx, size(Ad, 1))
+
+See also `add_low_frequency_disturbance, add_resonant_disturbance`
+"""
+function add_disturbance(sys::AbstractStateSpace{Continuous}, Ad::AbstractMatrix, Cd::AbstractMatrix)
+    A,B,C,D = ControlSystems.ssdata(sys)
+    T = eltype(A)
+    nx,nu,ny = sys.nx,sys.nu,sys.ny
+    Ae = [A Cd; zeros(T, size(Ad, 1), nx) Ad]
+    Be = [B; zeros(T, size(Ad, 1), nu)]
+    Ce = [C zeros(T, ny, size(Ad, 1))]
+    De = D
+    ss(Ae,Be,Ce,De)
+end
+
+function add_measurement_disturbance(sys::AbstractStateSpace{Continuous}, Ad::AbstractMatrix, Cd::AbstractMatrix)
+    A,B,C,D = ControlSystems.ssdata(sys)
+    T = eltype(A)
+    nx,nu,ny = sys.nx,sys.nu,sys.ny
+    Ae = [A zeros(T, nx, size(Ad, 1)); zeros(T, size(Ad, 1), nx) Ad]
+    Be = [B; zeros(T, size(Ad, 1), nu)]
+    Ce = [C Cd]
+    De = D
+    ss(Ae,Be,Ce,De)
+end
+
+function add_low_frequency_disturbance(sys::AbstractStateSpace{Continuous}, Ai::Integer; ϵ=0)
+    nx,nu,ny = sys.nx,sys.nu,sys.ny
+    Cd = zeros(nx, 1)
+    Cd[Ai] = 1
+    add_disturbance(sys, fill(-ϵ, 1, 1), Cd)
+end
+
+function add_low_frequency_disturbance(sys::AbstractStateSpace{Continuous}; ϵ=0)
+    Cd = sys.B
+    add_disturbance(sys, fill(-ϵ, 1, 1), Cd)
+end
+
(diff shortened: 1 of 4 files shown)


src/model_augmentation.jl
```diff
@@ -0,0 +1,64 @@
+"""
+    add_disturbance(sys::AbstractStateSpace{Continuous}, Ad::AbstractMatrix, Cd::AbstractMatrix)
+
+See CCS pp. 144
```
mfalt, line 4: I think the description should be a bit better here, I coupé bor figure out what the function was doing without reading the coffee. Maybe the expression for the new system matrix is enough.

mfalt: I understand that you probably use this to introduce disturbances, but it seems that it could be called something more general, like 'augment_system'.

src/model_augmentation.jl
```diff
@@ -0,0 +1,64 @@
...
+    T = eltype(A)
+    nx,nu,ny = sys.nx,sys.nu,sys.ny
+    Ae = [A Cd; zeros(T, size(Ad, 1), nx) Ad]
+    Be = [B; zeros(T, size(Ad, 1), nu)]
+    Ce = [C zeros(T, ny, size(Ad, 1))]
+    De = D
+    ss(Ae,Be,Ce,De)
+end
+
+function add_measurement_disturbance(sys::AbstractStateSpace{Continuous}, Ad::AbstractMatrix, Cd::AbstractMatrix)
```
mfalt, line 24: We should have a docstring if we export. Same for the rest

src/types/lqg.jl
```diff
@@ -124,143 +132,194 @@ function LQG(
...
-    if s ∈ (:L, :K, :Q1, :Q2, :R1, :R2, :qQ, :qR, :integrator, :P, :M)
+    if s ∈ (:L, :K, :Q1, :Q2, :R1, :R2, :qQ, :qR, :integrator, :P, :M, :N, :syse, :Lr)
         return getfield(G, s)
     end
     s === :A && return G.P.A
     s === :B && return G.P.B
     s === :C && return G.P.C
     s === :D && return G.P.D
     s ∈ (:sys, :P) && return getfield(G, :P)
-    s ∈ (:sysc, :controller) && return getfield(G, :sysc)
```
mfalt, line 186: Also breaking change.

src/types/lqg.jl
```diff
@@ -124,143 +132,194 @@ function LQG(
...
+    end
 
     n = size(A, 1)
     m = size(B, 2)
     p = size(C, 1)
     pm = size(M, 1)
+    pn = size(N, 2)
 
     # Extract interesting values
-    if G.integrator # Augment with disturbance model
```
mfalt, line 205: Why is this removed?

mfalt (changes requested):

src/types/lqg.jl
baggepinnen: Nope, there is another line that returns the controller

src/types/lqg.jl
baggepinnen: This is no longer stored in the struct, instead the extended statespace object `syse` is stored. adding an integrator is just a special case of adding disturbance models, corresponding to a low-frequency input disturbance

src/model_augmentation.jl
baggepinnen: Agree, all augmentation functions require better docstrings

baggepinnen: Btw, your first comment probably got eaten up by autocorrect ;)

mfalt: Wow, that was tough to read, fixed now.

src/model_augmentation.jl
```diff
@@ -0,0 +1,64 @@
+"""
+    add_disturbance(sys::AbstractStateSpace{Continuous}, Ad::AbstractMatrix, Cd::AbstractMatrix)
+
+See CCS pp. 144
+
+# Arguments:
+- `sys`: System to augment
+- `Ad`: The dynamics of the disturbance
+- `Cd`: How the disturbance states affect the states of `sys`. This matrix as the shape (sys.nx, size(Ad, 1))
```
albheim, line 9: This matrix **h**as the shape...

olof3:
We have been lacking a simple way of augmenting systems with disturbance/noise models, so it is good that something is finally happening!

* How is the intensity of the noise/disturbances handled? would perhaps be convenient to specify these just for the disturbances rather than some noise intensity for the whole A matrix.
* I guess for, e.g., a resonant disturbance,  1/(s^2 + 2*zeta*w0 + w0^2)  one would also like to keep track of the `B` matrix of its state-space realization?
* I think it should be possible to add arbitrary dynamics in the form of a linear system as disturbance models. Then I think that we should export some of the methods for creating systems from `DemoSystems`, at least `lag` and `resonant`.  (Looking over their names and arguments, making sure that they are acceptable). E.g., I.e., we don't nee `add_low_frequency_disturbance`
* If possible I think it would be desirable to separate how a disturbances affect the system dynamics from the disturbance model. Perhaps using some kind of partitioned state space. 
* The LQG struct seems a bit monolithic/inflexible. Some sort of problem data struct, containing the problem specification would seem useful though, e.g., in order to evaluate other linear controllers. We should fix the feedback function to accept `UniformScaling`, then all the sensitivity functions would just be one-liners, and perhaps wouldn't be needed to be included in the struct. Or do we feel a need for defining these

baggepinnen:
> * How is the intensity of the noise/disturbances handled? would perhaps be convenient to specify these just for the disturbances rather than some noise intensity for the whole A matrix.

It's not handled yet, maybe allowing the user to pass another parameter that is taking the place of the `1` in the C-matrix?

> * I guess for, e.g., a resonant disturbance,  1/(s^2 + 2_zeta_w0 + w0^2)  one would also like to keep track of the `B` matrix of its state-space realization?

Yes, if it's a known disturbance. The augmentation so far only handles unknown disturbances for which the B-matrix is only extended with zeros. 

> * I think it should be possible to add arbitrary dynamics in the form of a linear system as disturbance models. Then I think that we should export some of the methods for creating systems from `DemoSystems`, at least `lag` and `resonant`.  (Looking over their names and arguments, making sure that they are acceptable). E.g., I.e., we no need for `add_low_frequency_disturbance`, etc.

The `add_low_frequency_disturbance` really aims to achieve integral action in the LQG controller by modeling a constant input disturbance. It is a good idea to have a way to create typical disturbance models and add those, e.g., `add_input_disturbance(sys, step_model())` or something like that.

> * If possible I think it would be desirable to separate how a disturbances affect the system dynamics from various disturbance models that one would like to investigate. Perhaps using some kind of partitioned state space.

I've been thinking a bit about this and I don't think we need to implement a new system type. It could probably be solved by making the matrices have named components, e.g., through ComponentArrays.jl. I've used this strategy elsewhere and it works quite nicely. The `HeteroStateSpace` already accepts this kind of matrices.

> * The LQG struct seems a bit monolithic/inflexible. Some sort of problem data struct, containing the problem specification would seem useful though, e.g., in order to evaluate other linear controllers. We should fix the feedback function to accept `UniformScaling`, then all the sensitivity functions would just be one-liners, and perhaps wouldn't be needed to be included as `getproperty`. Or do we feel a need for defining these  generally?

I agree, I don't like the design all that much. Maybe it would be better to just break it completely and redesign it so that the structure only contains computed results that are generally useful, and all other results (that are now properties) are just implemented as regular functions instead. The purpose of defining functions is that it might be hard to remember the exact formulations for input/output sensitivity for MIMO systems if you don't have a reference book or pen and paper nearby. Having them as functions prevent you from making simple errors. Getting exactly the same realisation out as matlab also required some gymnastics, at least for `output_sensitivity`
 
> Perhaps you have some examples/use cases? Would be easier to discuss around something a bit more concrete. If I have the time, I will try to put something together as well. And I guess there should be plenty of examples in the literature, any suggestions?

Double-mass model (4 states) augmented with constant input disturbance (1 state) and resonant measurement disturbance (2 states) to reduce sensitivity at select frequencies. Motor-side states are measured, while references are provided for load-side states (C and M matrices are different) while noise acts on the two velocity states as well as the input-disturbance state. I'll put together some code soon.

PR #468, opened 2021-04-25
Lsim gpu

albheim:
Changed some lsim and ltitr code so it will support CuArrays or HeteroStateSpace with CuArrays.

Ran this test on master and new branch to make sure it kept same performance on normal arrays.
```julia
using ControlSystems, BenchmarkTools

ny, nx, nu = 10, 100, 20
dt = 0.01
A = randn(nx,nx)
B = randn(nx,nu)
C = randn(ny,nx)
D = zeros(ny,nu)
sys = ss(A, B, C, D)
sysd = c2d(sys, dt)

t = 0:dt:1
x0 = zeros(nx)

tmp = ones(nu)
u(x, t) = tmp
uu = transpose(reduce(hcat, u(x0, tt) for tt in t))

@benchmark lsim(sys, u, t; x0=x0)
@benchmark lsim(sys, uu, t; x0=x0)
@benchmark lsim(sysd, u, t; x0=x0)
@benchmark lsim(sysd, uu, t; x0=x0)
```

Results on master
```julia
julia> @benchmark lsim(sys, u, t; x0=x0)
BenchmarkTools.Trial: 
  memory estimate:  881.52 KiB
  allocs estimate:  3159
  --------------
  minimum time:     1.429 ms (0.00% GC)
  median time:      1.664 ms (0.00% GC)
  mean time:        1.935 ms (7.80% GC)
  maximum time:     22.504 ms (90.67% GC)
  --------------
  samples:          2572
  evals/sample:     1

julia> @benchmark lsim(sys, uu, t; x0=x0)
BenchmarkTools.Trial: 
  memory estimate:  3.84 MiB
  allocs estimate:  106
  --------------
  minimum time:     2.052 ms (0.00% GC)
  median time:      2.321 ms (0.00% GC)
  mean time:        2.529 ms (5.97% GC)
  maximum time:     14.941 ms (75.71% GC)
  --------------
  samples:          1971
  evals/sample:     1

julia> @benchmark lsim(sysd, u, t; x0=x0)
BenchmarkTools.Trial: 
  memory estimate:  421.77 KiB
  allocs estimate:  823
  --------------
  minimum time:     824.109 μs (0.00% GC)
  median time:      916.514 μs (0.00% GC)
  mean time:        970.925 μs (3.00% GC)
  maximum time:     15.533 ms (92.07% GC)
  --------------
  samples:          5125
  evals/sample:     1

julia> @benchmark lsim(sysd, uu, t; x0=x0)
BenchmarkTools.Trial: 
  memory estimate:  105.56 KiB
  allocs estimate:  20
  --------------
  minimum time:     714.580 μs (0.00% GC)
  median time:      788.074 μs (0.00% GC)
  mean time:        813.631 μs (0.55% GC)
  maximum time:     4.152 ms (79.61% GC)
  --------------
  samples:          6111
  evals/sample:     1
```
Results on new branch
```julia
julia> @benchmark lsim(sys, u, t; x0=x0)
BenchmarkTools.Trial: 
  memory estimate:  1.61 MiB
  allocs estimate:  2815
  --------------
  minimum time:     1.451 ms (0.00% GC)
  median time:      1.544 ms (0.00% GC)
  mean time:        1.697 ms (5.76% GC)
  maximum time:     18.358 ms (87.38% GC)
  --------------
  samples:          2930
  evals/sample:     1

julia> @benchmark lsim(sys, uu, t; x0=x0)
BenchmarkTools.Trial: 
  memory estimate:  4.10 MiB
  allocs estimate:  302
  --------------
  minimum time:     2.080 ms (0.00% GC)
  median time:      2.182 ms (0.00% GC)
  mean time:        2.456 ms (6.30% GC)
  maximum time:     8.137 ms (67.88% GC)
  --------------
  samples:          2029
  evals/sample:     1

julia> @benchmark lsim(sysd, u, t; x0=x0)
BenchmarkTools.Trial: 
  memory estimate:  421.81 KiB
  allocs estimate:  827
  --------------
  minimum time:     795.230 μs (0.00% GC)
  median time:      909.728 μs (0.00% GC)
  mean time:        994.502 μs (4.15% GC)
  maximum time:     28.030 ms (95.81% GC)
  --------------
  samples:          4990
  evals/sample:     1

julia> @benchmark lsim(sysd, uu, t; x0=x0)
BenchmarkTools.Trial: 
  memory estimate:  373.31 KiB
  allocs estimate:  216
  --------------
  minimum time:     688.868 μs (0.00% GC)
  median time:      764.116 μs (0.00% GC)
  mean time:        791.641 μs (2.06% GC)
  maximum time:     11.026 ms (90.31% GC)
  --------------
  samples:          6278
  evals/sample:     1
```

This code was used to compare normal system and ones using CuArrays
```julia
using ControlSystems, BenchmarkTools, CUDA

ny, nx, nu = 100, 1000, 200
dt = 0.01
T = Float32
A  = randn(T, nx,nx)
B  = randn(T, nx,nu)
C = randn(T, ny,nx)
D = zeros(T, ny,nu)
sys = ss(A, B, C, D)
sysd = c2d(sys, dt)
sysg = HeteroStateSpace(cu(A), cu(B), cu(C), cu(D))
sysdg = HeteroStateSpace(cu(sysd.A), cu(sysd.B), cu(sysd.C), cu(sysd.D), dt)

tmp = ones(nu)
u(x, t) = tmp
tmpg = cu(ones(nu))
ug(x, t) = tmpg

x0 = zeros(nx)
x0g = cu(zeros(nx))

t = 0:dt:1

uu = transpose(reduce(hcat, u(x0, tt) for tt in t))
uug = transpose(reduce(hcat, ug(x0g, tg) for tg in t))

@benchmark lsim(sys, u, t; x0=x0)
@benchmark lsim(sysg, ug, t; x0=x0g)

@benchmark lsim(sysd, u, t; x0=x0)
@benchmark lsim(sysdg, ug, t; x0=x0g)

@benchmark lsim(sysd, uu, t; x0=x0)
@benchmark lsim(sysdg, uug, t; x0=x0g)
```
which gave this result
```julia
julia> @benchmark lsim(sys, u, t; x0=x0)
BenchmarkTools.Trial: 
  memory estimate:  18.91 MiB
  allocs estimate:  3872
  --------------
  minimum time:     81.210 ms (0.00% GC)
  median time:      96.187 ms (0.00% GC)
  mean time:        97.154 ms (1.63% GC)
  maximum time:     112.403 ms (13.90% GC)
  --------------
  samples:          52
  evals/sample:     1

julia> @benchmark lsim(sysg, ug, t; x0=x0g)
BenchmarkTools.Trial: 
  memory estimate:  3.93 MiB
  allocs estimate:  82690
  --------------
  minimum time:     29.376 ms (0.00% GC)
  median time:      30.166 ms (0.00% GC)
  mean time:        35.255 ms (2.59% GC)
  maximum time:     150.480 ms (16.68% GC)
  --------------
  samples:          142
  evals/sample:     1

julia> @benchmark lsim(sysd, u, t; x0=x0)
BenchmarkTools.Trial: 
  memory estimate:  3.70 MiB
  allocs estimate:  831
  --------------
  minimum time:     14.601 ms (0.00% GC)
  median time:      16.871 ms (0.00% GC)
  mean time:        17.373 ms (2.32% GC)
  maximum time:     43.499 ms (57.95% GC)
  --------------
  samples:          288
  evals/sample:     1

julia> @benchmark lsim(sysdg, ug, t; x0=x0g)
BenchmarkTools.Trial: 
  memory estimate:  481.30 KiB
  allocs estimate:  20107
  --------------
  minimum time:     5.606 ms (0.00% GC)
  median time:      17.440 ms (0.00% GC)
  mean time:        18.180 ms (1.11% GC)
  maximum time:     293.485 ms (18.85% GC)
  --------------
  samples:          275
  evals/sample:     1

julia> @benchmark lsim(sysd, uu, t; x0=x0)
BenchmarkTools.Trial: 
  memory estimate:  3.47 MiB
  allocs estimate:  220
  --------------
  minimum time:     10.714 ms (0.00% GC)
  median time:      12.526 ms (0.00% GC)
  mean time:        13.298 ms (2.40% GC)
  maximum time:     37.849 ms (64.22% GC)
  --------------
  samples:          376
  evals/sample:     1

julia> @benchmark lsim(sysdg, uug, t; x0=x0g)
BenchmarkTools.Trial: 
  memory estimate:  223.44 KiB
  allocs estimate:  9603
  --------------
  minimum time:     2.641 ms (0.00% GC)
  median time:      2.774 ms (0.00% GC)
  mean time:        3.348 ms (3.66% GC)
  maximum time:     301.640 ms (25.66% GC)
  --------------
  samples:          1490
  evals/sample:     1
```

### DIFF
--- a/src/timeresp.jl
+++ b/src/timeresp.jl
@@ -160,7 +160,7 @@ end
 @deprecate lsim(sys, u, t, x0, method) lsim(sys, u, t; x0=x0, method=method)
 
 function lsim(sys::AbstractStateSpace, u::Function, Tf::Real, args...; kwargs...)
-    t = default_time_vector(sys, Tf)
+    t = _default_time_vector(sys, Tf)
     lsim(sys, u, t, args...; kwargs...)
 end
 
--- a/src/timeresp.jl
+++ b/src/timeresp.jl
@@ -216,7 +216,7 @@ e.g, `x0` should prefereably not be a sparse vector.
 If `u` is a function, then `u(x,i)` is called to calculate the control signal every iteration. This can be used to provide a control law such as state feedback `u=-Lx` calculated by `lqr`. In this case, an integrer `iters` must be provided that indicates the number of iterations.
 """
 function ltitr(A::AbstractArray, B::AbstractArray, u::AbstractVecOrMat,
-        x0::AbstractVecOrMat=zeros(eltype(A), size(A, 1)))
+        x0::AbstractVecOrMat=fill!(similar(A, size(A, 1)), 0))
 
     T = promote_type(LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(A), eltype(x0)),
                       LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(B), eltype(u)))
--- a/src/timeresp.jl
+++ b/src/timeresp.jl
@@ -110,7 +110,7 @@ plot(t,x, lab=["Position" "Velocity"], xlabel="Time [s]")
 ```
 """
 function lsim(sys::AbstractStateSpace, u::AbstractVecOrMat, t::AbstractVector;
-        x0::AbstractVector=zeros(Bool, sys.nx), method::Symbol=:unspecified)
+        x0::AbstractVecOrMat=zeros(Bool, sys.nx), method::Symbol=:unspecified)
     ny, nu = size(sys)
     nx = sys.nx
 
@@ -160,12 +160,12 @@ end
 @deprecate lsim(sys, u, t, x0, method) lsim(sys, u, t; x0=x0, method=method)
 
 function lsim(sys::AbstractStateSpace, u::Function, Tf::Real, args...; kwargs...)
-    t = _default_time_vector(sys, Tf)
+    t = default_time_vector(sys, Tf)
     lsim(sys, u, t, args...; kwargs...)
 end
 
 function lsim(sys::AbstractStateSpace, u::Function, t::AbstractVector;
-        x0::VecOrMat=zeros(sys.nx), method::Symbol=:cont)
+        x0::AbstractVecOrMat=zeros(sys.nx), method::Symbol=:cont)
     ny, nu = size(sys)
     nx = sys.nx
     u0 = u(x0,1)
@@ -179,23 +179,20 @@ function lsim(sys::AbstractStateSpace, u::Function, t::AbstractVector;
     dt = T(t[2] - t[1])
 
     if !iscontinuous(sys) || method === :zoh
+
         if iscontinuous(sys)
             dsys = c2d(sys, dt, :zoh)
         else
-            if sys.Ts != dt
-                error("Time vector must match sample time for discrete system")
-            end
+            @assert sys.Ts == dt "Time vector must match sample time for discrete system, $(dt) ≠ $(sys.Ts)"
             dsys = sys
         end
         x,uout = ltitr(dsys.A, dsys.B, u, t, T.(x0))
     else
-        s = Simulator(sys, u)
-        sol = solve(s, T.(x0), (t[1],t[end]), Tsit5())
-        x = sol(t)'
-        uout = Array{eltype(x)}(undef, length(t), ninputs(sys))
-        for (i,ti) in enumerate(t)
-            uout[i,:] = u(x[i,:],ti)'
-        end
+        f(dx,x,p,t) = dx .= sys.A*x .+ sys.B*u(x,t)
+        sol = solve(ODEProblem(f,x0,(t[1],t[end])), Tsit5())
+        x = sol(t).u
+        uout = transpose(reduce(hcat, u(x[i], t[i]) for i in eachindex(x)))
+        x = transpose(reduce(hcat, x))
     end
     y = transpose(sys.C*transpose(x) + sys.D*transpose(uout))
     return y, t, x, uout
@@ -217,16 +214,13 @@ e.g, `x0` should prefereably not be a sparse vector.
 
 If `u` is a function, then `u(x,i)` is called to calculate the control signal every iteration. This can be used to provide a control law such as state feedback `u=-Lx` calculated by `lqr`. In this case, an integrer `iters` must be provided that indicates the number of iterations.
 """
-@views function ltitr(A::AbstractMatrix, B::AbstractMatrix, u::AbstractVecOrMat,
-        x0::AbstractVector=zeros(eltype(A), size(A, 1)))
+function ltitr(A::AbstractArray, B::AbstractArray, u::AbstractVecOrMat,
+        x0::AbstractVecOrMat=zeros(eltype(A), size(A, 1)))
 
     T = promote_type(LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(A), eltype(x0)),
                       LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(B), eltype(u)))
 
     n = size(u, 1)
-
-    # Transposing u allows column-wise operations, which apparently is faster.
-    ut = transpose(u)
 
     # Using similar instead of Matrix{T} to allow for CuArrays to be used.
     # This approach is problematic if x0 is sparse for example, but was considered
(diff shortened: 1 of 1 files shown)


albheim:
One other thing is that if you are using GPU you currently have to send `x0` in since it wont be created correctly. I tried doing something like `...x0::AbstractVecOrMat=fill!(similar(A, size(A, 1)), 0)` instead of the `zeros(eltype(A), size(A, 1))` but it seemed to be slightly slower for the normal case so maybe it is better to leave it like it is and just have those who want to use GPU send in `x0`. This will however make it so that `step` for example will not work with GPU since x0 is created as `zeros(nx)` there.

src/timeresp.jl
```diff
@@ -189,13 +189,11 @@ function lsim(sys::AbstractStateSpace, u::Function, t::AbstractVector;
...
-        s = Simulator(sys, u)
-        sol = solve(s, T.(x0), (t[1],t[end]), Tsit5())
-        x = sol(t)'
-        uout = Array{eltype(x)}(undef, length(t), ninputs(sys))
-        for (i,ti) in enumerate(t)
-            uout[i,:] = u(x[i,:],ti)'
-        end
+        f(dx,x,p,t) = dx .= sys.A*x .+ sys.B*u(x,t)
+        sol = solve(ODEProblem(f,x0,(t[1],t[end])), Tsit5())
+        x = sol(t).u
```
baggepinnen, line 194: There is an option to supply `saveat=t` to `solve`, in which case `sol.u` corresponds exactly to `sol(t).u` without going through the interpolator. I'm not sure if there is a speed or allocation gain to be had from doing so?

src/timeresp.jl
```diff
@@ -217,35 +215,30 @@ e.g, `x0` should prefereably not be a sparse vector.
...
-    # Transposing u allows column-wise operations, which apparently is faster.
-    ut = transpose(u)
-
     # Using similar instead of Matrix{T} to allow for CuArrays to be used.
     # This approach is problematic if x0 is sparse for example, but was considered
     # to be good enough for now
     x = similar(x0, T, (length(x0), n))
 
     x[:,1] .= x0
-    mul!(x[:, 2:end], B, transpose(u[1:end-1, :])) # Do all multiplications B*u[:,k] to save view allocations
```
baggepinnen, line 237: Was there a problem with `mul!` on the GPU? Otherwise the `mul!` with views should avoid the allocation.

src/timeresp.jl
albheim: Nice, will have a look

src/timeresp.jl
albheim: There was something with the views that made me remove it. Will have a look at what was going on.

mfalt: I think your benchmark might be biased since you input u as a transpose btw

baggepinnen: Worse is that the array is not strided when using views due to the removal of the last row.
Transposes are not always less efficient, `A'B` is usually more efficient than `AB`.

olof3: I think that we've already had some offline discussions regarding the intput/output format of `lsim`, but I think that we decided to push the decision to the future. I opened a new issue instead of hijacking this thread, see #469.

src/timeresp.jl
albheim: Seems like for large systems it was barely noticeable (majority of time is spent in solve, and that was still similar), but for smaller systems it was actually making a difference so I will put it in.

src/timeresp.jl
albheim: @mfalt Yeah, makes sense. Totally missed that.

Trying some different versions of this specific line.

First calling lsim with discrete sys and u vec where (ny, nx, nu) = (100, 500, 200) and u vec is not a transpose anymore.
* `@views mul!(x[:, 2:end], B, transpose(u[1:end-1, :]))` error on GPU, 4.2ms median on CPU
* `@views mul!(x[:, 2:end], B, transpose(u)[:, 1:end-1])` 3.2s medain on GPU, 13.7ms median on CPU
* `x[:, 2:end] .= B * transpose(u[1:end-1, :])` 2.7ms median on GPU, 4.8ms median on CPU
* `x[:, 2:end] .= B * transpose(u)[:, 1:end-1]` 287ms median on GPU, 4.8ms median on CPU

The calling lsim with (nx, ny, nu) = (1, 5, 2) (only for CPU here).
* `@views mul!(x[:, 2:end], B, transpose(u[1:end-1, :]))` 31micros median on CPU
* `@views mul!(x[:, 2:end], B, transpose(u)[:, 1:end-1])` 35micros median on CPU
* `x[:, 2:end] .= B * transpose(u[1:end-1, :])` 24micros median on CPU
* `x[:, 2:end] .= B * transpose(u)[:, 1:end-1]` 24micros median on CPU

So the original with views and mul! does not work with GPU for some reason, I have not dug deeper into why. Of the others it seems like the GPU does not like to transpose a sliced matrix (100x slower) so that is why I switched it around. And this held even when not sending in the u vec as a transpose.

baggepinnen: I think the problem here is the data layout as discussed in https://github.com/JuliaControl/ControlSystems.jl/issues/469
 We need to chop `u` in a very unfortunate way, either to make a new copy and pay the allocation as in the last two examples, or use a view into a non-strided array, probably causing a generic fallback method to be called rather than an optimized BLAS/CUBLAS method. had the memory layout been transposed, i.e., `u in R(nu x T)`, we could use `@view` and get a strided array hitting a BLAS method and both GPU and CPU versions would be blazing fast without allocations.

baggepinnen: Example
```julia
@btime @views mul!($(x)[:, 2:end], $(B), $(u)[:, 2:end])   # 3.873 μs (0 allocations: 0 bytes)
@btime @views mul!($(x)[:, 2:end], $(B), $(u')[2:end, :]') # 29.313 μs (6 allocations: 336 bytes)
```

albheim: Yeah, I don't know what the best solution would be. 

If #469 will happen then I think that would solve the problem. 

I also asked in the CUDA.jl community and was told that the array type was too complex for them to do any specialized dispatch (mul! of a view of a transpose of a slice) but was told that this PR https://github.com/JuliaGPU/GPUArrays.jl/pull/352 would allow us to get it working but not optimized for GPU using the current CPU-optimized code.

albheim:
Closing this, most of this was updated in #480

PR #480, opened 2021-05-07
Switch u layout

albheim:
Switch the layout for u in the toolbox.

### DIFF
--- a/src/timeresp.jl
+++ b/src/timeresp.jl
@@ -112,16 +112,14 @@ plot(t,x', lab=["Position" "Velocity"], xlabel="Time [s]")
 ```
 """
 function lsim(sys::AbstractStateSpace, u::AbstractVecOrMat, t::AbstractVector;
-        x0::AbstractVector=zeros(Bool, sys.nx), method::Symbol=:unspecified)
+        x0::AbstractVecOrMat=zeros(Bool, sys.nx), method::Symbol=:unspecified)
     ny, nu = size(sys)
     nx = sys.nx
 
     if length(x0) != nx
         error("size(x0) must match the number of states of sys")
     end
-    if size(u) == (length(t),)
-        reshape!(u, :, 1) # Do we want to modify arguments? Or should this be copied?
-    else size(u) != (nu, length(t))
+    if size(u) != (nu, length(t))
         error("u must be of size (nu, length(t))")
     end
 
@@ -156,7 +154,7 @@ function lsim(sys::AbstractStateSpace, u::AbstractVecOrMat, t::AbstractVector;
 end
 
 function lsim(sys::AbstractStateSpace{<:Discrete}, u::AbstractVecOrMat; kwargs...)
-    t = range(0, length=size(u, 1), step=sys.Ts)
+    t = range(0, length=size(u, 2), step=sys.Ts)
     lsim(sys, u, t; kwargs...)
 end
 
@@ -169,7 +167,7 @@ function lsim(sys::AbstractStateSpace, u::Function, tfinal::Real, args...; kwarg
 end
 
 function lsim(sys::AbstractStateSpace, u::Function, t::AbstractVector;
-        x0::VecOrMat=zeros(sys.nx, 1), method::Symbol=:cont)
+        x0::AbstractVecOrMat=zeros(sys.nx, 1), method::Symbol=:cont)
     ny, nu = size(sys)
     nx = sys.nx
     u0 = u(x0,1)
@@ -219,7 +217,7 @@ e.g, `x0` should prefereably not be a sparse vector.
 If `u` is a function, then `u(x,i)` is called to calculate the control signal every iteration. This can be used to provide a control law such as state feedback `u=-Lx` calculated by `lqr`. In this case, an integrer `iters` must be provided that indicates the number of iterations.
 """
 @views function ltitr(A::AbstractMatrix, B::AbstractMatrix, u::AbstractVecOrMat,
-        x0::AbstractVector=zeros(eltype(A), size(A, 1)))
+        x0::AbstractVecOrMat=zeros(eltype(A), size(A, 1)))
 
     T = promote_type(LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(A), eltype(x0)),
                       LinearAlgebra.promote_op(LinearAlgebra.matprod, eltype(B), eltype(u)))
@@ -241,7 +239,7 @@ If `u` is a function, then `u(x,i)` is called to calculate the control signal ev
 end
 
 function ltitr(A::AbstractMatrix{T}, B::AbstractMatrix{T}, u::Function, t,
-    x0::VecOrMat=zeros(T, size(A, 1))) where T
+    x0::AbstractVecOrMat=zeros(T, size(A, 1))) where T
     iters = length(t)
     x = similar(A, size(A, 1), iters)
     uout = similar(A, size(B, 2), iters)
--- a/src/timeresp.jl
+++ b/src/timeresp.jl
@@ -10,18 +10,20 @@ Calculate the step response of system `sys`. If the final time `tfinal` or time
 vector `t` is not provided, one is calculated based on the system pole
 locations.
 
-`y` has size `(length(t), ny, nu)`, `x` has size `(length(t), nx, nu)`"""
+`y` has size `(ny, length(t), nu)`, `x` has size `(nx, length(t), nu)`"""
 function Base.step(sys::AbstractStateSpace, t::AbstractVector; method=:cont)
     lt = length(t)
     ny, nu = size(sys)
     nx = sys.nx
-    u = (x,t)->[one(eltype(t))]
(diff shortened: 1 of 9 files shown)


src/timeresp.jl
```diff
@@ -117,8 +117,8 @@ function lsim(sys::AbstractStateSpace, u::AbstractVecOrMat, t::AbstractVector;
     if length(x0) != nx
         error("size(x0) must match the number of states of sys")
     end
-    if !(size(u) in [(length(t), nu) (length(t),)])
-        error("u must be of size (length(t), nu)")
+    if !(size(u) in [(nu, length(t)) (length(t),)]) # TODO check this and reshape automatically?
```
baggepinnen, line 120:
```suggestion
    if !(size(u) in ((nu, length(t)), (length(t),))) # TODO check this and reshape automatically?
```
No need to allocate an array ;)

src/timeresp.jl
```diff
@@ -189,15 +189,12 @@ function lsim(sys::AbstractStateSpace, u::Function, t::AbstractVector;
...
         x,uout = ltitr(dsys.A, dsys.B, u, t, T.(x0))
     else
-        s = Simulator(sys, u)
-        sol = solve(s, T.(x0), (t[1],t[end]), Tsit5())
-        x = sol(t)'
-        uout = Array{eltype(x)}(undef, length(t), ninputs(sys))
-        for (i,ti) in enumerate(t)
-            uout[i,:] = u(x[i,:],ti)'
-        end
+        f(dx,x,p,t) = dx .= sys.A*x .+ sys.B*u(x,t)
```
baggepinnen, line 192: Here we have an opportunity to do better in case arrays are static and we take up the direct dependence on StaticArrays. The `dx .=` will fail for static arrays.

baggepinnen (commented):
Looks good. Would be nice to also have a method that accepts preallocated `x`, but can be a separate PR. No tests need tweaking to accomodate the changes?

albheim:
> Looks good. Would be nice to also have a method that accepts preallocated `x`, but can be a separate PR. No tests need tweaking to accomodate the changes?

Yeah, I think there might be some test (along with maybe other dependent methods) to fix before this is ready. Just started it and used it to check if I could push. Will fix the rest when I have time, but that will likely be after the weekend. But you are very welcome to have a look at it if you want.

src/timeresp.jl
```diff
@@ -117,8 +119,10 @@ function lsim(sys::AbstractStateSpace, u::AbstractVecOrMat, t::AbstractVector;
     if length(x0) != nx
         error("size(x0) must match the number of states of sys")
     end
-    if !(size(u) in [(length(t), nu) (length(t),)])
-        error("u must be of size (length(t), nu)")
+    if size(u) == (length(t),)
+        reshape!(u, :, 1) # Do we want to modify arguments? Or should this be copied?
+    else size(u) != (nu, length(t))
+        error("u must be of size (nu, length(t))")
```
albheim, line 125: Not sure if this is the way to handle it? I think I changed all the tests that sent in single dim `u` to send in 1xn dims instead, so maybe add test for this also if we use this.

Is reshaping `u` okay here, or should we copy?

albheim:
Had some time this morning so tried to go through a bit more. One error outside of test when running tests, and haven't had time to find it yet. Need to head out now so thought I would push what I had.

src/timeresp.jl
albheim: I haven't used static arrays at all, so not familiar with why that would fail. What would be the static array friendly option?

src/timeresp.jl
albheim: Okay, removed this for now. So now it only accepts `u` that is shaped `(nu, length(t))` since that was easier to handle in some other cases and I think having special cases usually only becomes confusing.

src/timeresp.jl
```diff
@@ -189,15 +191,16 @@ function lsim(sys::AbstractStateSpace, u::Function, t::AbstractVector;
...
-        x = sol(t)'
-        uout = Array{eltype(x)}(undef, length(t), ninputs(sys))
-        for (i,ti) in enumerate(t)
-            uout[i,:] = u(x[i,:],ti)'
+        function f(dx,x,p,t) 
+            A, B, u = p
+            dx .= A*x .+ B*u(x,t)
         end
+        p = (sys.A, sys.B, u)
+        sol = solve(ODEProblem(f,x0,(t[1],t[end]),p), Tsit5(); saveat=t)
```
mfalt, line 199:
```suggestion
        sol = solve(ODEProblem(f,x0,(t[1],t[end]),p), Tsit5(); saveat=t, kwargs...)
```

src/timeresp.jl
```diff
@@ -189,15 +191,16 @@ function lsim(sys::AbstractStateSpace, u::Function, t::AbstractVector;
         end
         x,uout = ltitr(dsys.A, dsys.B, u, t, T.(x0))
     else
-        s = Simulator(sys, u)
-        sol = solve(s, T.(x0), (t[1],t[end]), Tsit5())
-        x = sol(t)'
-        uout = Array{eltype(x)}(undef, length(t), ninputs(sys))
-        for (i,ti) in enumerate(t)
-            uout[i,:] = u(x[i,:],ti)'
+        function f(dx,x,p,t)
```
mfalt, line 194: There is no need to have this be an internal function in lsim. That will probably force DifferentialEquations to have to recompile every time since `f` is a new function every time lsim is called.

mfalt:
Tests look good, feel free to merge.

src/timeresp.jl
```diff
@@ -10,18 +10,20 @@ Calculate the step response of system `sys`. If the final time `tfinal` or time
...
+    #tmp = [one(eltype(t))] # Slightly less allocations but maybe no real difference, 539.624 μs (4916 allocations: 721.61 KiB)
+    #u = (x,t)->tmp
     x0 = zeros(nx)
     if nu == 1
         y, tout, x, _ = lsim(sys, u, t, x0=x0, method=method)
     else
-        x = Array{Float64}(undef, lt, nx, nu)
-        y = Array{Float64}(undef, lt, ny, nu)
+        x = Array{Float64}(undef, nx, lt, nu)
+        y = Array{Float64}(undef, ny, lt, nu)
```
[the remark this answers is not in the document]
albheim: I thought this is what we said for the first two arguments, and the last one made sense to me to leave based on how it was accessed. What would the other alternative be, nx x nu x lt?

Currently freqresp has the frequency index in the first dimension if I'm not mistaken, why would it be the most natural to have it last?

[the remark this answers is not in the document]
baggepinnen: Also, `Float64` should probably be made generic?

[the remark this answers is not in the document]
baggepinnen: `freqresp` probably needs a change corresponding to the one in this PR to place frequency in the last dimension, since a single call to `evalfr` generates a matrix `ny x nu` of frequency responses at a single frequency.

src/timeresp.jl
baggepinnen: A static array is static, hence trying to mutate it with `dx .= something` is not allowed. A static-arrays friendly version would not use any mutation and just write the code as if it allocated, because static arrays are stack-allocated and hence the allocation is free. This code is still correct from a regular array perspective, so any optimizations for static arrays would be in form of another method, i.e.,
```julia
f(x::SArray,p,t) = sys.A*x .+ sys.B*u(x,t)
```

src/timeresp.jl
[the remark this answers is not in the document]
baggepinnen: One can always use `PermutedDimsArray` in the end to shape the array the way we want without copying. The user can then, if they choose, call `copy` on it to obtain a regular array, or use the `PermutedDimsArray` like it was a regular array
```julia
julia> PermutedDimsArray(randn(1,2,3), (2,3,1))
2×3×1 PermutedDimsArray(::Array{Float64, 3}, (2, 3, 1)) with eltype Float64:
```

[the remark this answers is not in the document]
olof3: `PermutedDimsArray` seems  like a reasonable fix.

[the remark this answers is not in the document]
olof3: Would probably be better to avoid introducing `lt`, consider just using `length(t)` or possibly call it `nt` instead.

src/timeresp.jl
albheim: Ahh, thought you only meant that A and B should be static. But everything is going to be static. Sounds interesting, will have to have a look at that.

Should it then be up to the user to define x0 and then the simulation is done with that array type for x?

src/timeresp.jl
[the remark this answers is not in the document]
albheim: > Also, `Float64` should probably be made generic?

Seem reasonable, copied the way used in impulse.

> Would probably be better to avoid introducing `lt`, consider just using `length(t)` or possibly call it `nt` instead.

I switched to using `length(t)` instead.

PR #482, opened 2021-05-10
Better handling of problematic cases for delay systems

mfalt:


baggepinnen:
It's in general nice if there's a way to get rid of warnings to not have 1000s of messages if your calling in a loop. I usually provide
```julia
@warn "message" maxlog=10
```
or something like that to print the message at most 10 times.

PR #483, opened 2021-05-12
Gangof4 and Nyquistplot fixes

olof3:
Gang of four:
* Changed nonstandard naming of transfer functions.
* Avoid overly nonminal realizations when forming gang of four.

Nyquist plot:
* Added critical point
* Ability to set multiple Ms circles at arbitrary values (not just Ms=1)
* Renamed the Mt circle to unit_circle, it does not correspond to Mt = 1.

### DIFF
--- a/src/analysis.jl
+++ b/src/analysis.jl
@@ -462,7 +462,7 @@ function delaymargin(G::LTISystem)
 end
 
 function robust_minreal(G, args...; kwargs...)
-    try 
+    try
         return minreal(G, args...; kwargs...)
     catch
         return G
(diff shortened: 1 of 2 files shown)


olof3:
I would be happy to get rid of `nyquist`, it is only used in nyquist plot, and it would be just as easy, or perhaps even easier to just use `freqresp`. 

We might also consider removing the unit circle from the Nyquist plot.

olof3:
Removed the Ms_circles by default, but we could have one with `Ms=1.5` or `Ms=1.4` or whatever.

For some reason the axis equal and the ranges for the Nyquist plot does not seem to work without the circles. Anyone has an idea how to fix it?

baggepinnen:
There is a PR https://github.com/JuliaControl/ControlSystems.jl/pull/434 ready for merging that implements several similar fixes to nyquistplot

olof3:
I had some recollection of that we discussed something similar, but thought that i disappeared somewhere. Would have been good if thatone was merged. That one looks better in most regards, except the docstring which currently is a bit wrong, also the gaincircles should perhaps be removed / replaced with unit_circle.

Is it convenient to cherry pick the gangof4 commit?

baggepinnen:
> Is it convenient to cherry pick the gangof4 commit?
Cherry picking might be problematic since both PR's change the same exact lines of code. Is probably easier to check out the PR branch and make the changes in a new commit there

baggepinnen:
Ah, the changes to gofplot can probably be cherry-picked without problem

baggepinnen:
>  also the gaincircles should perhaps be removed / replaced with unit_circle

The gaincircles is deactivated by default in that PR

olof3:
Sounds good.

olof3:
> There is a PR #434 ready for merging that implements several similar fixes to nyquistplot

@baggepinnen I merged your PR one into this branch. The plots are actually quite decent now.

The PR to dev (and the one that should be merged) is #486. Made this one by mistake, but kept this one since the discussion took off here.

olof3:
@JuliaControlBot test-plots

baggepinnen:
Closed in favor of #486

PR #485, opened 2021-05-12
Avoid unnecessarily large realization for feedback of TransferFunction

olof3:


src/connections.jl
```diff
@@ -161,6 +161,19 @@ Returns `L/(1+L)` or `P1/(1+P1*P2)`
 feedback(L::TransferFunction) = L/(1+L)
 feedback(P1::TransferFunction, P2::TransferFunction) = P1/(1+P1*P2)
 
+function feedback(G1::TransferFunction{<:TimeEvolution,<:SisoRational}, G2::TransferFunction{<:TimeEvolution,<:SisoRational})
+    if !issiso(G1) || !issiso(G2)
+        error("MIMO TransferFunction feedback isn't implemented.")
+    end
+    G1num = G1.matrix[1,1].num
```
baggepinnen, line 168:
```suggestion
    G1num = numpoly(G1)[]
```
`numpoly/denpoly` should be used instead I guess, in case the internal representation changes

olof3: Good idea, I'll update. Is `only(x)` or `x[]` recommended, I haven't seen the latter one, but I guess they are equivalent?

albheim:
Looks good, this would solve the symbolics problem I assume?

baggepinnen:
Nice, this type of improvements make the transfer-function misery slightly less miserable :)

olof3:
The tests didn't seem to run on any commits except the first one (where it failed)?

> Looks good, this would solve the symbolics problem I assume?

Yes, now it works :)

baggepinnen:
ah, sneaky green checkmark then. If we're going to pursue this dev branch strategy we need to find a way to make CI behave as for the master branch, perhaps we can figure out how others do it.

albheim:
I think we could just add it to the test.yml
```
name: CI 

on:
  pull_request:
    branches:
      - master
      - dev          <--- this
  push:
    branches:
      - master
    tags: 
      - '*'

jobs:
  runtests: 
...
```

PR #486, opened 2021-05-12
Gangof4 fixes

olof3:
Gang of four:
* Changed nonstandard naming of transfer functions.
* Avoid overly nonminal realizations when forming gang of four.

Nyquist plot:
* Added critical point
* Ability to set multiple Ms circles at arbitrary values (not just Ms=1)
* Renamed the Mt circle to unit_circle, it does not correspond to Mt = 1.

### DIFF
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -788,7 +788,7 @@ function gangoffourplot(P::Union{Vector, LTISystem}, C::Vector, args...; minimal
     titles = fill("", 1, plotphase ? 8 : 4)
     # Empty titles on phase
     titleIdx = plotphase ? [1,2,5,6] : [1,2,3,4]
-    titles[titleIdx] = ["S = 1/(1+PC)", "D = P/(1+PC)", "N = C/(1+PC)", "T = PC/(1+PC)"]
+    titles[titleIdx] = ["S = 1/(1+PC)", "P/(1+PC)", "C/(1+PC)", "T = PC/(1+PC)"]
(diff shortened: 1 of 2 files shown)


src/plotting.jl
```diff
@@ -788,7 +803,7 @@ function gangoffourplot(P::Union{Vector, LTISystem}, C::Vector, args...; minimal
     titles = fill("", 1, plotphase ? 8 : 4)
     # Empty titles on phase
     titleIdx = plotphase ? [1,2,5,6] : [1,2,3,4]
-    titles[titleIdx] = ["S = 1/(1+PC)", "D = P/(1+PC)", "N = C/(1+PC)", "T = PC/(1+PC)"]
+    titles[titleIdx] = ["S = 1/(1+PC)", "P/(1+PC)", "C/(1+PC)", "T = PC/(1+PC)"]
```
albheim, line 806: Skipped PS= and CS= on purpose? Can see that it might not be very needed, just thought it looked a little off having an equals sign in two of them and two were missing.

olof3: The sensitivity and complementary functions are commonly denoted by `S` and `T`. There is no such common notation for the other two, so I find this quite natural. One might consider using `PS` and `CS`, but that seems redundant. Or `G_yd` and `G_un`.

olof3:
@JuliaControlBot test-plots

baggepinnen:
@JuliaControlBot test-plots

PR #487, opened 2021-05-13
Add tustin c2d/d2c method

baggepinnen:


src/discrete.jl
```diff
@@ -49,7 +49,16 @@ function c2d_x0map(sys::AbstractStateSpace{<:Continuous}, Ts::Real, method::Symb
...
         x0map = I(nx)
-    elseif method === :tustin || method === :matched
+    elseif method === :tustin
+        a > 0 || throw(DomainError("A positive a must be provided for method Tustin"))
+        AI = (I(nx)-a*A)
+        Ad = AI\(I(nx)+a*A)
+        Bd = 2a*(AI\B)
+        # Cd = C/AI
+        Cd = C/AI 
+        Dd = a*Cd*B + D
```
albheim, line 59: Just quickly googled and found [this](https://math.aalto.fi/~jmalinen/MyPSFilesInWeb/milano-talk.pdf) talk from some math department in finland where they present the transformation in a similar way. The input-output mapping seems to be exactly the same, they have sigma=1/a, but I think the states will be different in this and their implementation since theirs has split the 2a to sqrt(2a) each on B and C. Not sure what is correct, but felt I should mention it.

baggepinnen: I looked around quite a bit before implementing this, and many references I found were actually wrong and did not result in `c2d/d2c` being inverses of each other. Regarding the constant on `C/B/both`, it doesn't matter much since `C(sI -A)\B` will be the same, I just chose the version that avoided the sqrt to get slightly simpler expressions.

src/discrete.jl
```diff
@@ -49,7 +49,16 @@ function c2d_x0map(sys::AbstractStateSpace{<:Continuous}, Ts::Real, method::Symb
...
+        x0map = I(nx)
```
olof3, line 60: Would be good with a reference.

 What is `a` (not clear from the docs)? I guess one would like to specify a prewarp frequency (rather than its inverse?)? Perhaps `ω_prewarp`, `prewarp_freq`, or possibly `α`?

Should an error be thrown if the user tries to specify a prewarp frequency for other methods than `:tustin`.
 
* Update docstring.
* Remove comment left in code.

test/test_discrete.jl
```diff
@@ -78,11 +78,14 @@ end
 
 # forward euler
 @test c2d(C_111, 1, :fwdeuler).A == I + C_111.A
-@test d2c(c2d(C_111, 0.01, :fwdeuler), :fwdeuler) ≈ C_111
-@test d2c(c2d(C_212, 0.01, :fwdeuler), :fwdeuler) ≈ C_212
-@test d2c(c2d(C_221, 0.01, :fwdeuler), :fwdeuler) ≈ C_221
-@test d2c(c2d(C_222_d, 0.01, :fwdeuler), :fwdeuler) ≈ C_222_d
-@test d2c(c2d(G, 0.01, :fwdeuler), :fwdeuler) ≈ G
+method = :tustin
```
olof3, line 81: This line doesn't seem to do anything?

Perhaps would be good with some more explicit tests for `:tustin` with/without prewarping, rather than just going back and forth with `c2d` and `d2c`.

src/discrete.jl
```diff
@@ -49,7 +49,16 @@ function c2d_x0map(sys::AbstractStateSpace{<:Continuous}, Ts::Real, method::Symb
     elseif method === :fwdeuler
         Ad, Bd, Cd, Dd = (I+Ts*A), Ts*B, C, D
         x0map = I(nx)
-    elseif method === :tustin || method === :matched
+    elseif method === :tustin
+        a > 0 || throw(DomainError("A positive a must be provided for method Tustin"))
+        AI = (I(nx)-a*A)
```
olof3, line 54: Why `I(nx)` here but just `I` further down. Would probably just have just used `X` or something instead of `AI`. 

Won't there be type instability from that `x0map` is either `Diagonal` or dense (?)

olof3 (commented):
Will be a nice addition to have Tustin transformations.

src/discrete.jl
albheim: Yeah, definitely not my expertise and I assumed you knew what you were doing. Just thought I would check since it looked different, and had a quick look at the stationary behavior from inputs to states and it seems like your version mimics that of the continuous system while the version I found does not, not sure why I didn't check that before my comment.

test/test_discrete.jl
baggepinnen: added tests vs matlab

src/discrete.jl
baggepinnen: type instability and incosistency fixed

src/discrete.jl
baggepinnen: added `f_prewarp` and documented what it does. The behaviour of `f_prewarp` is the same as frequency prewarping in matlab.

olof3: Ah, `f_prewarp` is nice, but then I would assume it to be in Hz = 1/s and not rad/s. That doesn't seem to be how it is handled.

Personally, I quite like Hz, but I guess people are more used to rad/s

baggepinnen: matlab calls the argument `PrewarpFrequency` and treats it as `rad/s` so I figured this was consistent with that. All other functions in the toolbox that accepts frequencies does so in `rad/s` so it would be strange to deviate from it here. For a user to figure out what keyword to use for prewarping, they would have to read the docstring anyway.

olof3: Personally, I wouldn't continue reading the docstring after finding `f_prewarp` in the function signature :)

Come to think of it, why not `w_prewarp`, if we are to avoid unicode. Anything other than `f` would be a real foot saver.

baggepinnen: Hmm, I can't see how `f = frequency = Hz`, but if it is a big problem we can rename it. `w_prewarp` is not really short for anything though, so it relies on the reader making the connection `w = ω = natural frequency`

olof3: Much appreciated. When I google `frequency` the first thing I get is a panel with `f = 1 / T` (i.e., Hz) :)

baggepinnen: I've changed it to `w_prewarp`. A large number of docstrings in the package talks about frequencies, implying frequencies in rad/s though, including` freqresp, bode, nyquist, sigma`.

olof3: Yes, there is definitely some abuse of terminology. As long as `w/\omega` is in rad/s and `f` is in Hz, then I'm quite happy, but perhaps we should be more careful with this.

PR #488, opened 2021-05-13
add frequency in Hz to dampreport

baggepinnen:


albheim:
Is it often you want to have both types of frequencies, or is it just to cover so that you always have the one currently needed? Just thinking if it becomes cluttered when printing, but it seems like it should not be any overwhelming amounts of text. The alternative could be to have a parameter that sets if you want it in Hz or rad/s, but maybe that just becomes messier to have to deal with.

Looks good to me.

baggepinnen:
The table is narrow enough (82 chars) to fit into the julia style-guide recommendation of 92 cols, so I think it should be fine. I often work in Hz, some in rad/s, the easiest is to just show both, makes the hassle of testing printing slightly easier also to just have one case.

PR #489, opened 2021-05-13
add interface to dsp

baggepinnen:
needs more tests


src/dsp.jl
```diff
@@ -0,0 +1,31 @@
+ControlSystems.tf(p::DSP.PolynomialRatio, h::Real = 1) = tf(DSP.coefb(p), DSP.coefa(p), h)
+ControlSystems.tf(p::DSP.ZeroPoleGain, h::Real = 1) = tf(DSP.PolynomialRatio(p), h)
```
albheim, line 2: Don't think you need the ControlSystems specifier, same in two other spots further down in the same file.

baggepinnen: correct, artifact of the development procedure :P I'll remove when I get around to push some tests :)

PR #491, opened 2021-05-13
fix traces in rlocus

baggepinnen:
prevent traces from jumping around


src/pid_design.jl
```diff
@@ -115,12 +115,14 @@ rlocus
 @recipe function rlocus(p::Rlocusplot; K=Float64[])
     P = p.args[1]
     K = isempty(K) ? range(1e-6,stop=500,length=10000) : K
+    K isa Number && (K = range(1e-6,stop=K,length=10000))
```
albheim, line 118:
```suggestion
@recipe function rlocus(p::Rlocusplot; K=500)
    P = p.args[1]
    K = K isa Number ? range(1e-6,stop=K,length=10000) : K
```

olof3:
LGTM

PR #494, opened 2021-05-13
use MatrixEquations.jl

baggepinnen:
The PR adopts MatrixEquations.jl and gets rid of our implementations of `care, dare, dlyap` but keeps the interfaces and the function names, since MatrixEquations uses a slightly different convention.


imciner2:
Nice. I actually started on this a while ago alongside adding cross-term support to the LQR, but then I saw a comment by I think @olof3 that they wanted to just rewrite the internal DARE/CARE functions instead, so I stopped the porting.

baggepinnen:
Yeah, the option to reimplement functionality will still exist, but until that has happened, ME provides some nice additional features we did not have, and was generally more robust than our own implementations :)

olof3:
Yes, I haven't had too much time to deal with that unfortunately. I think that would be the ideal solution, but anything is much better than what currently is in the toolbox.

baggepinnen:
I'll pull this now since it's non-breaking. The possibility to make use of ControlMatrixEquations at a later stage remains a possibility if that is deemed beneficial

PR #495, opened 2021-05-13
allow balance when converting tf to ss

baggepinnen:
the reason it defaults to `false` is that the balancing often fails, producing an annoying warning.

### DIFF
--- a/src/types/StateSpace.jl
+++ b/src/types/StateSpace.jl
@@ -53,7 +53,7 @@ end
 """ If D=0 then convert to correct size and type, else, create 1x1 matrix"""
 function fix_D_matrix(T::Type,B,C,D)
     if D == 0
-        D = fill(zero(T), size(C,1), size(B,2))
+        D = zeros(T, size(C,1), size(B,2))
     else
         D = to_matrix(T, D)
     end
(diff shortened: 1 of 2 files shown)


src/types/conversion.jl
```diff
@@ -78,13 +78,12 @@ Base.convert(::Type{HeteroStateSpace}, s::StateSpace) = HeteroStateSpace(s)
 Base.convert(::Type{StateSpace}, s::HeteroStateSpace) = StateSpace(s.A, s.B, s.C, s.D, s.Ts)
 Base.convert(::Type{StateSpace}, s::HeteroStateSpace{Continuous}) = StateSpace(s.A, s.B, s.C, s.D)
 
-function Base.convert(::Type{StateSpace}, G::TransferFunction{TE,<:SisoTf{T0}}) where {TE,T0<:Number}
-
+function Base.convert(::Type{StateSpace}, G::TransferFunction{TE,<:SisoTf{T0}}; kwargs...) where {TE,T0<:Number}
     T = Base.promote_op(/,T0,T0)
     convert(StateSpace{TE,T}, G)
```
albheim, line 83:
```suggestion
    convert(StateSpace{TE,T}, G; kwargs...)
```
I assume you missed this? Or I'm not understanding something here.

src/types/StateSpace.jl
```diff
@@ -53,7 +53,7 @@ end
 """ If D=0 then convert to correct size and type, else, create 1x1 matrix"""
```
albheim, line 53:
```suggestion
""" If D=0 then create zero matrix of correct size and type, else, convert D to correct size and type"""
```

olof3:
```suggestion
""" If D=0 then create zero matrix of correct size and type, else, convert D to correct type"""
```
I agree. Sounds a bit off that the size of D should be converted in that case, though.

olof3:
Looks good.

When we looked at this the last time, it seemed like there was a relatively easy way to put up the statespace realizations of each SISO transfer function that improved the balancing quite a bit in many cases. But that's for another day.

src/types/StateSpace.jl
albheim:
```suggestion
""" If D=0 then create zero matrix of correct size and type, else, convert D to correct shape and type"""
```
Is what I intended to write, since `to_matrix` might to a reshape if D is a vector. But either is fine for me.

PR #498, opened 2021-05-13
correct handling of x0 in lsimplot

baggepinnen:


albheim:
Seems good, if you want more control (access to all keywords in `lsim`) you can easily run `lsim` and plot it yourself, but `x0` seems reasonable to have access to.

LGTM

albheim (approved):

PR #502, opened 2021-05-13
pi place and tests

albheim:
Named it `placePI` to conform to the other methods there.


src/pid_design.jl
```diff
@@ -302,3 +302,32 @@ function loopshapingPI(P,ωp; ϕl=0,rl=0, phasemargin = 0, doplot = false)
...
+    kp, ki, C = placePI(P, ω₀, ζ; series=false)
+
+Selects the parameters of a PI-controller such that the poles are placed
+to match the poles of s^2 + 2ζω₀ + ω₀^2.
+
+Default is to return the parameters for the parallel form, but if `series=true`
+they will be returned on series form.
+"""
+function placePI(P, ω₀, ζ; series=false)
+    num = numpoly(P)[].coeffs
```
baggepinnen, line 316:
```suggestion
    num = numvec(P)[]
```
but this might give coefficients in the reverse order

albheim: Haha, thought we should have something like that... Had all my variables in switched order first.

src/pid_design.jl
```diff
@@ -302,3 +302,32 @@ function loopshapingPI(P,ωp; ϕl=0,rl=0, phasemargin = 0, doplot = false)
...
+"""
+    kp, ki, C = placePI(P, ω₀, ζ; series=false)
+
+Selects the parameters of a PI-controller such that the poles are placed
+to match the poles of s^2 + 2ζω₀ + ω₀^2.
+
+Default is to return the parameters for the parallel form, but if `series=true`
+they will be returned on series form.
+"""
+function placePI(P, ω₀, ζ; series=false)
```
olof3, line 315: I guess the kwarg should be `time` in analogy with the pid function? Although I feel that something more descriptive than `time` would be very good.

src/pid_design.jl
```diff
@@ -302,3 +302,32 @@ function loopshapingPI(P,ωp; ϕl=0,rl=0, phasemargin = 0, doplot = false)
...
+    c, d = den
+    tmp = (a^2*ω₀^2 - 2*a*b*ω₀*ζ + b^2)
+    kp = -(a*c*ω₀^2 - 2*b*c*ω₀*ζ + b*d) / tmp
+    ki = (ω₀^2*(b*c - a*d)) / tmp
+    C = pid(;kp, ki)
+
+    if !series
+        ki /= kp 
+    end
+    return kp, ki, C
```
olof3, line 332: Should the parameters be returned? Anyway, I would say that the controller is the most interesting object and should be first?

Also, if we add a PID place, the controller would be the fourth element which seems a bit inconsistent.

Perhaps the PI parameters should be in a tuple?

test/test_pid_design.jl
```diff
@@ -23,4 +23,9 @@ kp,ki,C = loopshapingPI(P,10; phasemargin = 30, doplot = false)
 _,_,_,pm = margin(P*C)
 @test pm[] > 30
 
+P = tf(1,[1, 1])
+kp,ki,C = placePI(P, 2, 0.7)
+@test pole(feedback(P, C)) ≈ [-1.4 + 1.4282856857085702im, -1.4 - 1.4282856857085702im]
```
olof3, line 28:
```suggestion
@test pole(feedback(P, C)) ≈ [-1.4 + √2.04im, -1.4 - √2.04im]
```

test/test_pid_design.jl
```diff
@@ -23,4 +23,9 @@ kp,ki,C = loopshapingPI(P,10; phasemargin = 30, doplot = false)
...
+@test [kp, ki] ≈ [1.7999999999999998, 2.2222222222222223]
```
olof3, line 29:
```suggestion
@test [kp, ki] ≈ [9/5, 20/9]
```
I'd guess that this is what you want to check?

src/pid_design.jl
baggepinnen: `pid` has `series` as keyword well, in this case it's about `series / parallel form`.

src/pid_design.jl
baggepinnen: If the function is used to tune a PI controller that is implemented in hardware, you want the parameters. One option is to return a named tuple, `(; kp, ki)` and `(; kp, ki, kd)`, they destructure like tuples but also have the self-documenting names in there, so you know what you get back without reading the docstring.

olof3: Okay, so we should keep the parameters and return them in a tuple.

Named tuple seems convenient, but what about (kp, ki) vs (kp, Ti)? I guess we would need a Val argument to avoid type instability?

src/pid_design.jl
olof3: Hmm, according to https://en.wikipedia.org/wiki/PID_controller (and it seems consistent with Advanced PID Control), there are 

* standard form: `Kp*(1 + 1/Ti/s + Td*s)`
* parallel form: `Kp + Ki/s + Kd*s`
* series form: `Kc*(1 + 1/τi/s)*(τd*s + 1)`

So for the `pid` function I guess one should just have to specify `form = :standard/:parallel/:series`? Currently there are two boolean inputs `time` and `series` for specifying one of four different forms. I.e. one combination is nonstandard, and it is not really clear to me how those combinations map to the above, although I could guess.

For the PI case that we are discussing here it is about standard/parallel? So the kwarg should actually be `parallel=false`.

baggepinnen: Two completely different control systems (one European one Chinese) we are working with at work both define a PI controller as
```
Kp + 1/(s*Ti)
```
which is somewhere inbtween. In fact, I believe most robot systems out there define PI controllers like this. It's a parallel form but the integral parameter treated as integral time rather than gain

src/pid_design.jl
baggepinnen: Val arguments are almost never required anymore since the compiler is much more aggressive with constant propagation, so if the option
```
placePI(..., form=:Ti)
```
is hard coded, the compiler will likely treat the `form` as a compile-time constant and propagate it

src/pid_design.jl
albheim: Hmm, seems like a lot of cases to cover. I also recognize the ones that Olof linked and was a bit confused with the time keyword. 

Would it be easier to just add that as a fourth form and do `form = :standard/:parallel/:series/:parallel_time` or something like that instead of having two keywords? I guess supporting all of them should not be a lot of extra work since most transformations are easy. Can just do all calculations on one form in the toolbox and add a `pi_form_convert` method that converts from that oen to any of the other ones.

src/pid_design.jl
albheim: Returns as named tuple now, with different names depending on selected form.

albheim:
If we prefer this way, using form and returning named tuples, we should make the interface similar for all other PI methods as well

src/pid_design.jl
baggepinnen: A function that converts between different conventions is perhaps a good idea, it would be a mess to support all conventions in every single function that handles pid parameters :/

src/pid_design.jl
olof3: > Val arguments are almost never required anymore
Great to hear.

src/pid_design.jl
olof3: > Two completely different control systems (one European one Chinese) we are working with at work both define a PI controller as
Kp + 1/(s*Ti)

In that case Ti doesn't correspond to integral time.
I've actually also encountered that formulation in the wild :P  

If this is something that you want then we should give it a name that avoids accidental usage, perhaps :nonstandard ?

Yes, some conversion functionality from standard or (parallel) form would be good.

albheim:
Okay to merge this, and leave additional fixes for later?

src/pid_design.jl
```diff
@@ -302,3 +302,50 @@ function loopshapingPI(P,ωp; ϕl=0,rl=0, phasemargin = 0, doplot = false)
...
+Selects the parameters of a PI-controller such that the poles of 
+closed loop between `P` and `C` are placed to match the poles of 
+`s^2 + 2ζω₀ + ω₀^2`.
+
+The form keyword allows you to choose which form the PI parameters
+should be returned on. 
+* `:standard` - `Kp*(1 + 1/Ti/s)`
+* `:series` - `Kc*(1 + 1/τi/s)`
+* `:parallel` - `Kp + Ki/s`
+* `:Ti` - `Kp + 1/(s*Ti)`
```
olof3, line 318:
```suggestion
* `:Ti` - `Kp + 1/(s*Ti)`   (non-standard form sometimes used in industry)
```

A bit confusing that it is named as a parameter. Or if we call it :nonstandard (which is a quite polite thing to call it)

src/pid_design.jl
```diff
@@ -302,3 +302,50 @@ function loopshapingPI(P,ωp; ϕl=0,rl=0, phasemargin = 0, doplot = false)
...
 end
+
+"""
+    piparams, C = placePI(P, ω₀, ζ; form=:standard)
+
+Selects the parameters of a PI-controller such that the poles of 
+closed loop between `P` and `C` are placed to match the poles of 
+`s^2 + 2ζω₀ + ω₀^2`.
+
+The form keyword allows you to choose which form the PI parameters
```
olof3, line 313:
```suggestion
The `form` keyword allows you to choose which form the PI parameters
```

olof3:
> Okay to merge this, and leave additional fixes for later?

Yes, but would be good to get it done somewhat soon. Seems like most of the issues have been resolved. Would be good to have `pid` work in the same way.

src/pid_design.jl
```diff
@@ -302,3 +302,50 @@ function loopshapingPI(P,ωp; ϕl=0,rl=0, phasemargin = 0, doplot = false)
...
+
+`piparams` is a named tuple with the controller parameters.
+
+`C` is the transfer function of the controller.
+
+"""
+function placePI(P, ω₀, ζ; form=:standard)
+    num = numvec(P)[]
+    den = denvec(P)[]
+    length(den) == 2 || error("Can only place poles using PI if the system if of first order.")
```
olof3, line 328:
```suggestion
    length(den) == 2 || error("Can only place poles using PI if the system if of first order.")
```
```suggestion
    if length(den) != 2 || length(num) > 2
        error("Can only place poles using PI for proper first-order systems")
    end
```
Hmm, perhaps we should add something like this.

src/pid_design.jl
albheim: Yeah, was thinking about that. Took it from bagges exampel above. Maybe switch to :parallel_time or something similar?

src/pid_design.jl
albheim: I skipped that since I for some reason assume that `length(num)>length(den)`, but I guess that is if you want the system to be realisable and you can create systems which are not in the toolbox.

PR #508, opened 2021-05-14
Use nstates instead of nx in lsimplot

albheim:
delay_example did not run for me, seemed to be this

### DIFF
--- a/src/types/TransferFunction.jl
+++ b/src/types/TransferFunction.jl
@@ -24,7 +24,6 @@ end
 
 noutputs(G::TransferFunction) = size(G.matrix, 1)
 ninputs(G::TransferFunction) = size(G.matrix, 2)
-nstates(G::TransferFunction) = length(pole(G)) # Does not work for MIMO 
 
 #####################################################################
 ##                          Misc. Functions                        ##
--- a/src/types/TransferFunction.jl
+++ b/src/types/TransferFunction.jl
@@ -24,6 +24,7 @@ end
 
 noutputs(G::TransferFunction) = size(G.matrix, 1)
 ninputs(G::TransferFunction) = size(G.matrix, 2)
+nstates(G::TransferFunction) = length(pole(G)) # Does not work for MIMO 
 
 #####################################################################
 ##                          Misc. Functions                        ##
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -139,7 +139,7 @@ lsimplot
     s2i(i,j) = LinearIndices((ny,1))[j,i]
     for (si,s) in enumerate(systems)
         s = systems[si]
-        y, t = lsim(s, p.args[2:end]...; x0 = x0 == 0 ? zeros(s.nx) : x0)
+        y, t = lsim(s, p.args[2:end]...; x0 = x0 == 0 ? zeros(nstates(s)) : x0)
         seriestype := iscontinuous(s) ? :path : :steppost
         for i=1:ny
             ytext = (ny > 1) ? "Amplitude to: y($i)" : "Amplitude"


olof3:
I think we are moving in the other direction actually, and that we instead should define the `getproperty` for `.nx`. And preferably deprecate `nstates`, `noutputs`, `ninputs`. But I'm not sure that this is the plan or if there is a plan. @mfalt @baggepinnen

I think that I prefer `sys.nx` since it is shorter.

baggepinnen:
I generally prefer nx, but a function is much easier to broadcast and I don't think we should not depreciate those since it's a hassle to broadcast getproperty. Defining properties for transfer functions would be good though.

albheim:
Ah, okay. I thought it was going towards `nstates`. 

After a quick look it seems that `tf`, `sisorational` and `sisozpk` does not have either, `delaylti` has `nstates` and `ss` has `nx`.

Could we just have all of them implement `getproperty` on `nu, ny, nx` and then have one spot where we define `ninputs(sys::LtiSystem) = sys.nu` and same for the other?

mfalt:
I do prefer `nstates(sys)` for it's clarity in the package. But I think defining getproperty so the user can use .nx is fine.

olof3:
> a function is much easier to broadcast

Okay, then we should definitely keep them around.

> After a quick look it seems that tf, sisorational and sisozpk does not have either, 

States are for statespace realizations. But I guess `order` or something like that could be useful for transfer functions, although it feels like we are coping quite well by just looking at the length of the denominator polynomial directly. Not sure what we do for `zpk`

> delaylti has nstates and ss has nx.

Then we should define `nstates` for `ss` if that is convenient for broadcasting.

When you bring it up, delay systems have infinitely many states, so I don't think we should define `nstates` or `nx` for those (although it was me who implemented `nstates` for it). Rather one should have to call that function on the statespace realization used for representing the delay system, and `nstates` on the `DelayLTISystem` should throw an error.

> I do prefer nstates(sys) for it's clarity in the package.

I partly see your point, and although I in general prefer long names, I don't see any problems here. I also think it is important with clarity in the package, and for this reason I think it is important to keep things short when there is no risk of confusion and then use longer names for less obvious things. 

This piece of code is a good example:

https://github.com/JuliaControl/ControlSystems.jl/blob/484b00fc870ca83825a6e677fcb143e1ba89f327/src/connections.jl#L371

of course, one could have defined variables `G_ny` `Δ_nu`, but that seems quite cumbersome and could potentially introduce errors, rather than just using `G.ny`, `Δ.nu` directly.


Here is an other less extreme example, but already here i find it a lot harder to quickly see what goes on in the first version of the code.
```julia
    A = [s1.A                             zeros(T, nstates(s1), nstates(s2));
            zeros(T, nstates(s2), nstates(s1))           s2.A]
```

```julia
    A = [s1.A                         zeros(T, s1.nx, s2.nx);
           zeros(T, s2.nx, s1.nx)            s2.A]
```


The number of states also very much feels like an inherent `property` of a statespace realization, so to me it feel much more natural to access it as a property rather than having to call a function on the object.

baggepinnen:
> When you bring it up, delay systems have infinitely many states, so I don't think we should define nstates or nx for those (although it was me who implemented nstates for it). Rather one should have to call that function on the statespace realization used for representing the delay system, and nstates on the DelayLTISystem should throw an error.

I don't quite understand the suggestion, which statespace realization? Wouldn't it just be an annoyance and break a lot of generic code to not define `nx` for delay systems? 

> The number of states also very much feels like an inherent property of a statespace realization, so to me it feel much more natural to access it as a property rather than having to call a function on the object.

So does `length, size, ndims` of an array? Those are still functions in julia because functions are much more powerful than properties. The only upside with properties is visual.

olof3:
> I don't quite understand the suggestion, which statespace realization? 

The class of delay systems the `DelayLTISytem` models is those that can be expressed as a linear fractional transformation between a statespace system `P` and a number of "delay buffers". The delay system has infinitely many states, while `P` has a finite number of stats. (Currently `P` is implemented with a special type `PartitionedStateSpace`, but as we have discussed previously , we might move away from that. But that is an unrelated issue.)

> Wouldn't it just be an annoyance and break a lot of generic code to not define nx for delay systems?

If it does, then I guess that something is wrong. Now I'm curious if something breaks, so I think I'll try it out and possibly make a PR.

> >    The number of states also very much feels like an inherent property of a statespace realization, so to me it feels much more natural to access it as a property rather than having to call a function on the object.
> So does length, size, ndims of an array? Those are still functions in julia because functions are much more powerful than properties. The only upside with properties is visual.

Ah, yes, that's a fair point. We should definitely keep the function versions around.

There are a few places in the code where `nstates(sys)` looks slightly better, but in many cases it would look significantly worse. If we want to go for consistency then I think we should go with `.nx`, but otherwise we can just take what feels more natural or whatever we feel like.

In the currently discussed case I'd say `.nx` only looks insignificantly better, so in this case I don't `nstates`.

baggepinnen:
> The class of delay systems the DelayLTISytem models is those that can be expressed as a linear fractional transformation between a statespace system P and a number of "delay buffers". The delay system has infinitely many states, while P has a finite number of stats. 

Ah, this makes sense. I was somehow only thinking about the `PartitionedStateSpace` type. 

About the system dimensions, I find it the cleanest to introduce local variables with the sizes, e.g,
```
nx,nu,ny = sys.nx, sys.nu, sys.ny
```
and then use those in the function body. When doing that, 
```
nx,ny,nu = nstates(sys), noutputs(sys), ninputs(sys)
```
would actually be more descriptive, but I often write the former since it's shorter.

olof3:
```julia
ny, nu = size(sys)
nx = nstates(sys)
```
slightly shorter, although two lines.

So if one is repeatedly using the dimensions I think this is quite nice, and here I'm in slightly in favor of `nstates`.

However, if one introduces `nu`, `nx` etc, in a lot of places in the code, then it does not seem very unclear to just use `sys.nx` if one only needs it in a few places.

However, it will be quite annoying if there is more than one system.

src/types/TransferFunction.jl
```diff
@@ -24,6 +24,7 @@ end
 
 noutputs(G::TransferFunction) = size(G.matrix, 1)
 ninputs(G::TransferFunction) = size(G.matrix, 2)
+nstates(G::TransferFunction) = length(pole(G)) # Does not work for MIMO
```
baggepinnen, line 27: This is perhaps not a good idea, do we really have to pay the price for an eigenvaluecomputation just to get this number.? If it's a siso system, it's enough to check the length of the denominator polynomial.

albheim, line 27:
```suggestion
```

albheim: Japp, ingen bra lösning. Tog bort.

PR #509, opened 2021-05-16
Pid parameters updates

albheim:
Suggestion of how to handle PID parameters.

The forms @olof3 mentioned in #502 felt familiar to me and they seemed to be somewhat standard, and then extend that with the `paralleltime` version that @baggepinnen was talking about and we get these
* standard form: `Kp*(1 + 1/Ti/s + Td*s)`
* series form: `Kc*(1 + 1/τi/s)*(τd*s + 1)`
* parallel form: `Kp + Ki/s + Kd*s`
* paralleltime form: `Kp + 1/Ti/s + Kd*s`
To remove the handling of the conversion between different forms it was lifted out of the `placePI` function, and then I also thought maybe we might as well have the user call it separately.

Not sure this is desired since I don't use it a lot myself. Is it preferred to have that both params and controller are always returned or should controller be returned from design and then you can extract parameters in desired form after?

Either way, this is not completed. It is just implemented for one method and not integrated with the rest since I didn't know how we wanted it.

### DIFF
--- a/src/pid_design.jl
+++ b/src/pid_design.jl
@@ -305,26 +305,15 @@ function loopshapingPI(P,ωp; ϕl=0,rl=0, phasemargin = 0, doplot = false)
 end
 
 """
-    piparams, C = placePI(P, ω₀, ζ; form=:standard)
+    C = placePI(P, ω₀, ζ)
 
 Selects the parameters of a PI-controller such that the poles of 
 closed loop between `P` and `C` are placed to match the poles of 
 `s^2 + 2ζω₀ + ω₀^2`.
 
-The `form` keyword allows you to choose which form the PI parameters
-should be returned on. 
-* `:standard` - `Kp*(1 + 1/Ti/s)`
-* `:series` - `Kc*(1 + 1/τi/s)`
-* `:parallel` - `Kp + Ki/s`
-* `:Ti` - `Kp + 1/(s*Ti)`   (non-standard form sometimes used in industry)
-
-`piparams` is a named tuple with the controller parameters.
-
-`C` is the transfer function of the controller.
-
+`C` is the returned transfer function of the controller.
 """
-function placePI(P, ω₀, ζ; form=:standard)
-    P = tf(P)
+function placePI(P::TransferFunction{<:Continuous, <:SisoRational{T}}, ω₀, ζ; form=:standard) where T
     num = numvec(P)[]
     den = denvec(P)[]
     if length(den) != 2 || length(num) > 2
@@ -339,17 +328,54 @@ function placePI(P, ω₀, ζ; form=:standard)
     tmp = (a*c*ω₀^2 - 2*b*c*ζ*ω₀ + b*d)
     Kp = -tmp / (a^2*ω₀^2 - 2*a*b*ω₀*ζ + b^2)
     Ti = tmp / (ω₀^2*(a*d - b*c))
-    C = pid(;kp=Kp, ki=Ti, time=true, series=true) 
+    # Use T somehow
+    return pid(;kp=Kp, ki=Ti, time=true, series=true) 
+end
 
-    if form === :standard
-        return (;Kp, Ti), C
-    elseif form === :series
-        return (;Kc=Kp, τi=Ti), C
+placePI(sys::LTISystem, args...; kwargs...) = placePI(tf(sys), args...; kwargs...)
+
+"""
+    params = pidparams(C; form=:standard)
+
+Extract PID parameters from a system representation of the controller. 
+Parameters can be extracted as one of several common representations, 
+options are 
+* `:standard` - `Kp*(1 + 1/(Ti*s) + Td*s)` 
+* `:series` - `Kc*(1 + 1/(τi*s))*(τd*s + 1)`
+* `:parallel` - `Kp + Ki/s + Kd*s`
+* `:paralleltime` - `Kp + 1/(Ti*s) + Kd*s`
+where the returned value is a named tuple with the converted parameters in.
+"""
+function pidparams(P::TransferFunction{<:Continuous, <:SisoRational{T}}; form=:standard) where T
+    num = numvec(P)[]
+    den = denvec(P)[]
+    num = [zeros(3 - length(num)); num]
+    den = [zeros(2 - length(num)); den]
+
+    if den[1] != 0 # Integrator
+        num ./= den[1]
+    else
+        num ./= den[2]
+    end
+    
+    K = num[2]
+    Ti = K / num[3]
+    Td = num[1] / K
+    
+    if form === :series
+        Kc = Kp*(Ti - sqrt(Ti*(-4*Td + Ti)))/(2*Ti)
+        τi = Ti/2 - sqrt(Ti*(-4*Td + Ti))/2
+        τd = Ti/2 + sqrt(Ti*(-4*Td + Ti))/2
+        return (;Kc, τi, τd)
     elseif form === :parallel
-        return (;Kp=Kp, ki=Kp/Ti), C
-    elseif form === :Ti
-        return (;Kp=Kp, Ti=Ti/Kp), C
+        return (;Kp=Kp, Ki=Kp/Ti, Kd)
+    elseif form === :paralleltime
+        return (;Kp=Kp, Ti=Ti/Kp, Td=Td)
+    elseif form === :standard
+        return (;Kp, Ti, Td)
     else
         error("Form $(form) not supported")
     end
 end
+
+pidparams(sys::LTISystem; form=:standard) = pidparams(tf(sys); form)
--- a/src/pid_design.jl
+++ b/src/pid_design.jl
@@ -1,4 +1,4 @@
-export pid, pidplots, rlocus, leadlink, laglink, leadlinkat, leadlinkcurve, stabregionPID, loopshapingPI, placePI
+export pid, pidplots, rlocus, leadlink, laglink, leadlinkat, leadlinkcurve, stabregionPID, loopshapingPI, placePI, pidparams
 
 """
     C = pid(; kp=0, ki=0; kd=0, time=false, series=false)
@@ -345,12 +345,15 @@ options are
 * `:parallel` - `Kp + Ki/s + Kd*s`
 * `:paralleltime` - `Kp + 1/(Ti*s) + Kd*s`
 where the returned value is a named tuple with the converted parameters in.
+
+The parameters can be accessed as `params.Kp` or `params["Kp"]` from the named tuple,
+or they can be unpacked using `Kp, Ti, Td = values(params)`.
 """
 function pidparams(P::TransferFunction{<:Continuous, <:SisoRational{T}}; form=:standard) where T
     num = numvec(P)[]
     den = denvec(P)[]
     num = [zeros(3 - length(num)); num]
-    den = [zeros(2 - length(num)); den]
+    den = [zeros(2 - length(den)); den]
 
     if den[1] != 0 # Integrator
         num ./= den[1]
@@ -358,9 +361,9 @@ function pidparams(P::TransferFunction{<:Continuous, <:SisoRational{T}}; form=:s
         num ./= den[2]
     end
     
-    K = num[2]
-    Ti = K / num[3]
-    Td = num[1] / K
+    Kp = num[2]
+    Ti = Kp / num[3]
+    Td = num[1] / Kp
     
     if form === :series
         Kc = Kp*(Ti - sqrt(Ti*(-4*Td + Ti)))/(2*Ti)
@@ -368,9 +371,9 @@ function pidparams(P::TransferFunction{<:Continuous, <:SisoRational{T}}; form=:s
         τd = Ti/2 + sqrt(Ti*(-4*Td + Ti))/2
         return (;Kc, τi, τd)
     elseif form === :parallel
-        return (;Kp=Kp, Ki=Kp/Ti, Kd)
+        return (;Kp=Kp, Ki=Kp/Ti, Kd=Kp*Td)
     elseif form === :paralleltime
-        return (;Kp=Kp, Ti=Ti/Kp, Td=Td)
+        return (;Kp=Kp, Ti=Ti/Kp, Td=Kp*Td)
     elseif form === :standard
         return (;Kp, Ti, Td)
     else


olof3:
This "reverse-engineering" of the parameters seems like an interesting and quite nice idea.

Although I in one way think it is nice that `placePI` only returns the controller, i feel like quite often it is nice to take a look at the parameters. It seems alright that they are returned with the controller  (although as the second argument).

Perhaps it is most convenient (rather than doing this reverse-engineering) if in `placePI`, the parameters for the standard form are computed, and then either returned, and that the user than can call a conversion function to get the desired form,  or that conversion function is called at the end of `placePI`. I don't really have a preference here.

I think it would be nice if `pid` would take the argument `form`, but I guess that is what you are aiming for?

Eventually it would also be nice if one for `pid` could specify the time-constant of the low-pass filter, and also directly obtain the controller as a StateSpace. There seems to be uncountable different ways to add low-pass filters, but the way it is done in Advanced PID control seems reasonable for a start.

baggepinnen:
> Eventually it would also be nice if one for pid could specify the time-constant of the low-pass filter, and also directly obtain the controller as a StateSpace

I second this, pretty much 100% of the time I need exactly these two, add a filter and then call ss. Would be nice to have the feature built in for numerical robustness as well as stability in the sense that `ss(g)` might change the realization returned, but `pid(StateSpace, ...)` could be made to return a certain realization at all times.

albheim:
> Perhaps it is most convenient (rather than doing this reverse-engineering) if in `placePI`, the parameters for the standard form are computed, and then either returned, and that the user than can call a conversion function to get the desired form, or that conversion function is called at the end of `placePI`. I don't really have a preference here.

Yeah, I'll update it to this. 

> I think it would be nice if `pid` would take the argument `form`, but I guess that is what you are aiming for?

Yeah, this was the plan.

> Eventually it would also be nice if one for `pid` could specify the time-constant of the low-pass filter, and also directly obtain the controller as a StateSpace. There seems to be uncountable different ways to add low-pass filters, but the way it is done in Advanced PID control seems reasonable for a start.

Had a go at that, last commit is not complete still but it contains a suggestion on how this could be approached. Not sure how I should introduce it in the form other than standard, so currently I just made a choice which seemed reasonable to me. But opinions are welcome here.

The `standard2form` and `form2standard` currently does not support `N` for a filter, the equations became very messy then. But it should not be a big thing to add if this structure looks good to you @olof3 and @baggepinnen, and when that works this should support both state space and filters in the `pid` function.

src/pid_design.jl
```diff
@@ -1,19 +1,78 @@
...
+created on state space form. If state space is selected either
+the derivative part has to be zero or `N` has to be provided to 
+allow for a state space realization. The state space will be
+returned on controllable canonical form.
+
+Options for `form` are 
+* `:standard` - `Kp*(1 + 1/(Ti*s) + Td*s/(1+Td*s/N))` 
+* `:series` - `Kc*(1 + 1/(τi*s))*(1 + τd*s/(1+τd*s/N))`
+* `:parallel` - `Kp + Ki/s + Kd*s/(1+Kd*s/N)`
+* `:paralleltime` - `Kp + 1/(Ti*s) + Kd*s/(1+Kd*s/N)`
```
albheim, line 17: Is it weird to have the same parameter name over different forms even though they do not represent the same thing? For example standard Ti is not the same as paralleltime Ti when converting between the forms.

src/pid_design.jl
```diff
@@ -328,47 +397,28 @@ function placePI(P::TransferFunction{<:Continuous, <:SisoRational{T}}, ω₀, ζ
...
-    
+function standard2form(form; params...)
+    Kp = get(params, :Kp, 0)
+    Ti = get(params, :Ti, inf)
+    Td = get(params, :Td, 0)
     if form === :series
-        Kc = Kp*(Ti - sqrt(Ti*(-4*Td + Ti)))/(2*Ti)
-        τi = Ti/2 - sqrt(Ti*(-4*Td + Ti))/2
-        τd = Ti/2 + sqrt(Ti*(-4*Td + Ti))/2
+        Kc = Kp/2 * (1 + sqrt(1 - 4*Td/Ti))
```
olof3, line 419: The series form is more restrictive than the other forms (since it can't have complex zeros (?)). Would probably be good with a helpful error message.

src/pid_design.jl
```diff
@@ -381,4 +431,38 @@ function pidparams(P::TransferFunction{<:Continuous, <:SisoRational{T}}; form=:s
...
+"""
+    params = form2standard(form; params...)
+
+Convert parameters from other forms to standard form.
+* `:standard` - `Kp*(1 + 1/(Ti*s) + Td*s)` 
+* `:series` - `Kc*(1 + 1/(τi*s))*(τd*s + 1)`
+* `:parallel` - `Kp + Ki/s + Kd*s`
+* `:paralleltime` - `Kp + 1/(Ti*s) + Kd*s`
+"""
+function form2standard(form; params...)
```
olof3, line 443: Shouldn't this function be used in the tf version for `pid`, seems like quite duplicated code.

The names for these functions are very non-descriptive. Would be good if they at least contained something with  `pid`and `param`.
Would it be feasible to detect what parameterizaiton the input has based on the fieldnames in the tuples? 
Then, one could perhaps get away with only one function pid_params_convert or convert_pid_params?

Are we planning on exporting this function? I would say perhaps not.

src/pid_design.jl
```diff
@@ -1,19 +1,78 @@
...
+C3 = pid(StateSpace; Kp=2, Ti=3)
+C4 = pid(StateSpace; form=:standard, Kp=2, Ti=3, Td=1, N=4)
+```
 """
-function pid(; kp=0., ki=0., kd=0., time=false, series=false)
+function pid(; form=:standard, params...)
     s = tf("s")
-    if series
-        return time ? kp*(one(kp) + one(kp)/(ki*s) + kd*s) : kp*(one(kp) + ki/s + kd*s)
+    N = get(params, :N, inf)
```
olof3, line 30: Not sure if this is the filtering that we should have (although this is definitely quite common) but the filtering that is used in Advanced PID Control, where different parameterizations are used for PI and PID.

It is definitely quite a mess to handle a lot of different approaches to filtering, but I would say that those two should be the top priority. Perhaps some other ones should be used as well.

How to get this into manageable code seems like a challenge.

There is a problem that different filters are used for PI and PID.

src/pid_design.jl
```diff
@@ -1,19 +1,78 @@
...
+        Kp = get(params, :Kp, 0)
+        Ti = get(params, :Ti, inf)
+        Kd = get(params, :Kd, 0)
+        return Kp + 1/(Ti*s) + Kd*s/(1+Kd*s/N)
+    else
+        error("Form $(form) not supported")
+    end
+end
+
+function pid(::Type{StateSpace}; form=:standard, params...)
```
olof3, line 56: Should the default output type be ss or tf? PID controllers are often expressed in the frequency domain, but ss in general give less numerical issues. Personally, I have a slight preference for ss.

olof3 (commented):
Good that there is progress on this!

src/pid_design.jl
olof3: That is correct, the `Ti` does not correspond to integral time, so it would be good to denote it with something else.

`paralleltime` is not a proper form, but for some reason it is not uncommon that software engineers implement PID controllers this way. I wouldn't mind getting rid of it altogether, but @baggepinnen had some use for it.

baggepinnen: That is how most industrial servos I know of implement the velocity control loop

src/pid_design.jl
albheim: Yeah, artifact from the coding process. Wrote pid first, then pidss where I realised I wanted the form2standard. But you are correct, should be used there.

src/pid_design.jl
albheim: Ahh, okay. Just looked up and took the first I found in the book, but see now that there is an alternative right below. If that form is preferred it seems almost easier to handle since the filter is outside the PI/PID controller. But will have a look at it next time I sit with this.

src/pid_design.jl
albheim: I don't mind either way. Is there a preference for what form should be used for the controller? I just took controllable canonical form here.

src/pid_design.jl
albheim: Yeah, was thinking what would happen here. But you are right, it should probably be handled in some way.

src/pid_design.jl
albheim: Do you have a preference for the names of the parameters, and maybe also what it should be called @baggepinnen 

I don't really mind if you want it to be Ti as well, just thinking if this could lead to confusion. But I feel like as long as it is clearly documented it should be fine.

src/pid_design.jl
albheim: Is the current version more what you were thinking?

src/pid_design.jl
baggepinnen: Coming to think about it a bit more. Most industrial servos have an outer position loop as well. If you consider the inner PI loop to constitute the expression `1/Ti + s*Kkd` and wrap a P controller outside of that, you get the standard form. In other words, as seen from the position, the cascade forms a PID structure, but if only the velocity loop is considered, you get the `:paralleltime` version above. I guess we can ditch `:paralleltime` then, as if only a PI controller for the velocity loop is considered, i.e., no position control, it's trivial to use the parallel form and just invert the integral parameter manually.

src/pid_design.jl
```diff
@@ -1,19 +1,78 @@
...
 """
-function pid(; kp=0., ki=0., kd=0., time=false, series=false)
+function pid(; form=:standard, params...)
     s = tf("s")
-    if series
-        return time ? kp*(one(kp) + one(kp)/(ki*s) + kd*s) : kp*(one(kp) + ki/s + kd*s)
+    N = get(params, :N, inf)
+    if form === :standard
+        Kp = get(params, :Kp, 0)
+        Ti = get(params, :Ti, inf)
```
baggepinnen:
```suggestion
        Ti = get(params, :Ti, typeof(Kp)(Inf))
```

src/pid_design.jl
```diff
@@ -1,19 +1,78 @@
...
+        Td = get(params, :Td, 0)
+        return Kp*(1 + 1/(Ti*s) + Td*s/(1+Td*s/N)) 
+    elseif form === :series
+        Kc = get(params, :Kc, 0)
+        τi = get(params, :τi, inf)
```
baggepinnen:
```suggestion
        τi = get(params, :τi, typeof(Kc)(Inf))
```

src/pid_design.jl
```diff
@@ -1,19 +1,78 @@
...
+        τd = get(params, :τd, 0)
+        return Kc*(1 + 1/(τi*s))*(1 + τd*s/(1+τd*s/N))
+    elseif form === :parallel
+        Kp = get(params, :Kp, 0)
+        Ki = get(params, :Ki, 0)
+        Kd = get(params, :Kd, 0)
+        return Kp + Ki/s + Kd*s/(1+Kd*s/N)
+    elseif form === :paralleltime
+        Kp = get(params, :Kp, 0)
+        Ti = get(params, :Ti, inf)
```
baggepinnen:
```suggestion
        Ti = get(params, :Ti, typeof(Kp)(Inf))
```

src/pid_design.jl
```diff
@@ -1,19 +1,78 @@
...
+        Kd = get(params, :Kd, 0)
+        return Kp + 1/(Ti*s) + Kd*s/(1+Kd*s/N)
+    else
+        error("Form $(form) not supported")
+    end
+end
+
+function pid(::Type{StateSpace}; form=:standard, params...)
+    params = form2standard(form; params...)
+    N = get(params, :N, inf)
```
baggepinnen:
```suggestion
    N = get(params, :N, Inf)
```

src/pid_design.jl
```diff
@@ -1,19 +1,78 @@
...
+    Kp = get(params, :Kp, 0)
+    Ti = get(params, :Ti, inf)
```
baggepinnen:
```suggestion
    Ti = get(params, :Ti, typeof(Kp)(Inf))
```

src/pid_design.jl
```diff
@@ -1,19 +1,78 @@
...
+    Td = get(params, :Td, 0)
+    if Td != 0 && N != inf
```
baggepinnen:
```suggestion
    if Td != 0 && isfinite(N)
```

src/pid_design.jl
```diff
@@ -328,47 +397,28 @@ function placePI(P::TransferFunction{<:Continuous, <:SisoRational{T}}, ω₀, ζ
...
-        num ./= den[2]
-    end
-    
-    Kp = num[2]
-    Ti = Kp / num[3]
-    Td = num[1] / Kp
-    
+function standard2form(form; params...)
+    Kp = get(params, :Kp, 0)
+    Ti = get(params, :Ti, inf)
```
baggepinnen, line 399:
```suggestion
    Ti = get(params, :Ti, typeof(Kp)(Inf))
```

src/pid_design.jl
olof3: Interesting. In the case were I encountered a "parallel time" form in the wild it was due to to  a misunderstanding. In this case I see that it makes a lot more sense.

src/pid_design.jl
olof3: Sure, sounds good.

src/pid_design.jl
albheim: Hmm, not sure I'm looking at this correctly but I don't get any integral part since the 1/(Ti*s) is on the qd signal which is s*q. Both qd and qdr enters the Kp to Kv line right?

baggepinnen: You do get integral action through the position loop. This kind of loop is a bit awkward to think about at first. One thing it allows is to have zero position feedback, something the standard form does not.

src/pid_design.jl
olof3: Even, if it is not exported, I think the function name should be better :P
Perhaps also an initial underscore.
`_pidṕarams_standard2other` and `_pidṕarams_other2standard` would be reasonably descriptive.`

albheim: I have made some local updates where it almost is working nicely with detecting parameters, and combining into a single function `convert_pid_params(target_form; params...)`. The problem is that I wanted to fix the tests also but got a bit stuck there since `pidplots` was not as easy to move over to this format and I could not decide how I wanted to do it, so put it to the side. But can push up what I have and you can have a look. Always good to get some input on how it should be designed.

src/pid_design.jl
```diff
@@ -339,17 +416,59 @@ function placePI(P, ω₀, ζ; form=:standard)
...
+* `:parallel` - `Kp + Ki/s + Kd*s`
+with default `params` values `Kp=Kc=1`, `Ti=τi=Inf`, `Ki=Kd=Td=τd=0`.
+
+`params` should be supplied with parameter names corresponding to the names used in the above 
+equations.
+"""
+function convert_pid_params(target; params...)
+    if haskey(params, :Kc)
+        Kc = params[:Kc]
+        τi = get(params, :τi, typeof(Kc)(Inf))
```
albheim, line 441: This typeof will make it fail when calling with an integer Kc and not supplying taui. Same for Kp/Ti. Not sure if that will be annoying or if there is a nicer way of doing it.

src/pid_design.jl
```diff
@@ -1,64 +1,124 @@
...
+function pidplots(P::LTISystem, args...; 
+        Kp=0.0, Kc=0.0, Ki=0.0, Kd=0.0, Ti=eltype(Kp)(Inf), τi=eltype(Kc)(Inf), Td=0.0, τd=0.0, 
+        ω=0, grid = false, kwargs...)
+    params = if Kc != 0
+        (;Kc, τi, τd)
+    elseif Ki != 0 || Kd != 0
+        (;Kp, Ki, Kd)
+    elseif Kp != 0
+        (;Kp, Ti, Td)
+    else
```
albheim, line 92: I'm still not sure I think this way of handling parameters is the best, selecting the form based on what keyword arguments were entered. Here it felt like it became a bit messy for example since I could not catch it with kwargs and send to the `convert_pid_params` to be handled.

src/pid_design.jl
```diff
@@ -245,33 +310,47 @@ arg(P) + arg(C) = -π
 
 If `P` is a function (e.g. s -> exp(-sqrt(s)) ), the stability of feedback loops using PI-controllers can be analyzed for processes with models with arbitrary analytic functions
 
-See also `stabregionPID`, `loopshapingPI`, `pidplots`
+See also `loopshapingPI`, `pidplots`
 """
-function stabregionPID(P, ω = _default_freq_vector(P,Val{:bode}()); kd=0, doplot = true)
+function stabregionPID(P, ω = _default_freq_vector(P,Val{:bode}()); form=:standard, kd=0)
```
albheim, line 315: Why is it called PID? Seems to only deal with PI.

Removed the `doplot` since it was not used.

src/pid_design.jl
```diff
@@ -1,64 +1,124 @@
 export pid, pidplots, rlocus, leadlink, laglink, leadlinkat, leadlinkcurve, stabregionPID, loopshapingPI, placePI
 
 """
-    C = pid(; kp=0, ki=0; kd=0, time=false, series=false)
-
-Calculates and returns a PID controller on transfer function form.
-- `time` indicates whether or not the parameters are given as gains (default) or as time constants
-- `series` indicates  whether or not the series form or parallel form (default) is desired
+    C = pid([type=StateSpace;] Tf=0, params...)
```
olof3, line 4: Again, feels slightly weird to provide parameters as kwargs, but if this is the most convenient way, then I don't mind too much.

olof3 (commented):
Looks good, seems to be on track.

src/pid_design.jl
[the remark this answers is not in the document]
albheim: We don't supply parameters to this method, we get them back. So it is about what form we want them back in.

src/pid_design.jl
albheim: I'm not sure it is. Your suggestion would be to pass them as a named tuple instead? I have nothing against that, just felt like this was probably a convenient way to define things, `pid(Kp=3, Ti=0.1)` or `pid((; Kp=2, Ti=0.1))` are pretty smiilar, but I would say the first is clearer. Or assuming you already have your params tuple, `pid(params...)` vs `pid(params)` is slightly shorter in the second case.

src/pid_design.jl
albheim: But we have to check which parameters where sent in so we know which ones to use for the grid, so how would I get away from this check here?

Yeah, missing or nothing could be more understandable in one way but then we loose the initial value that is used in the gridding.

src/pid_design.jl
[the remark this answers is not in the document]
olof3: For the other functions yes, but this one seems to produce a plot?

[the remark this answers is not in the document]
albheim: Not sure we are on the same page here. 

It produces a plot and returns a tuple of the fig and a vectors of parameters. The form is needed so we know what form the parameters should be returned on. Or am I missing something?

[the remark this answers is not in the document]
olof3: Aha, there are some parameters in the end, I didn't even notice those. Or it is actually a vector of parameters (?), that gives marginally stable closed loop systems.

Although the marginally stabilizing controllers does not seem like too useful of an output, I guess that the selected `pidform` should affect the plot, which is currently not the case. I.e., the labels in the plot will be wrong, and the naming of the temporary variables used for fiddling around with the parameters might have misleading names.

```julia
kps = map(x->values(x)[1], params)
```
etc.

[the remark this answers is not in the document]
olof3: It is not indicated what controller parameters that are returned, so if there is a need to return those, that should be specified.

src/pid_design.jl
olof3: Yes, that looks good, I would perhaps a factor so that the second state corresponds to the filtered control error.

```julia
A = [0 1 0; 0 0 1; 0 -2/Tf^2 -2/Tf]
B = [0; 0; 2 / Tf^2]
C = p.Kp * [1/p.Ti 1 p.Td]
D = 0
```

There is another version that one could consider, but this one looks nice and symmetrical.

olof3: Another thing is that one would like a first order filter for `pi` controllers. Not sure how get to that in a good way?

src/pid_design.jl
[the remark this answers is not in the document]
albheim: Yeah, I honestly didn't really think too much about it since I just wanted to update all the functions to use a similar interface. The plot naming is something that should be done if we allow for a form keyword.

src/pid_design.jl
albheim: Should it default to a first order filter for PI, or should that be an option? In the Advanced PID I thought they suggested the same filter for both, just different method of setting Tf, but I might have misunderstood.

src/pid_design.jl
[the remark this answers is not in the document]
albheim: Realised that you send in the D-parameter, so I guess that could be used to figure out the form. If feels like this is getting very messy, not sure I like handling all the forms all the time. Maybe just expose `convert_pid_param` instead, and have all our internal functions expect and return values in one of the forms, would be so much easier to handle and probably less confusing for the user to.

src/pid_design.jl
baggepinnen: You can use `typemax` to get the maximum value a type can represent
```julia
julia> typemax(Float64)
Inf

julia> typemax(Int)
9223372036854775807
```
not sure if it makes sense here though, but looking at how the value is used, I think it should be fine. Using `float` should probably also be fine since the value will enter a division which will produce floats from ints anyway. Another, perhaps more general solution is to give it the type 
```julia
Base.promote_op(/, typeof(Kc), typeof(Kc))

julia> Base.promote_op(/, Int, Int)
Float64
```
which avoids calling `float` for types where it does not make sense, e.g., `Particles`.

mfalt:
There is a lot here to digest, so I can't really comment on the PR as a whole. But i have a few comments:

1. We should probably either: not spend too much time on pidplots, since we want to turn everything into recipes, or: try to convert the code into a recipe.
2. I am not a fan of the restriction to the abstract types StateSpace and TransferFunction as the first argument to `pid`. I would expect a function with this structure to be able to handle `T=typeof(sys): pid(T, ...)`. One option is to just convert/promote to type T at the end, but then we have to make sure that we handle abstract types properly.

albheim:
> 1. We should probably either: not spend too much time on pidplots, since we want to turn everything into recipes, or: try to convert the code into a recipe.

I agree, felt I should do something about them since I started to change the interface or parameters, but didn't really have the time and energy to redo it properly so it was mostly a quick hack to get it to work with something somewhat similar.

> 2. I am not a fan of the restriction to the abstract types StateSpace and TransferFunction as the first argument to `pid`. I would expect a function with this structure to be able to handle `T=typeof(sys): pid(T, ...)`. One option is to just convert/promote to type T at the end, but then we have to make sure that we handle abstract types properly in covert/promote.

Sounds reasonable, then we would also create the controller with the same time evolution I assume, would that be just a call to `c2d` in the end? We don't keep any info on what method was used for discretization so we can't guarantee same type is used, but I guess just use `c2d` default? Or would it be better to calculate an analytical discretization of a PID and use that if time evolution is discrete?

olof3:
> I am not a fan of the restriction to the abstract types StateSpace and TransferFunction as the first argument to pid. I would expect a function with this structure to be able to handle T=typeof(sys): pid(T, ...). One option is to just convert/promote to type T at the end, but then we have to make sure that we handle abstract types properly in covert/promote.

I agree that this may seem slightly a bit off, on the other hand, however I don't feel like we want to worry about conversions etc. What about `Val(:ss)` and  Val(`:tf)`?

> then we would also create the controller with the same time evolution

Good point. Regardless, it should be possible to specify the time-evolution. I guess that we should preferably write up the discretizations explicitly, rather than relying on `c2d`.

albheim:
Finally sat down and had a look at this again. A bit unsure of how we exactly want this to look, but here is a new suggestion where I switched back the way of choosing form to supplying keywords instead of choosing form based on the parameter names that were used. It is not fixed all through, just wanted to check if the new method signature seemed reasonable to you?

I just went with ss/tf for now and not arbitrary type, since I felt it was messy enough for now. 

I'm still not completely sure on how we want the discrete version implemented. Do we implement it so we assume the user has already calculated the desired discrete parameters, or do we calculate some discrete parameters based on continuous ones supplied? I can see both being useful and it feels like covering all bases quickly becomes cumbersome with all keywords that would be needed to control all behaviours.

albheim:
> * Internally, when the parameters correspond to standard form, use, kp, Ti/ti, Td/td
> * Return named tuples, seems make sense, would avoid much uncertainty.

Both of these are okay with me, not the part I thought got messy really. What I had the most problem with was the use of identifying the form the user wanted by the names of kwargs.

> * For `pid` and other cases where the interpretation depends on the parameters, I think `kp`, `param_i`, `param_d` would be better than `ki`, `kd` even if it looks ugly, otherwise many people will probably be tripped up. But this is the least important of these three points, would also be a non-breaking change.

I have no strong opinion here, and in the cases where they are normal arguments I agree. Though there are some places where they are kwargs where it will matter more, and will be breaking, though I don't see that as much of a problem since much of this PR will be breaking anyway.

src/pid_design.jl
```diff
@@ -1,48 +1,100 @@
...
-
-Calculates and returns a PID controller on transfer function form.
-- `time` indicates whether or not the parameters are given as gains (default) or as time constants
-- `series` indicates  whether or not the series form or parallel form (default) is desired
+    C = pid(kp, ki, kd=0; form=:standard, Tf=0., Ts=0., state_space=false)
+
+Calculates and returns a PID controller. 
+
+The `form` can be chosen as one of the following
+* `:standard` - `kp*(1 + 1/(ki*s) + kd*s)`
```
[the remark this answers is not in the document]
albheim: Changed to `param_p, param_i, param_d` now.

src/pid_design.jl
```diff
@@ -1,48 +1,100 @@
...
     else
-        return time ? kp + one(kp)/(ki*s) + kd*s : kp + ki/s + kd*s
+        pid_tf(kp, ki, kd; form, Ts, Tf)
     end
 end
 
-"""
-    pidplots(P, args...; kps=0, kis=0, kds=0, time=false, series=false, ω=0)
+function pid_tf(kp, ki, kd; form, Ts, Tf)
+    kp, ki, kd = convert_pidparams_to_standard(kp, ki, kd, form)
```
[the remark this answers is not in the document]
albheim: So all for state space being default? Should I change the keyword to `transfer_function=false` then, since it feels better to me at least to trigger a setting by turning it to true. Or maybe just `tf=false` to make it nicer to write? Or is it nicer to have a `representation=:ss/:tf` keyword?

Currently I export `pid_tf` and `pid_ss` as well as `pid`, not sure if this is needed?

albheim:
Are the doctests not running anymore? I assumed they might fail since I haven't changed stuff there yet.

PR #510, opened 2021-05-17
restrict u to Mat in lsim

baggepinnen:
since we in fact do require it

### DIFF
--- a/src/timeresp.jl
+++ b/src/timeresp.jl
@@ -108,7 +108,7 @@ y, t, x, uout = lsim(sys,u,t,x0=x0)
 plot(t,x', lab=["Position" "Velocity"], xlabel="Time [s]")
 ```
 """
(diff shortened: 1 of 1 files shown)


baggepinnen:
I'm also a bit torn. Minimizing friction is worthwhile, but trying to be clever can sometimes backfire and make it more confusing in the end, since having time in the first dimension sometimes works and sometimes not.

olof3:
Yes, it will be quite annoying to create the `u` with the new convention. It's even worse for complex input signals where one can't use `'` :/

> Just thinking if it would be kinder to the user to allow vectors and catch it in the if statement that checks u size instead, to give them a clearer tip on how to solve it?

Yes, I think we should do this.

With a warning, one could consider fixing the transposition, although I also agree with

> Also agree that these things might back-fire.

PR #512, opened 2021-05-17
add controller function

baggepinnen:
This is yet another small PR that aims to break up the quite large https://github.com/JuliaControl/ControlSystems.jl/pull/467 into smaller chunks. 
This PR adds a function `controller` that takes state feedback and observer gains and returns the controller


olof3:
This seems like something that could occasionally be convenient.

I'm a little bit skeptical to using such a generic name for a function like this, though.
Wouldn't it be good with a more descriptive name, both to have a slightly better idea of what the function is about, and also if the user would like to introduce a variable called `controller` for example.

I'd say that `sscontroller` is slightly better,  but there are probably even better alternatives.
 
Hopefully the corresponding  `predictor` method could be renamed accordingly as well.

baggepinnen:
The most descriptive name I can come up with is `observer_statefeedback_controller`, but perhaps this can be conveyed in a slighltly shorter form?
Some alternatives
- `sf_obsv_controller`
- `innovation_form_controller`

olof3:
Yes, I think something like that would be alright. Doesn't one sometimes say "observer-based state feedback", so `obsv_sf_controller`?

src/matrix_comps.jl
```diff
@@ -653,7 +653,25 @@ function predictor(sys::ST, R1, R2) where ST <: AbstractStateSpace
...
+Such that `feedback(sys, cont)` produces a closed-loop system with eigenvalues given by `A-KC` and `A-BL`.
+
+# Arguments:
+- `sys`: Model of system
+- `L`: State-feedback gain `u = -Lx`
+- `K`: Observer gain
+"""
+function controller(sys, L::AbstractMatrix, K::AbstractMatrix)
+    A,B,C,D = ssdata(sys)
+    ss(A - B*L - K*C + K*D*L, K, L, 0, sys.timeevol)
```
albheim, line 676: Wouldn't this correspond to the output of the controller being `-Lx`?
```suggestion
    ss(A - B*L - K*C + K*D*L, K, -L, 0, sys.timeevol)
```

albheim: Ohh, you want the output to be `Lx` so that you can use negative feedback. Seems reasonable, but could maybe confuse more people than me. Maybe just state very clearly that the output of the controller is `Lx` and not `-Lx`.

olof3: Mm, yeah I guess this is alright.

But what about the case when one wants the controller to treat the measurement and reference signals differently. I would say that this convention is still acceptable, although it would feel weird to add the measurement signal to the reference signal, instead of subtracting it.

baggepinnen: When you introduce reference signals as well it can get very messy. In RobustAndOptimalControl.jl I'm experimenting with using a type `ExtendedStateSpace` which is similar to, and should probably be merged with `PartitionedStateSpace`, such that one can defined the controller with references as
```julia
function extended_controller(K)
    nx,nu,ny = K.nx, K.nu, K.ny
    A,B,C,D = ssdata(K)
    ss(A, B, -B, zeros(0,nx), C, D21=D, D22=-D)
end
```
i.e., it has two inputs, `r` and `y`. I haven't landed on how to represent everything yet though, I have also experimented with systems with named signals, ideally this should be used as well to not rely on indices everywhere.

olof3: Sounds good, not sure either what makes the most sense, so good that you are thinking about it. Yes, it seems more like a `PartitionedStateSpace`. We could change `PartitionedStateSpace`, I don't think we really need to explicitly use such a type for the delay systems.

Perhaps it won't even be needed here either with appropriate name of the signals.

Note: In this case one would have to use positive feedback?

baggepinnen:
I've tried to come up with a better name here. matlab has a simiar function that is named `lqgreg`
https://se.mathworks.com/help/control/ref/ss.lqgreg.html
They accept some kalman filter type into that which we don't do here. Also, a state feedback controller with observer does not have to be derived from an LQG problem. 

There is really two aspects here, the controller has an internal observer with gain `K`, and a state feedback. Can we come up with a name that is not horribly long and captures both of those aspects?  Perhaps `obsever_controller` is sufficient since an observer by definition observes the states and any controller using an observer would thus be a state feedback controller from observed states?

olof3:
I think what you said, but with a small reverse, i.e., `obsv_sf_controller` is fine. https://www.google.com/search?q="observer+based+state+feedback"

baggepinnen:
I changed both predictor and controller to
- `observer_predictor`
- `observer_controller`
Looks nice and symmetric, and `observer_controller` kind of implies state feedback so the slight loss in descriptiveness should be fine, the docstring will clarify if any questions arise.

PR #514, opened 2021-05-20
fix conversion for custom types

baggepinnen:
There was a type parameter missing in the convert method, this led to the method below never being called. 
I also changed the implentation for the fallback `numeric_type(sys::LTISystem)` to just grab the numeric type from the A-matrix of the system. This will never be called for our types, but it will make sure `numeric_type` works for all custom StateSpace types even if they haven't implemented `numeric_type(::Type{CustomStateSpace})`

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -8,7 +8,8 @@ numeric_type(::Type{TransferFunction{TE,S}}) where {TE,S} = numeric_type(S)
 numeric_type(::Type{<:StateSpace{TE,T}}) where {TE,T} = T
 numeric_type(::Type{<:HeteroStateSpace{TE,AT}}) where {TE,AT} = eltype(AT)
 numeric_type(::Type{<:DelayLtiSystem{T}}) where {T} = T
-numeric_type(sys::LTISystem) = eltype(sys.A)
+numeric_type(sys::AbstractStateSpace) = eltype(sys.A)
+numeric_type(sys::LTISystem) = numeric_type(typeof(sys))
 
 
 to_matrix(T, A::AbstractVector) = Matrix{T}(reshape(A, length(A), 1))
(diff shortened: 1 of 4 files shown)


src/utilities.jl
```diff
@@ -8,7 +8,7 @@ numeric_type(::Type{TransferFunction{TE,S}}) where {TE,S} = numeric_type(S)
 numeric_type(::Type{<:StateSpace{TE,T}}) where {TE,T} = T
 numeric_type(::Type{<:HeteroStateSpace{TE,AT}}) where {TE,AT} = eltype(AT)
 numeric_type(::Type{<:DelayLtiSystem{T}}) where {T} = T
-numeric_type(sys::LTISystem) = numeric_type(typeof(sys))
+numeric_type(sys::LTISystem) = eltype(sys.A)
```
olof3, line 11: Hmm, perhaps it should be done differently altogether, but this change doesn't look quite right.

Doesn't it break `numeric_type(tf(1))` for example? Doesn't this change override the definitions just above?

And if we are going to keep this, I think that it should at least not be defined for more than `AbstractStateSpace`, otherwise it seems quite doubtful to expect to find an `A` matrix.

baggepinnen: You're right, it requires specialization for `tf/ss`

olof3: Looks alright.

mfalt:
The big picture question here is how we should handle converts to non-concrete types. I guess we need (or prefer?) them in some cases, but then we should probably try to be more consistent? For example, if we have
`convert(::Type{TransferFunction}, ...`
should'nt we also have
`convert(TransferFunction{TE,SisoRational},...`?
What is the usual solution here? Do most packages allow converts to non-concrete types?

baggepinnen:
> What is the usual solution here? Do most packages allow converts to non-concrete types?

I'm not sure, can't recall having seen it very often at least, but `Base` does it for e.g., arrays
```julia
convert(::Type{T}, a::AbstractArray) where {T<:Array} = a isa T ? a : T(a)
```
by just redirecting to the constructor.
I just pushed a commit that does appear to solve the problem that failed the test, and also gets rid of an old comment in the code

PR #516, opened 2021-05-23
implement minreal through Kalman decomposition

baggepinnen:
The old `minreal` failed if the realization wasn't minimal, which kind of defeats the purpose of the function.

### DIFF
--- a/src/types/StateSpace.jl
+++ b/src/types/StateSpace.jl
@@ -425,14 +425,14 @@ function minreal(sys::ST,
     nnc = size(nullspace(C'; atol, rtol), 2) # number of non-controllable modes
     nc = sys.nx-nnc
 
-    nno = size(nullspace(O; atol, rtol), 2) # number of non-controllable modes
+    nno = size(nullspace(O; atol, rtol), 2) # number of non-observable modes
     no = sys.nx-nno
 
     T1 = diagonalizing(P, true)
 
     Q_block = inv(T1')*Q* inv(T1)
 
-    Q11 = Q_block[1:nc, 1:nc] # TODO: I just guessed to put no here
+    Q11 = Q_block[1:nc, 1:nc] 
     U1 = diagonalizing(Q11, false)
 
     Σ12b = abs.(diag(U1*Q11*U1')) # should be block(Σ₁², 0) 
--- a/src/types/StateSpace.jl
+++ b/src/types/StateSpace.jl
@@ -391,19 +391,73 @@ function Base.show(io::IO, sys::AbstractStateSpace)
     end
 end
 
-
-
-
 """
-`minsys = minreal(s::StateSpace, tol=sqrt(eps()))` is implemented via `baltrunc` and returns a system on diagonal form.
+Helper function for minreal. Returns `T` such that T*P*inv(T) is diagonal. If `unit`, T will diagonalize `P` to an identity matrix.
 """
-function minreal(s::AbstractStateSpace, tol=sqrt(eps()))
-    s = baltrunc(s, atol=tol, rtol = 0)[1]
-    try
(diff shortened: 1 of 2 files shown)


src/types/StateSpace.jl
```diff
@@ -391,19 +391,73 @@ function Base.show(io::IO, sys::AbstractStateSpace)
...
+Helper function for minreal. Returns `T` such that T*P*T' is diagonal. If `unit`, T will diagonalize `P` to an identity matrix.
+"""
+function diagonalizing(P::AbstractMatrix, unit=false; atol=1e-8)
+    if unit
+        S,U = eigen(Hermitian(P), sortby=x->-real(x))
+        i = findlast(>(atol*S[1]), S)
+        rm = length(S)-i
+        S[i+1:end] .= 0
+        T = (U*pinv(Diagonal(sqrt.(S))))'
+    else
```
olof3, line 404: This falls back to `eigen`, so I think it would be more readable to not consider this case separately. For `unit = false` there is also truncation going on, but not here?

src/types/StateSpace.jl
```diff
@@ -391,19 +391,73 @@ function Base.show(io::IO, sys::AbstractStateSpace)
     end
 end
 
-
-
+"""
+Helper function for minreal. Returns `T` such that T*P*T' is diagonal. If `unit`, T will diagonalize `P` to an identity matrix.
```
olof3, line 395: Does  `P` have to be Hermitian. Positive definite? What about the truncation?

I think a more expressive function name would be good.

src/types/StateSpace.jl
```diff
@@ -391,19 +391,73 @@ function Base.show(io::IO, sys::AbstractStateSpace)
...
+    T4i = ControlSystems.blockdiag(Diagonal(1 ./ sqrt.(sqrt.(Σ12))), 1.0I(nx-noc))
+
+    T = inv(T4i)*T2*T1
+
+    Pz = T*P*T'
+    Qz = inv(T')*Q*inv(T)
+    if norm(Pz[1:noc, 1:noc]-Qz[1:noc, 1:noc]) > sqrt(eps())
+        @warn("minreal: Result may be inaccurate")
     end
+    sysb = ss(T*sys.A*inv(T), T*sys.B, sys.C*inv(T), sys.D, sys.timeevol)
```
olof3, line 452:
```suggestion
    sysb = ss(T*sys.A/T, T*sys.B, sys.C/T, sys.D, sys.timeevol)
```

src/types/StateSpace.jl
```diff
@@ -391,19 +391,73 @@ function Base.show(io::IO, sys::AbstractStateSpace)
...
+    no = sys.nx-nno
+
+    T1 = diagonalizing(P, true)
+
+    Q_block = inv(T1')*Q* inv(T1)
+
+    Q11 = Q_block[1:nc, 1:nc] 
+    U1 = diagonalizing(Q11, false)
+
+    Σ12b = abs.(diag(U1*Q11*U1')) # should be block(Σ₁², 0)
```
olof3, line 438: `Σ₁²` Something wrong with the unicode? It could be that it is github that displays it poorly. I've already posted one issue regarding "hats on x's"

olof3 (commented):
Was not super easy to see what was going on. I could take a closer look after clarifications.

src/types/StateSpace.jl
olof3: I was thinking about the `svd`, must have marked the wrong rows.

src/types/StateSpace.jl
baggepinnen: inv is locally shadowed to use `pinv` with the supplied tolerances

PR #518, opened 2021-05-24
`hz` keyword in `nyquistplot` similar to in `bodeplot`

baggepinnen:


olof3:
Looks good. For future work we could consider an argument `frequnits` instead, to be able to specify, e.g, `kHz`.

PR #522, opened 2021-05-26
Maintain type for zpk in c2d

albheim:
Fixed up #444 which is based on #441

### DIFF
--- a/src/discrete.jl
+++ b/src/discrete.jl
@@ -229,7 +229,7 @@ end
 
 
 function c2d(G::TransferFunction{<:Continuous, T}, Ts, args...; kwargs...) where {T}
-    @assert issiso(G) "c2d(G::TransferFunction, h) not implemented for MIMO systems" 
+    issiso(G) || error("c2d(G::TransferFunction, h) not implemented for MIMO systems")
     sys = ss(G)
     sysd = c2d(sys, Ts, args...; kwargs...)
     # Extract the basic type (SisoRational/SisoZpk) without the parametric values
--- a/src/discrete.jl
+++ b/src/discrete.jl
@@ -227,11 +227,11 @@ function c2d_poly2poly(p,h)
 end
 
 
-function c2d(G::TransferFunction{<:Continuous, T}, h, args...; kwargs...) where {T<:SisoTf}
+function c2d(G::TransferFunction{<:Continuous, T}, h, args...; kwargs...) where {T}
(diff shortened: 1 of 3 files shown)


src/discrete.jl
```diff
@@ -228,12 +228,14 @@ function c2d_poly2poly(p, Ts)
...
-    ny, nu = size(G)
-    @assert (ny + nu == 2) "c2d(G::TransferFunction, Ts) not implemented for MIMO systems"
+function c2d(G::TransferFunction{<:Continuous, T}, Ts, args...; kwargs...) where {T}
+    issiso(G) || error("c2d(G::TransferFunction, h) not implemented for MIMO systems")
     sys = ss(G)
     sysd = c2d(sys, Ts, args...; kwargs...)
-    return convert(TransferFunction, sysd)
+    # Extract the basic type (SisoRational/SisoZpk) without the parametric values
+    # This allows the parametric type promotion to be handled by convert
+    BT = Base.typename(T).wrapper
```
baggepinnen, line 237: This approach is fragile, it relies on an internal implementation detail and can easily break, even in non-breaking releases of Julia. The correct approach here is either of 
- have different methods
-  destruct the type in the type parameters
- have a helper function that does something similar to `Base.typename(T).wrapper ` but using the type system and dispatch

baggepinnen: Example
```julia
tftype(::TransferFunction{<:Any, <:ControlSystems.SisoRational}) = ControlSystems.SisoRational
julia> tftype(tf(1))
SisoRational
```
Not too nice though as it requires one method per type. Perhaps something more clever can be deviced

albheim: Okay, will have a look at that. Different methods was discussed in the other thread as the way to go since we might want to handle the cases differently, and it seems like an easy way to start.

baggepinnen: This does it for all types using metaprogramming
```julia
for T in (ControlSystems.SisoRational, ControlSystems.SisoZpk)
    @eval tftype(::TransferFunction{<:Any, <:$T}) = $T
end
```

albheim: That also seems pretty convenient. I don't really have a preference, rewrote it with two methods before I saw this.

baggepinnen: if we want to do different things based on the type, different methods is for sure the way to go

baggepinnen (approved):

PR #523, opened 2021-05-26
Remove asserts

albheim:
Read in the [docs](https://docs.julialang.org/en/v1/base/base/#Base.@assert) that assert was suggested to only be used for debugging since it can be disabled at some optimization levels. We had some argument checks where the function is only supposed to work under some criteria which was implemented with `@assert`, had a quick go at exchanging them for `error` instead.

### DIFF
--- a/src/analysis.jl
+++ b/src/analysis.jl
@@ -75,7 +75,7 @@ Compute the determinant of the Matrix `sys` of SisoTf systems, returns a SisoTf 
 # TODO: improve this implementation, should be more efficient ones
 function det(sys::Matrix{S}) where {S<:SisoZpk}
     ny, nu = size(sys)
-    @assert ny == nu "Matrix is not square"
+    ny == nu || error("sys matrix is not square")
     if ny == 1
         return sys[1, 1]
     end
--- a/src/freqresp.jl
+++ b/src/freqresp.jl
@@ -79,7 +79,7 @@ function (sys::TransferFunction)(s)
 end
 
 function (sys::TransferFunction)(z_or_omega::Number, map_to_unit_circle::Bool)
-    @assert isdiscrete(sys) "It only makes no sense to call this function with discrete systems"
+    isdiscrete(sys) || error("It only makes no sense to call this function with discrete systems")
     if map_to_unit_circle
         isreal(z_or_omega) ? evalfr(sys,exp(im*z_or_omega.*sys.Ts)) : error("To map to the unit circle, omega should be real")
     else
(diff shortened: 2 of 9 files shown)


mfalt:
I haven't checked the difference, but maybe we should be consistent with `error(...)` and `throw(ArgumentError(...))` where I have a slight preference for the latter.
Looks good otherwise.

albheim:
> I haven't checked the difference, but maybe we should be consistent with `error(...)` and `throw(ArgumentError(...))` where I have a slight preference for the latter.
> Looks good otherwise.

Yes, was thinking that maybe I should try to throw specific arguments more. It seems we have a lot of just `error` in the toolbox, but that is no reason for why it should continue. I'll see if I can update some of them.

olof3:
Good that you are fixing this @albheim 

Should we, in general, think more about what kind of errors we throw, i.e. `DimensionMismatch`. I don't think we are doing this now? Not sure to what degree it makes sense to worry about this.

albheim:
I don't know, tried to read a bit how the julia community is reasoning about it and it seems like there are (as expected) opinions in all directions. I think the most important thing is that there is an error (you get the stacktrace) and some helpful message (you get an idea what happened and can then use stacktrace to find origin of what happened). Not all errors are easily covered by the existing types, and I'm not sure I think we should start creating our own error types just for that. But if there is an obvious standard type, why not use it.

olof3:
> I don't know, tried to read a bit how the julia community is reasoning about it and it seems like there are (as expected) opinions in all directions. I think the most important thing is that there is an error (you get the stacktrace) and some helpful message (you get an idea what happened and can then use stacktrace to find origin of what happened). Not all errors are easily covered by the existing types, and I'm not sure I think we should start creating our own error types just for that. But if there is an obvious standard type, why not use it.

Okay, sounds good, Yeah, not need to come up with new types of errors.

imciner2:
`error()` is not the same as `throw(ArgumentException())`, since `error()` will use the generic `ErrorException` instead, so when verifying arguments one of either `ArgumentError`, `DomainError`, or `DimensionMismatch` should be used instead. For instance, the case of checking that `Ts` is non-negative is best thrown as a `DomainError` instead of `ArgumentError` because negative numbers would be outside the domain of `Ts`.

albheim:
> `error()` is not the same as `throw(ArgumentException())`, since `error()` will use the generic `ErrorException` instead, so when verifying arguments one of either `ArgumentError`, `DomainError`, or `DimensionMismatch` should be used instead. For instance, the case of checking that `Ts` is non-negative is best thrown as a `DomainError` instead of `ArgumentError` because negative numbers would be outside the domain of `Ts`.

Yeah, `DomainError` seems nicer for those cases

PR #525, opened 2021-05-27
Bump version to 0.10.1

mfalt:


mfalt:
@JuliaRegistrator register

PR #529, opened 2021-05-29
Return a result structure from lsim

baggepinnen:
This is by design a non-breaking change to the lsim interface since the structure allows both getindex and destructuring to behave just like if lsim returned a tuple of arrays like before.  Indeed, the tests for lsim were not touched in this commit.

This commit also adds a plot recipe to the result structure. All plot recipes for lsimplot, stepplot, impulseplot have been replaced by the new recipe. This is a breaking change since the names of the previous plots no longer exist. A slight change is that the plots for a step response no longer show the text "step response", but I think that's an acceptable change, the user can supply any title they prefer themselves.

### DIFF
--- a/src/ControlSystems.jl
+++ b/src/ControlSystems.jl
@@ -116,6 +116,8 @@ using MacroTools
 
 abstract type AbstractSystem end
 
+
+include("types/result_types.jl")
 include("types/TimeEvolution.jl")
 ## Added interface:
 #   timeevol(Lti) -> TimeEvolution (not exported)
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -113,110 +113,39 @@ function getLogTicks(x, minmax)
 end
 
 
-@userplot Lsimplot
-
-"""
-    fig = lsimplot(sys::LTISystem, u, t; x0=0)
-    lsimplot(LTISystem[sys1, sys2...], u, t; x0)
-
-Calculate the time response of the `LTISystem`(s) to input `u`. If `x0` is
-not specified, a zero vector is used.
-"""
-lsimplot
-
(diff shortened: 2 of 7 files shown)


olof3 (commented):
Nice, seems like the right way to go.

> A slight change is that the plots for a step response no longer show the text "step response", but I think that's an acceptable change

Seems acceptable. One could possibly pass it along in the struct, it is somewhat nice to see at least. Not sure what happens if multiple response types are used in the same plot.

> together with the system that was used to produce the results.

Are you thinking about using it for interpolation, otherwise it seems more clear to only keep the information that is used, i.e. continuous/discrete (?).

I guess it would be nice to be able to use the plot recipe with data that was user generated, e.g., by some method outside of ControlSystems, i.e., not having to specify a system, but that can of course be achieved regardless.

In general, even without a system type, I guess there could be quite a bit of compilation of the recipe since there might be quite a few different types of `SimResult` that are generated. Not sure how much of a problem this might be?

> Any form of interpolation etc. can be added later if we find that useful,

Yes, I definitely think we should wait until there is a very clear need for it.

baggepinnen:
One could potentially store a metadata field in the struct that contains additional information regarding how the result was produced. 

In the simplest case, this could be a string or a symbol, e.g., `metadata = :step` to indicate that the result was produced using the function `step`. Even more advanced and flexible use could be to store a `Dict` that can contain arbitrary information, but I'm not sure what else one would like to store.

baggepinnen:
I'm more or less done with this PR. Additional metadata can be added at a later point should someone feel compelled to add it. If there are no objections, I will model the freqresp return structure similarly in a new PR when this has been merged

PR #530, opened 2021-05-31
return similarity transform from `balreal`

baggepinnen:
Because it's useful to have if other things need transformation as well, e.g., Kalman gain or covariance matrices.

Also, this PR rethrows the full error if `covar` fails, catching the error and providing much less information in it's place is not good practice.

### DIFF
--- a/src/matrix_comps.jl
+++ b/src/matrix_comps.jl
@@ -168,8 +168,9 @@ function covar(sys::AbstractStateSpace, W)
     func = iscontinuous(sys) ? lyap : dlyap
     Q = try
         func(A, B*W*B')
-    catch
-        error("No solution to the Lyapunov equation was found in covar")
+    catch e
+        @error("No solution to the Lyapunov equation was found in covar.")
+        rethrow(e)
     end
     P = C*Q*C'
     if !isdiscrete(sys)
--- a/src/matrix_comps.jl
+++ b/src/matrix_comps.jl
@@ -494,9 +494,9 @@ end
 
 
 """
-`sysr, G = balreal(sys::StateSpace)`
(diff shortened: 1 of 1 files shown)


src/matrix_comps.jl
```diff
@@ -493,9 +494,9 @@ end
 
 
 """
-`sysr, G = balreal(sys::StateSpace)`
+`sysr, G, T = balreal(sys::StateSpace)`
 
-Calculates a balanced realization of the system sys, such that the observability and reachability gramians of the balanced system are equal and diagonal `G`
+Calculates a balanced realization of the system sys, such that the observability and reachability gramians of the balanced system are equal and diagonal `G`. `T` is the similarity transform between the old state `x` and the newstate `z` such that `Tz = x`.
```
olof3, line 499:
```suggestion
Calculates a balanced realization of the system sys, such that the observability and reachability gramians of the balanced system are equal and diagonal `G`. `T` is the similarity transform between the old state `x` and the new state `z` such that `Tz = x`.
```

olof3:
Looks alright.

Some thoughts when looking through this. Perhaps change `S` to `Σ` in baltrunc for consistency. It seems like it would slightly simplify the code if just `Σ` was returned, rather than `diagm(Σ)`. An other good reason is for consistency with `svd`/`eigen`.

baggepinnen:
> Some thoughts when looking through this. Perhaps change S to Σ in baltrunc for consistency. It seems like it would slightly simplify the code if just Σ was returned, rather than diagm(Σ). An other good reason is for consistency with svd/eigen.

I have nothing agains those suggestions, but they are orthogonal to the change in this PR. I opened https://github.com/JuliaControl/ControlSystems.jl/issues/533 to track

PR #531, opened 2021-05-31
Move plot globals to runtests

albheim:
There is plotting done in more tests than `test_plots.jl`, and `test_plots` is last in the list of tests. So for the other tests that run plot commands they will not have set up the globals correctly and I think this might help with all the `GKS: GKS not in proper state. GKS must be either in the state WSAC or SGOP in routine ...` messages spamming the CI.

### DIFF
--- a/test/framework.jl
+++ b/test/framework.jl
@@ -1,3 +1,11 @@
+# Set plot globals
+ENV["PLOTS_TEST"] = "true"
+ENV["GKSwstype"] = "nul"
+
+using Plots
+gr()
+default(show=false)
+
 using ControlSystems
 # Local definition to make sure we get warnings if we use eye
 eye_(n) = Matrix{Int64}(I, n, n)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,11 +1,3 @@
-# Set plot globals
-ENV["PLOTS_TEST"] = "true"
-ENV["GKSwstype"] = "nul"
-
-using Plots
-gr()
-default(show=false)
-
 using ControlSystems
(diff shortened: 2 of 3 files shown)


baggepinnen:
It would be nice to place the initialization code in `framework.jl` with all the other init code, this way one only have to run `framework.jl` in order to manually execute other test scripts.

albheim:
How do I get it to rerun the plot generation/testing?

albheim:
@JuliaControlBot test-plots

mfalt:
I think only owners can
@JuliaControlBot test-plots

albheim:
> I think only owners can
> @JuliaControlBot test-plots
> 
> Opening and closing it should also retrigger.
> https://github.com/JuliaControl/ControlSystems.jl/blob/bde0c055a115c433f8204c74f5f8e205dd929c45/.github/workflows/PingControlExamplePlots.yml#L5
> 
> 
> This is a security measure since a change in the PR could potentially expose github secrets by changing the workflow. So it is currently limited to automatically running on first opening a PRs by any collaborators (it gets annoying of triggered on every commit) or when an admin triggers it with the comment you tried. This could probably be extended to collaborators without problem.
> https://github.com/JuliaControl/ControlSystems.jl/blob/bde0c055a115c433f8204c74f5f8e205dd929c45/.github/workflows/PRCommentResponder.yaml#L35

Yeah, guessed it might have been that. Didn't look through the CI code too carefully after I found the command, thought I might as well just try it. Not something I feel I need very often either so for me it is fine.

PR #532, opened 2021-05-31
WIP: Adaptive sampling

mfalt:
Implements functionality for adaptive sampling of functions based on the expected use. For example sampling of frequency response, knowing that it will be used in a loglog plot for magnitude and lin-log plot for phase.

All the code in `src/freqresp.jl` is temporary for now, and only showrthand for bode has been implemented as proof of concept.
I amm happy for any suggestions on the code in `src/sampler.jl`


mfalt:
There is a significant improvement in finding peaks and notches with the approach, and it should be possible to improve further to better sample when computing margins. (Plots are not updated to use the new code)
The current code computes scalings quite a few more times than nessesary, but is not significantly slower
```julia
# Tested on 1x2 StateSpace
# Old behavior
@btime freqresp(sys, exp.(LinRange(-2, 1, 425)))
  1.089 ms (10063 allocations: 1.62 MiB)

# New adaptive with no regard to mag and phase
@btime freqresp(sys, (0.01, 10)) # 425 long grid
  4.908 ms (24063 allocations: 2.96 MiB)

# New adaptive with no regard to loglog and linlog of mag and phase
@btime freqresp(sys, (0.01, 10), style=:bode) # 419 long grid
  15.405 ms (79414 allocations: 5.52 MiB)
```
I think a final version can be expected to be somewhere between the two later alternatives, closer to the first for larger systems. I defininetely think we should try to replace `default_freq_vec` with default limits and this code.

src/freqresp.jl
```diff
@@ -19,6 +19,66 @@ of system `sys` over the frequency vector `w`."""
     [evalfr(sys[i,j], s)[] for s in s_vec, i in 1:ny, j in 1:nu]
 end
 
+# TODO Most of this logic should be moved to the respective options, e.g. bode
+# TODO Less copying of code
+@autovec () function freqresp(sys::LTISystem, lims::Tuple; style=:none)
```
albheim, line 24: You return a tuple later on, so depending on which of the values should be `vec`'d you need to give the indices for those values in the tuple. No indices is if the function returns no tuple, but the whole object that should be `vec`'d.

albheim:
```suggestion
@autovec (1) function freqresp(sys::LTISystem, lims::Tuple; style=:none)
```

Would then be the way assuming the grid should not be changed.

src/sampler.jl
```diff
@@ -0,0 +1,120 @@
...
+        push_multiple!(values, rightvalue)
+        num_gridpoints += 2
+    end
+    return num_gridpoints
+end
+
+# TODO We can scale when saving instead of recomputing scaling
+# Svectors should also be faster in general
+function is_almost_linear(xleft, midpoint, xright, leftvalue, midvalue, rightvalue, xscale, yscales; reltol_dd=0.05, abstol_dd=1e-2)
+    # We assume that x2-x1 \approx  x3-x1, so we need to check that y2-y1 approx y3-y2
```
albheim, line 104:
```suggestion
    # We assume that x2-x1 \approx  x3-x2, so we need to check that y2-y1 approx y3-y2
```

Assume this is what is meant?

src/sampler.jl
```diff
@@ -0,0 +1,120 @@
...
+    #mineps in scaled version, abs to avoid assuming monotonly increasing scale
+    if (abs(xright - xleft) >= mineps) && (num_gridpoints < maxgridpoints) && !is_almost_linear(xleft, midpoint, xright, leftvalue, midvalue, rightvalue, xscale, yscales; reltol_dd, abstol_dd)
+        num_gridpoints = refine_grid!(values, grid, func, xleft,    midpoint, leftvalue, midvalue,   xscale, yscales, maxgridpoints, mineps, num_gridpoints; reltol_dd, abstol_dd)
+        num_gridpoints = refine_grid!(values, grid, func, midpoint, xright,   midvalue,  rightvalue, xscale, yscales, maxgridpoints, mineps, num_gridpoints; reltol_dd, abstol_dd)
+    else
+        # No more refinement needed, add computed points
+        push!(grid, midpoint_identity)
+        push!(grid, xscale[2](xright))
+        push_multiple!(values, midvalue)
+        push_multiple!(values, rightvalue)
```
albheim, line 95: Why create a specific method for this and not just use the one line implementation in the method?

mfalt: It used to be a weird implementation with 4 different method. Could definetely be inlined now

src/freqresp.jl
mfalt: Thanks. This is what happens when you copy code and forget to look at it again..

olof3 (commented):
Very nice, everything would be an improvement.

> I defininetely think we should try to replace default_freq_vec with default limits and this code.

Yes, it should definitely be replaced. If it is done properly, I'm not sure that we need all the scalings.


Have you seen that there are some adpative functionalities already in `PlotUtils`, it is perhaps similar to this PR (?), but I haven't looked to closely. I still think that it could be justified with a version that works well with `freqresp`. We could consider if we should use the same kwargs in that case.

```julia
julia> g = w -> abs.(freqresp(P, [w])[1])
julia> plot(adapted_grid(g, (1e4, 1e8), max_recursions=12, max_curvature=0.005), xscale=:log10, yscale=:log10, legend=false, mark=:dot)
```

mfalt:
I updated so that the grid is used for `bode, nyquist` and `sigma` now. It is currently not used in plots since we get some issues because we (currently) allow multiple systems to be submitted. Here are some illistractions. Still using the same defaults. I think we need to improve nyquist slightly. Otherwise I think it looks good (markers included to show grid). Notice that although the systems are mostly the same, the grids differ slightly.

```julia
## Test bode
sys1 = (s+1)/(s^2+ 0.1s + 1)
sys2 = 0.1*(s+1)/(s^2+ 0.1s + 100)
sys = [sys1 sys2]
mag, phase, w = bode(sys)
mag, phase = dropdims.([mag,phase],dims=1)
plot(w, mag', layout=(2,1), m=(:o,2), yscale=:log10, xscale=:log10, subplot=1, lab="", size=(800,600))
plot!(w, phase', yscale=:identity, xscale=:log10, subplot=2, m=:o, lab="")
``` 
bode

```julia
## Test Nyquist
sys1 = (s+1)/(s^2+ 0.1s + 1)
sys2 = 0.8*(s+1)/(s^2+ 0.1s + 100)
sys = [sys1 sys2]
res, ims, w = nyquist(sys)
res, ims = dropdims.([res,ims],dims=1)
plot(res', ims', m=(:o,2), lab="", size=(600,600))
``` 
nyquist

```julia
## Test Sigma
sys1 = (s+1)/(s^2+ 0.1s + 1)
sys2 = 0.1*(s+1)/(s^2+ 0.1s + 100)
sys = [sys1 sys2]
sig, w = sigma(sys)
sig = dropdims(sig,dims=1)
plot(w, sig, m=(:o,2), yscale=:log10, xscale=:log10, lab="", size=(600,600))
``` 
sigma

src/ControlSystems.jl
```diff
@@ -145,6 +145,7 @@ include("types/zpk.jl")
 include("types/lqg.jl") # QUESTION: is it really motivated to have an LQG type?
 
 include("utilities.jl")
+include("sampler.jl")
```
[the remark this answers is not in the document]
mfalt: changed to `auto_grid`

src/freqresp.jl
```diff
@@ -19,6 +19,66 @@ of system `sys` over the frequency vector `w`."""
...
+    # Create imaginary freq vector s
+    func, xscale, yscales = if iscontinuous(sys)
+        if style === :none
+            f = (w) -> (evalfr(sys, w*im),)
```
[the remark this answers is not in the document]
mfalt: I created an `eval_frequency(sys::LTISystem, w::Real)` for now. But I agree that we should do something here.

mfalt:
Some bodeplots of benchmarks provided by @olof3 . I think every single one failed for one reason or another with the old code. I compared to manually sending in 1000 logspaced gridpoints of the old code, with the default of the new code. (Manually selecting y-ranges for problem 3 and 4).

It is clear that we need some improvements, but at least the new code works (compared to the old code), and it is almost always better than manually supplying a grid. Note that the default generates quite few points (about 300).

```julia
# Single notch 
Ω = [10]
G = zpk(im*Ω, [0; im*Ω .+ 1], complex(prod(Ω)))

# Single notch, low gain
Ω = [10]
G = 1e-6*zpk(im*Ω, [0; im*Ω .+ 1], complex(prod(Ω)))

# 3 notches
Ω = [10, 100, 1000]
G = zpk(im*Ω, im*Ω .+ 1, complex(prod(Ω)))

# 1/s + 3 notches
Ω = [10, 100, 1000]
G = zpk(im*Ω, [0; im*Ω .+ 1], complex(prod(Ω)))

# FOTD
G = DemoSystems.fotd(τ=100)
```
Old 1:
bode1_old
New 1:
bode1
Old 2:
bode2_old
New 2:
bode2
Old 3:
bode3_old
New 3: (Note, only 300 sample points by default)
bode3
Old 4:
bode4_old
New 4:
bode4
New 4 with `mineps=1e-9`, i.e. about 1000 points:
bode4_extra
Old 5:
bode5_old
New 5:
bode5

src/freqresp.jl
```diff
@@ -1,22 +1,45 @@
+function eval_frequency(sys::LTISystem, w::Real)
```
baggepinnen, line 1: Does this inline? If not, we should probably encourage that for performance reasons.

src/freqresp.jl
```diff
@@ -100,50 +123,87 @@ end
...
-@autovec (1, 2) bode(sys::LTISystem) = bode(sys, _default_freq_vector(sys, Val{:bode}()))
+@autovec (1, 2) bode(sys::LTISystem; kwargs...) = bode(sys, _default_freq_lims(sys, Val{:bode}()); kwargs...)
+
+@autovec (1, 2) function bode(sys::LTISystem, lims::Tuple; kwargs...)
+    f = (w) -> begin
+        fr = eval_frequency(sys, w)
+        (abs.(fr), angle.(fr))
+    end
+    ys, grid = auto_grid(f, lims, (log10, exp10), (log10, identity); kwargs...)
+    angles = cat(ys[2]...,dims=3)
```
baggepinnen, line 139: This compiles a unique function for each length of the vector. Does it handle 10k points?

src/freqresp.jl
```diff
@@ -100,50 +123,87 @@ end
...
 @autovec (1) function sigma(sys::LTISystem, w::AbstractVector)
     resp = freqresp(sys, w)
-    sv = dropdims(mapslices(svdvals, resp, dims=(2,3)),dims=3)
+    sv = dropdims(mapslices(svdvals, resp, dims=(1,2)),dims=2)
     return sv, w
 end
-@autovec (1) sigma(sys::LTISystem) = sigma(sys, _default_freq_vector(sys, Val{:sigma}()))
+# TODO: Not tested, probably broadcast problem on svdvals in auto_grid
+@autovec (1) function sigma(sys::LTISystem, lims::Tuple; kwargs...)
+    f = (w) -> begin
```
baggepinnen, line 179:
```suggestion
    f = function (w)
```
This is a much nicer syntax to create anonymous function. Same goes in many places.

src/freqresp.jl
mfalt: Good point, will check.

src/freqresp.jl
mfalt: Does it actually? I would assume that it stops at some length. In any case, I'll compare performance to
`reduce((x,y) -> cat(x,y,dims=3), ys[2])` or some left associative equivalent.

src/freqresp.jl
```diff
@@ -1,22 +1,45 @@
...
+
+`sys_fr` has size `(ny, nu, length(w))`
+"""
+@autovec (1) function freqresp(sys::LTISystem, w_vec::AbstractVector{<:Real})
     #if isa(sys, StateSpace)
     #    sys = _preprocess_for_freqresp(sys)
     #end
-    ny,nu = noutputs(sys), ninputs(sys)
-    [evalfr(sys[i,j], s)[] for s in s_vec, i in 1:ny, j in 1:nu]
+    mapfoldl(w -> eval_frequency(sys, w), (x,y) -> cat(x,y,dims=3), w_vec)
```
baggepinnen, line 23: I have a feeling that this line is horribly slow. Madreduce and friends have a bad rep due to their naive implementation. The reduce with cat will allocate O(N2) memory?

src/freqresp.jl
baggepinnen: I think it will be hard to beat the simple
```julia
function cat3(T::AbstractVector{<:AbstractMatrix})
    N = length(T)
    traj = similar(T[1], size(T[1])..., N)
    for i = 1:N
        traj[:,:,i] = T[i]
    end
    traj
end
```

src/freqresp.jl
mfalt: I'll test how it behaves in practice, i was hoping that it was smart enough to reuse the memory. It is nice to avoid allocating matrices wherever possible, but not if it's significantly more expensive.
I'll also try with a `mapcat3` version based on your code above.

src/freqresp.jl
olof3: Is there any good reason for not just calling this `freqresp`? To me, that would make a lot more sense.

src/sampler.jl
```diff
@@ -0,0 +1,120 @@
...
+    end
+    return values, grid
+end
+
+function push_multiple!(vecs::Tuple, vals::Tuple)
+    push!.(vecs, vals)
+end
+
+# Given range (xleft, xright) potentially add more gridpoints. xleft is assumed to be already added, xright will be added
+function refine_grid!(values, grid, func, xleft, xright, leftvalue, rightvalue, xscale, yscales, maxgridpoints, mineps, num_gridpoints; reltol_dd=0.05, abstol_dd=1e-2)
```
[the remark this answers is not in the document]
olof3: You partly already do this in  `is_almost_linear`.

src/ControlSystems.jl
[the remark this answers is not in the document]
olof3: Fair enough.

olof3:
Perhaps using `max_recursions` (as in `adapted_grid` in `PlotUtils`) would be more convenient and handle tricky cases more gracefully, than `max_num_gridpoints` and keeping track of `num_gridpoints`?

src/sampler.jl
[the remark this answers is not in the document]
mfalt: I think you're right

src/freqresp.jl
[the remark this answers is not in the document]
mfalt: I am pretty sure that it makes a difference in reducing the number of points in loglog scale for example.

mfalt:
> Perhaps using `max_recursions` (as in `adapted_grid` in `PlotUtils`) would be more convenient and handle tricky cases more gracefully, than `max_num_gridpoints` and keeping track of `num_gridpoints`?

Max recursions would more or less correspond to mineps. The only difference is that mineps can be selected independently on start_gridpoints, and max recursions could be selected independently on xlims. Not sure which is most intuitive.

max_num_gridpoints essentially avoids that the code gets stuck with crazy allocations if the algorithm starts to go down to max_recursions (N) at every point (which would result in 2^N points). The typical case only goes down N steps at very few points.

src/freqresp.jl
baggepinnen: Good point, it's exactly freqresp for a scalar frequency

olof3:
> The only difference is that mineps can be selected independently on start_gridpoints, and max recursions could be selected independently on xlims. Not sure which is most intuitive.

Not sure, it does not seem trivial to specify it in a good way, e.g., `mineps = (xscale[1](lims[2])-xscale[1](lims[1]))/10000`. I wouldn't mind an unscaled version though `dw_min`, again, depends on how the resolution of features should be emphasized.

> max_num_gridpoints essentially avoids that the code gets stuck with crazy allocations if the algorithm starts to go down to max_recursions (N)

Unless the sampling is broken, that would have to be a very pathological function. To me, 2000 vs 10_000 seems like a rather small difference. Would also seem annoying if the first  part of the response is accurately sampled, and latter part is not. Regardless, the user would have to adjust `mineps` to get an acceptable plot in acceptable time.

olof3:
Yes, I think that is the case, but I don't think it would be a big difference. I think the question is what behavior that one really wants, if the resolution of features in the frequency response should be independent of frequency or not.

I am wondering if it would be feasible to only look at variations of the complex-frequency response, without really considering the frequency as is done in this PR, but otherwise using the same approach. E.g., should bubbles in the Nyquist curve be sampled in the same way regardless of their "width" (which varies the depending on the scaling? Should the three notches in the test cases be resolved similarly? If there are only a finite number of features, I would say that is desirable. However if there is a lot of crazy stuff going on at high frequencies, then perhaps not.

Regardless, we want to decide on some default strategy for just `freqresp(sys)`. What choice are you thinking about for this?
My feeling is that a sensible choice for this would work very well for all the plots that we are interested, but it  it is better to test and see.

I guess all of these decisions really depend on what features in the frequency response that people are interested in. Personally I think I would like to have more equal resolution of features. The test systems that I provided are possibly too simplistic, but capture some annoying issues at least.

src/auto_grid.jl
```diff
@@ -0,0 +1,121 @@
...
+    # x2 = xscale.(midpoint)
+    # x3 = xscale.(xright)
+    y1 = apply_tuple2(yscales, leftvalue)
+    y2 = apply_tuple2(yscales, midvalue)
+    y3 = apply_tuple2(yscales, rightvalue)
+    d1 = (y2-y1)
+    d2 = (y3-y2)
+    # Essentially low second derivative compared to derivative, so that linear approximation is good.
+    # Second argument to avoid too small steps when derivatives are small
+    norm(d1-d2) < max(reltol_dd*max(norm(d1), norm(d2)), abstol_dd)
```
olof3, line 117: Shouldn't this hold component-wise? Would probably be good with some mimo test cases.

PR #534, opened 2021-06-05
Modal form

baggepinnen:
This PR implements functionality to bring systems into modal form, similar to the function `canon(sys, 'modal')` in matlab
https://se.mathworks.com/help/control/ref/lti.canon.html

I opted to give the function it's own name, partly because I do not like to have a ladder with `if-elseif` statements in a huge function that dispatches using a string argument (like `c2d` or matlabs `canon`). 

To get to modal form, I use a new function `cdf2rdf` which has adopted its name from the matlab and scipy functions with the same name, (complex diagonal form to real diagonal form).


src/matrix_comps.jl
```diff
@@ -589,6 +589,99 @@ function similarity_transform(sys::ST, T) where ST <: AbstractStateSpace
...
+    # Implementation inspired by scipy https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.cdf2rdf.html
+    # with the licence https://github.com/scipy/scipy/blob/v1.6.3/LICENSE.txt
+    D,V = E
+    n = length(D)
+
+    function complex_indices(D::AbstractVector)
+        complex_eigs = imag.(D) .!= 0
+        findall(complex_eigs)
+    end
+
```
olof3, line 615: It is often a good case to introduce a function for better readability. In this case the implemented functionality is so simple that I think it would be more readable not to do this.

src/matrix_comps.jl
```diff
@@ -589,6 +589,99 @@ function similarity_transform(sys::ST, T) where ST <: AbstractStateSpace
...
+"""
+    Db, Vb = cdf2rdf(E::Eigen)
+
+Convert an `Eigen` factorization `(D,V)` of matrix `X`, with the properties `V*D = X*V` and `D` diagonal (complex diagonal form, cdf), to a real, block-diagonal form `Db,Vb` such that `Vb*Db = X*Vb` with `Db` having real entries with the real parts of the eigenvalues on the diagonal, and the imaginary parts of eigenvalues, if any, on the off diagonals (real diagonal form, rdf). 
+"""
+function cdf2rdf(E::Eigen)
+    # Implementation inspired by scipy https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.cdf2rdf.html
+    # with the licence https://github.com/scipy/scipy/blob/v1.6.3/LICENSE.txt
+    D,V = E
```
olof3, line 608: Personally, I prefer `\Lambda` for denoting the eigenvalues, since they are not returned in a diagonal matrix as in matlab.

src/matrix_comps.jl
```diff
@@ -589,6 +589,99 @@ function similarity_transform(sys::ST, T) where ST <: AbstractStateSpace
...
+blockdiagonalize(A::AbstractMatrix{<:Real}) = cdf2rdf(eigen(A))
+
+"""
+    Db, Vb = cdf2rdf(E::Eigen)
+
+Convert an `Eigen` factorization `(D,V)` of matrix `X`, with the properties `V*D = X*V` and `D` diagonal (complex diagonal form, cdf), to a real, block-diagonal form `Db,Vb` such that `Vb*Db = X*Vb` with `Db` having real entries with the real parts of the eigenvalues on the diagonal, and the imaginary parts of eigenvalues, if any, on the off diagonals (real diagonal form, rdf). 
+"""
+function cdf2rdf(E::Eigen)
+    # Implementation inspired by scipy https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.cdf2rdf.html
+    # with the licence https://github.com/scipy/scipy/blob/v1.6.3/LICENSE.txt
```
olof3, line 607: Not sure if this is an acceptable way to add a license, possibly okay since it is BSD3.

src/matrix_comps.jl
```diff
@@ -589,6 +589,99 @@ function similarity_transform(sys::ST, T) where ST <: AbstractStateSpace
...
+function modal_form(sys::AbstractStateSpace)
+    function complex_indices(A::Matrix) # assumes A on block diagonal form
+        findall(diag(A, -1) .!= 0)
+    end
+    A,B,C,D = ssdata(sys)
+    Ab,T = blockdiagonalize(sys.A)
+    # Calling similarity_transform looks like a detour, but this implementation allows modal_form to work with any AbstractStateSpace which implements a custom method for similarity transform
+    sysm = similarity_transform(sys, T)
+    sysm.A .= Ab # sysm.A should already be Ab after similarity_transform, but Ab has less numerical noise
+    if ControlSystems.issiso(sysm)
```
olof3, line 665:
```suggestion
    if issiso(sysm)
```
?

src/matrix_comps.jl
```diff
@@ -589,6 +589,99 @@ function similarity_transform(sys::ST, T) where ST <: AbstractStateSpace
...
+        signflips = ones(Int, sysm.nx)
+        for i in ci
+            if sysm.C[i] < 0
+                signflips[i] = -1
+                signflips[i .+ 1] = -1
+            end
+        end
+        T = diagm(signflips)
+        sysm = similarity_transform(sysm, T)
+        sysm.A .= Ab # Ab unchanged by diagonal T
```
olof3, line 678: Is this to avoid numerical errors? I guess yes, but better if it is written explicitly in the comment.

test/test_matrix_comps.jl
```diff
@@ -133,4 +133,58 @@ allpoles = [
 @test sort(pcl, by=LinearAlgebra.eigsortby) ≈ sort(allpoles, by=LinearAlgebra.eigsortby) 
 @test cont.B == K
 
+
+
+
+@testset "Modal form" begin
+    @info "Testing Modal form"
+
+    function isblockdiagonal(A)
```
olof3, line 142: Here is a case that I guess should return false?
```julia
julia> isblockdiagonal(diagm(0 => ones(5), -1 => ones(4), 1 => -ones(4)))
true
```

`isblockdiagonal` is also not great name, since it actually checks more 

```julia
julia> isblockdiagonal([ones(2,2) zeros(2,2); zeros(2,2) ones(2,2)])
false
```

src/ControlSystems.jl
```diff
@@ -39,6 +39,8 @@ export  LTISystem,
         balreal,
         baltrunc,
         similarity_transform,
+        modal_form,
+        blockdiagonalize,
```
olof3, line 43: Should we really export `blockdiagonalize`? It doesn't really have anything to do with ControlSystems, so it would probably be helpful to clarify where it is from anyway.

olof3 (commented):
This is a good function to add.

The name is fine, but what names are we thinking for other canonical forms?

If we add this function, then I say that we should remove `diagonalize`. Also, isn't the complex diagonal form what is usually meant be `modal_form`, i.e., there should be an argument (possibly `Val`) to specify when the other (real/complex) form should be used. Also, for complex `A` matrices (:smile:), it should not be possible to use the real form.

baggepinnen:
Matlab uses modal form to mean the same as in this PR
https://se.mathworks.com/help/control/ref/lti.canon.html
`modal_form` should, as you say, not work for complex systems, and it doesn't. It will fail with a method error on `blockdiagonalize`, since that requires a real-valued matrix.  

I agree that there is a slight name conflict with `diagonalize`. Perhaps `diagonalize` could be called `diagonal_form`, which is distinct from `modal_form` except for real-coefficient systems with only real poles, in which case it happens to be equivalent (possibly up to ordering, but in practice the ordering is determined by `eigen`). Similarily, if we introduce functions to convert to other canonical forms, they could be called `companion_form` or something similar.

olof3:
> Matlab uses modal form to mean the same as in this PR
https://se.mathworks.com/help/control/ref/lti.canon.html

Ah, they do actually. It is more commonly referred to as "real modal form" in the literature and the version with diagonal A matrix is just "modal form". See here for example https://see.stanford.edu/materials/lsoeldsee263/11-eig.pdf

We could of course have `real_modal_form` and `modal_form`, but a argument for specifying real canoncial form seems better to me.

baggepinnen:
https://juliacontrol.github.io/RobustAndOptimalControl.jl/dev/#Model-transformations

PR #535, opened 2021-06-09
remove incorrect warning in pzmap

baggepinnen:


albheim (commented):
Around line 772, should there not be some group --> i as there are for the other series? I'm very unfamiliar with recipes, but seemed logical to me. 

Is there some easy way to reference stuff that is outside of the default scope of the review? Github does not give the same ability to comment on lines it seems.

olof3:
> Is there some easy way to reference stuff that is outside of the default scope of the review? Github does not give the same ability to comment on lines it seems.

If you click the line you can get a permalink to that line. Would also be interested to now if there is something better.

albheim:
> > Is there some easy way to reference stuff that is outside of the default scope of the review? Github does not give the same ability to comment on lines it seems.
> 
> If you click the line you can get a permalink to that line. Would also be interested to now if there is something better.

Yeah, was thinking about doing that but the annoying thing was that I couldn't even do that in this interface. I can do that when browsing code in a repo, but here I was in the "Files Changed" for a PR and had expanded the code shown, and then I didn't get any options when clicking on the line numbers. Only thing that happened was that the line was marked yellow.

olof3:
> Yeah, was thinking about doing that but the annoying thing was that I couldn't even do that in this interface.

Ah, now I see what you mean. That is extremely annoying.

baggepinnen:
There are buttons on the top and bottom left corners of any code view that looks like arrows, those expand the code view. 

> Around line 772, should there not be some group --> i as there are for the other series? I'm very unfamiliar with recipes, but seemed logical to me.

Probably not, that series is displaying the unit circle for discrete systems so it should not be grouped with any of the systems.

albheim:
> There are buttons on the top and bottom left corners of any code view that looks like arrows, those expand the code view.

Yeah, but I don't get the same tools (the plus sign) when hovering over lines on the expanded lines.

> > Around line 772, should there not be some group --> i as there are for the other series? I'm very unfamiliar with recipes, but seemed logical to me.
> 
> Probably not, that series is displaying the unit circle for discrete systems so it should not be grouped with any of the systems.

Ahh, that seems reasonable then.

PR #536, opened 2021-06-09
change formatting in dampreport

baggepinnen:
This PR changes the formatting flag from `e` to `g` in dampreport. 
`g` makes the formatting adaptive, choosing a reasonable way to format the number based on its value. 
From wikipedia https://en.wikipedia.org/wiki/Printf_format_string

> g: double in either normal or exponential notation, whichever is more appropriate for its magnitude. g uses lower-case letters, G uses upper-case letters. This type differs slightly from fixed-point notation in that insignificant zeroes to the right of the decimal point are not included. Also, the decimal point is not included on whole numbers

```
# BEFORE
|     Pole      |   Damping     |   Frequency   |   Frequency   | Time Constant |
|               |    Ratio      |   (rad/sec)   |     (Hz)      |     (sec)     |
+---------------+---------------+---------------+---------------+---------------+
|  -1.707e+00   |  3.392e-02    |  5.032e+01    |  8.008e+00    |  5.858e-01    |
|  +5.029e+01 im|               |               |               |               |
|  -1.707e+00   |  3.392e-02    |  5.032e+01    |  8.008e+00    |  5.858e-01    |
|  -5.029e+01 im|               |               |               |               |
|  -9.518e-01   |  6.901e-03    |  1.379e+02    |  2.195e+01    |  1.051e+00    |
|  +1.379e+02 im|               |               |               |               |
|  -9.518e-01   |  6.901e-03    |  1.379e+02    |  2.195e+01    |  1.051e+00    |
|  -1.379e+02 im|               |               |               |               |

# AFTER
|     Pole      |   Damping     |   Frequency   |   Frequency   | Time Constant |
|               |    Ratio      |   (rad/sec)   |     (Hz)      |     (sec)     |
+---------------+---------------+---------------+---------------+---------------+
|  -1.707       |  0.03392      |  50.32        |  8.008        |  0.5858       |
|  +50.3      im|               |               |               |               |
|  -1.707       |  0.03392      |  50.32        |  8.008        |  0.5858       |
|  -50.3      im|               |               |               |               |
|  -0.9518      |  0.006901     |  137.9        |  21.95        |  1.051        |
|  +138       im|               |               |               |               |
|  -0.9518      |  0.006901     |  137.9        |  21.95        |  1.051        |
|  -138       im|               |               |               |               |
```


baggepinnen:
Since formatting and style are subjective, it would be nice if someone with a keen sense of style, like @mfalt , would weigh in :)

albheim:
I can say that for the example you provided I prefer the updated look at least.

olof3:
Definitely better with the updated look, for the considered system. (an extremely poorly damped system, but with time-scales on the order of seconds). I guess one should also consider the formatting for systems with much slower and faster dynamics (at least MHz or GHz). Not sure how well it will look in that case.

Anyway, here are some other suggestions
* I guess dampreport is for getting a rough idea of the system dynamics, then it seems enough with 3 significant digits.
* If possible, don't write the imaginary parts for the poles on separate lines.
* Since pretty much everyone is only working real-coefficient systems, then I guess it would be quite sensible to group conjugate pairs, e.g., `-1.707 ± 50.3im`, instead of having two lines with essentially the same information. Still, I would be very happy if single complex poles also worked :)

mfalt:
Your output is definetely easier to read, and if I understand it correctly will make no difference compared to before when the exponents are large/small. So I have no objections. But I have never used dampreport either :)

baggepinnen:
> Definitely better with the updated look, at least for the considered system. (an extremely poorly damped system, but with time-scales on the order of seconds). I guess one should also consider the formatting for systems that both has dynamics on this time scale as well as dynamics that is much faster / slower.
> 
> Anyway, here are some other suggestions
> 
> * I guess dampreport is for getting a rough idea of the system dynamics, then it seems enough with 3 significant digits.
> * If possible, don't write the imaginary parts for the poles on separate lines.
> * Since pretty much everyone is only working real-coefficient systems, then I guess it would be quite sensible to group conjugate pairs, e.g., `-1.707 ± 50.3im`, instead of having two lines with essentially the same information. Still, I would be very happy if single complex poles also worked :)

Those are all good suggestions, especially the pm operator would be nice to use for real-coefficient systems.

I was actually wondering about the time constants in this example, is the way we're calculating them always valid no matter the total order of the system?

olof3:
> I was actually wondering about the time constants in this example, is the way we're calculating them always valid no matter the total order of the system?

As the "time constants of the poles" it seems to be alright, I think.

I guess there is the concept of average residence time (see e.g. Advanced PID Control) which is primarily used for stable not-very-resonant industrial processes, I guess that could be useful to print out as well, but not sure to what extent to do it for resonant systems, etc.

baggepinnen:
I made some further updates, we now have three cases, complex coeff systems is the most general, it prints all poles. Real coeff system with complex poles print using the ± operator, and real poles print using no  ± operator. The formatting for complex-coeff systems is perhaps not super optimal, but I got quite tired of trying and it should be better than before
```julia
julia> dampreport(ex_4)
|        Pole        |   Damping     |   Frequency   |   Frequency   | Time Constant |
|                    |    Ratio      |   (rad/sec)   |     (Hz)      |     (sec)     |
+--------------------+---------------+---------------+---------------+---------------+
| +0                 |  -1           |  0            |  0            |  -Inf         |
| -0.0597 ± 0.0171im |  0.961        |  0.0621       |  0.00989      |  16.7         |
| -0.0858            |  1            |  0.0858       |  0.0137       |  11.7         |
| -0.18              |  1            |  0.18         |  0.0287       |  5.55         |

julia> dampreport(1/(s+1+2im)/(s+2+3im))
|        Pole        |   Damping     |   Frequency   |   Frequency   | Time Constant |
|                    |    Ratio      |   (rad/sec)   |     (Hz)      |     (sec)     |
+--------------------+---------------+---------------+---------------+---------------+
| -1            -2im |  0.447        |  2.24         |  0.356        |  1            |
| -2            -3im |  0.555        |  3.61         |  0.574        |  0.5          |

julia> dampreport(sys)
|        Pole        |   Damping     |   Frequency   |   Frequency   | Time Constant |
|                    |    Ratio      |   (rad/sec)   |     (Hz)      |     (sec)     |
+--------------------+---------------+---------------+---------------+---------------+
| -1                 |  1            |  1            |  0.159        |  1            |
| +4                 |  -1           |  4            |  0.637        |  -0.25        |
| -4                 |  1            |  4            |  0.637        |  0.25         |
```

olof3:
Great improvements, it looks pretty good now.

One could consider switching the columns of frequency (rad/sec) and the damping ratio, but it is also nice that the frequency information is kept together.

PR #537, opened 2021-06-09
support hz keyword in sigmaplot similar to bodeplot

baggepinnen:


src/plotting.jl
```diff
@@ -739,9 +742,6 @@ Create a pole-zero map of the `LTISystem`(s) in figure `fig`, `args` and `kwargs
 pzmap
 @recipe function pzmap(p::Pzmap)
     systems = p.args[1]
-    if systems[1].nu + systems[1].ny > 2
-        @warn("pzmap currently only supports SISO systems. Only transfer function from u₁ to y₁ will be shown")
-    end
```
albheim, line 744: This was in another PR by itself. Guess it will be fine, probably just maybe merge the other one first?

PR #538, opened 2021-06-12
add additive identity element for statespace and TF

baggepinnen:


mfalt:
This is a bit problematic as implemented because dimensions are not the same.
```julia
sys = ss(rand(2,2))
zero(sys) + sys
```
will probably give dimension mismatch? So i think

baggepinnen:
You're right, that's a bit unfortunate. I removed the `zero` from type and added some tests for MIMO for zero from element.

PR #539, opened 2021-06-13
bugfix for negative real pole in damp

baggepinnen:


olof3:
Not sure if this is where we want to fix it? I think it would seem quite reasonable, and avoid these kinds of problems, if `pole`  always would return complex numbers?

There is somewhat of an inconsistency in the current behavior.

```julia
julia> G = tf(1, [1, 1])
TransferFunction{Continuous, ControlSystems.SisoRational{Int64}}
  1
-----
s + 1

julia> pole(G)
1-element Vector{ComplexF64}:
 -1.0 + 0.0im

julia> pole(ss(G))
1-element Vector{Float64}:
 -1.0
```

Also, this seems like a case where a simple regression test could become useful.

baggepinnen:
> Not sure if this is where we want to fix it? I think it would seem quite reasonable, and avoid these kinds of problems, if pole always would return complex numbers?

You're right in that we should probably always return the same type from `pole` for standard systems. Someone might however implement a special `RealPoleSystem` that always return a  real number, for which this function still needs to work.

PR #540, opened 2021-06-16
Fix struct_ctrb_obsv

### LINKED ISSUE #409: Issues with struct_ctrb_obsv and sminreal
I find that `struct_ctrb_obsv` has a tendency to return zeros a bit too often, for instance in the following simple example. 
```julia
julia> using ControlSystems.DemoSystems: resonant

julia> sminreal(resonant()*resonant())
StateSpace{Continuous, Float64, Matrix{Float64}}
D = 
 0.0

Continuous-time state-space model
```
The code has not been touched for a very long time and seems to be written by @jcrist 6 years ago, do you have any recollection of what it's supposed to do?

This function is used by `zpkdata(::StateSpace)` which in turn is used in, e.g., `pzmap`.

baggepinnen:
I didn't quite understand the old implementation of `struct_ctrb_obsv` so I reimplemented it completely. This version passes the tests, and the additional test that was previously failing from #409. 

The algorithm maintains a bitvector with state indices that have been affected by the input. It applies the `A` matrix `nx` times, corresponding roughly to calculating the last block of the controllability matrix. After this, the bit vector will indicate the structurally controllable states.

closes #409

### DIFF
--- a/src/simplification.jl
+++ b/src/simplification.jl
@@ -26,11 +26,9 @@ end
 # structurally controllable.
 function struct_ctrb_states(A::AbstractVecOrMat, B::AbstractVecOrMat)
     bitA = A .!= 0
-    d_cvec = cvec = vec(any(B .!= 0, dims=2))
-    while any(d_cvec .!= 0)
-        Adcvec = vec(any(bitA[:, findall(d_cvec)], dims=2))
-        cvec = cvec .| Adcvec
-        d_cvec = Adcvec .& .~cvec
+    x = vec(any(B .!= 0, dims=2)) # indexs vector indicating states that have been affected by input
+    for i = 1:size(A, 1) # apply A nx times, similar to controllability matrix
+        x = (bitA * x) .!= 0
     end
-    return cvec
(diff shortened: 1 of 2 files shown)


src/simplification.jl
```diff
@@ -26,11 +26,9 @@ end
 # structurally controllable.
 function struct_ctrb_states(A::AbstractVecOrMat, B::AbstractVecOrMat)
     bitA = A .!= 0
-    d_cvec = cvec = vec(any(B .!= 0, dims=2))
-    while any(d_cvec .!= 0)
-        Adcvec = vec(any(bitA[:, findall(d_cvec)], dims=2))
-        cvec = cvec .| Adcvec
-        d_cvec = Adcvec .& .~cvec
+    x = vec(any(B .!= 0, dims=2)) # indexs vector indicating states that have been affected by input
```
baggepinnen, line 29:
```suggestion
    x = vec(any(B .!= 0, dims=2)) # index vector indicating states that have been affected by input
```

PR #541, opened 2021-06-18
Yet another fix for  struct_ctrb_states

### LINKED ISSUE #475: Numerical stability issue with ss->zpk conversion?
ControlSystems.jl v0.9.2

### SS-system/odd result:
I have the following system:
```julia
A=[
 0.0  0.0   0.0;
 1.0  0.0  -0.00578297;
 0.0  1.0   0.0;
]
B=[1.0; 0.0; 0.0]
C=[0.0 0.0 1.0]
D=[0]

ss_sys=ss(A,B,C,D)
```

If I call `zpk()`, I get strange results:
```julia-repl
julia> zpk(ss_sys)

TransferFunction{Continuous, ControlSystems.SisoZpk{Float64, ComplexF64}}
   1.0
0.0---
   1.0
```

### Why is this odd?/correcting for numerical noise
If I look at the poly-rational view of the system, I get:

```julia-repl
tf(ss_sys)

TransferFunction{Continuous, ControlSystems.SisoRational{Float64}}
7.771561172376096e-16s^2 - 1.2776238400569184e-15s + 1.0000000000000007
-----------------------------------------------------------------------
                    1.0s^3 + 0.005782969999999999s
```
Which indicates the numerator is actually a constant of 1 with numerical noise "generating" false terms for `s^2` & `s^1`.

### Manual cleanup of numerical noise
If I create a cleaned-up version of this poly-rational view, I can generate a `zpk` view that generates useful output in my simulation:

```julia-repl
tf_sys = tf([1], [1, 0, 0.0057829713287632966, 0])
zpk(tf_sys)

TransferFunction{Continuous, ControlSystems.SisoZpk{Float64, ComplexF64}}
                        1.0
1.0---------------------------------------------
   (1.0s^2 + 0.0057829713287632966)(1.0s + -0.0)
```

### Further analysis
This time, if I manually "dirty-up" the poly-rational function, I can see where things go foul:

```julia-repl
tf_debug = tf([1e-15, -1e-15, 1], [1, 0, 0.0057829713287632966, 0])
zpk(tf_debug)

TransferFunction{Continuous, ControlSystems.SisoZpk{Float64, ComplexF64}}
           1.0s^2 - 1.0s + 9.999999999999999e14
1.0e-15---------------------------------------------
       (1.0s^2 + 0.0057829713287632966)(1.0s + -0.0)
```

As you can see, the poly-rational form wants to normalize the gain factor to ensure the most significant term of the polynomial (`s^2` in this case) has a coefficient of `1`.

In other words, numerical error is pushing the transfer function to a value of `H(s)=0`!!!

baggepinnen:
Closes #475

### DIFF
--- a/src/simplification.jl
+++ b/src/simplification.jl
@@ -42,7 +42,7 @@ function struct_ctrb_states(A::AbstractVecOrMat, B::AbstractVecOrMat)
     bitA = A .!= 0
     x = vec(any(B .!= 0, dims=2)) # index vector indicating states that have been affected by input
     for i = 1:size(A, 1) # apply A nx times, similar to controllability matrix
-        x = (bitA * x) .!= 0
+        x = x .| (bitA * x) .!= 0
     end
     x
 end
--- a/test/test_simplification.jl
+++ b/test/test_simplification.jl
@@ -12,6 +12,20 @@ using ControlSystems.DemoSystems: resonant
 R = resonant()*resonant()
 @test sminreal(R) == R # https://github.com/JuliaControl/ControlSystems.jl/issues/409
 
+# https://github.com/JuliaControl/ControlSystems.jl/issues/475
+A = [
+ 0.0  0.0   0.0;
+ 1.0  0.0  -0.00578297;
+ 0.0  1.0   0.0;
+]
+B = [1.0; 0.0; 0.0]
+C = [0.0 0.0 1.0]
+D = [0]
+
+ss_sys = ss(A,B,C,D)
+@test sminreal(ss_sys) == ss_sys
+
+
 ## MINREAL ##
 
 s = tf("s")


src/simplification.jl
```diff
@@ -42,7 +42,7 @@ function struct_ctrb_states(A::AbstractVecOrMat, B::AbstractVecOrMat)
     bitA = A .!= 0
     x = vec(any(B .!= 0, dims=2)) # index vector indicating states that have been affected by input
     for i = 1:size(A, 1) # apply A nx times, similar to controllability matrix
-        x = (bitA * x) .!= 0
+        x = x .| (bitA * x) .!= 0
```
albheim, line 45: Is there a reason this is not `x .= x .| (bitA...` or `x .|= (bitA...`? Wouldn't this do the same thing but not have to reallocate 
 `x` every loop? Might be a very minor increased cost so maybe does not really matter.

baggepinnen: You're right, we could save some allocations here. A `BitArray` of size `x` is a very minor allocation compared to the arrays allocations that result from the order reduction though :P

olof3: Perhaps even along the lines of
```suggestion
for k = 1:size(A, 1)
        for i in 1:size(A,1)
            x[i] = any(bitA[i,j] & x[j] for j=1:n)
        end
```
seemed to be quite a bit faster for some medium sized example that I tried.

PR #548, opened 2021-08-21
optional epsilon in dcgain

baggepinnen:


olof3:
Hmm, what is the use case for this? wouldn't be more clear and just as easy to just call  `evalfr(sys, -ϵ)` directly?

If the point is that you want analogous behavior between continuous-time and discrete-time systems then you would probably want to have `exp(-ϵ*G.Ts)?  for discrete-time systems?

Also, I guess `ϵ` is not quite a frequency as stated in the docs.

```
equal to G(0) for continuous-time systems and G(1) for discrete-time systems.
`ϵ` can be provided to evaluate the dcgain at a small
non-zero frequency.
```

baggepinnen:
The use case is when numerical problems cause the dcgain fail to evaluate in zero. If you call `evalfr` you need to manually adjust for discrete vs. cont. system. It's probably a good idea to use `exp(-ϵ*G.Ts)` for discrete systems though.

olof3:
Looks good now. Just curious about what numerical problems, some imperfect pole-zero cancellation at the origin?

baggepinnen:
> Just curious about what the numerical problems are, some imperfect pole-zero cancellation at the origin?

Yes, they appear very often when dealing with systems where you measure accelerations, causing you do get a non-zero `dcgain`

PR #550, opened 2021-09-29
bugfix and test perturbed dcgain

baggepinnen:


baggepinnen:
Hmm, it looks like I need to rebase something here.

PR #551, opened 2021-09-29
bugfix dcgain

baggepinnen:


baggepinnen:
The error on julia v1.6 is the same as previously encountered and related to printout testing

PR #555, opened 2021-10-05
remove version checks

albheim:
Remake of #356, removes the static version checks since we only support 1.6+ now.

### DIFF
--- a/src/types/DelayLtiSystem.jl
+++ b/src/types/DelayLtiSystem.jl
@@ -160,7 +160,7 @@ function Base.show(io::IO, sys::DelayLtiSystem)
     print(io, "\nP: ")
     show(io, sys.P.P)
 
-    println(io, "\n\nDelays: $(sys.Tau)")
+    print(io, "\n\nDelays: $(sys.Tau)")
 end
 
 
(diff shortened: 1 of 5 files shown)


albheim:
Seems [this](https://github.com/JuliaControl/ControlSystems.jl/commit/93feb05672d0e05e46eb76e509f06d781b92191c) added an extra print to the state space, which broke CI. Easy to fix, either just revert the print or fix 5 newlines in the tests.

@baggepinnen was there a reason to introduce those extra newlines?

baggepinnen:
The reason was that vectors of systems printed slightly weirdly without the change. Not a big deal though. I wonder why test pass on some Julia versions but not others.

baggepinnen (approved):

albheim:
The github CI shows pass on nightly even though it failed, since we have set it as allowed to fail (if that is what you were talking about).

baggepinnen:
Ah, it's perhaps not great that it signals green when the tests fail. I think travis would show that allowed subtests failed but report success on the overall test. 
Eventually we should perhaps look over our print tests since they have historically broke several times and it's a hassle to update them every second julia release.

albheim:
> Ah, it's perhaps not great that it signals green when the tests fail. I think travis would show that allowed subtests failed but report success on the overall test. 

Completely agree, remember this as something that was missing from github ci but maybe I'm wrong there.

PR #572, opened 2021-12-05
Fb/delaysimfix

baggepinnen:


baggepinnen:
The failures appear unrelated, probably a change in some linear algebra code in julia v1.7

PR #575, opened 2021-12-15
add diskmargin for SISO systems

baggepinnen:
Simultaneous MIMO requires structural singular value computation, but one loop at the time MIMO should be doable with the stuff present in this PR

Reference: https://arxiv.org/abs/2003.04771

### DIFF
--- a/src/ControlSystems.jl
+++ b/src/ControlSystems.jl
@@ -53,6 +53,7 @@ export  LTISystem,
         markovparam,
         margin,
         delaymargin,
+        diskmargin,
         gangoffour,
         # Connections
         append,
--- a/src/ControlSystems.jl
+++ b/src/ControlSystems.jl
@@ -54,6 +54,8 @@ export  LTISystem,
         margin,
         delaymargin,
         diskmargin,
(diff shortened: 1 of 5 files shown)


olof3:
I only skimmed the article and the code, but my feeling is that this would be a very good fit for a dedicated package under JuliaControl. Not sure what I think about it in ControlSystems.jl.

baggepinnen:
Matlab recommends diskmargin in many of their modern tutorials. It makes more sense to me to consider the diskmargin than to consider gain and phase margins since they tend to be optimistic. Here are some further references
- https://se.mathworks.com/help/robust/ug/stability-analysis-using-disk-margins.html
- https://se.mathworks.com/help/robust/ug/robust-controller-for-spinning-satellite.html
- https://www.youtube.com/watch?v=XazdN6eZF80
- https://www.youtube.com/watch?v=sac_IYBjcq0

baggepinnen:
See https://juliacontrol.github.io/RobustAndOptimalControl.jl/dev/ for diskmargin functionality

PR #576, opened 2021-12-16
add relative gain array

baggepinnen:
Both a function to compute and a plot

### DIFF
--- a/src/ControlSystems.jl
+++ b/src/ControlSystems.jl
@@ -54,6 +54,7 @@ export  LTISystem,
         margin,
         delaymargin,
         gangoffour,
+        relative_gain_array,
         # Connections
         append,
         series,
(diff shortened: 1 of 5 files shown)


olof3:
Good that you added freqresp for a single frequency!

I'm haven't used RGAs that much, but here are some preliminary comments:

* Should you really take the real of the computed matrices G .* pinv(G)? Didn't see this mentioned by Skogestad, but only skimmed it very quickly. Or is it that you want to make the matrix exactly real for `w=0`
* Couldn't it just be `rga(G, w::Number) = rga(G, [w])[1, :, :]` for the single-frequency case? seems like a lot of code duplication.
* Given that most people use RGA quite rarely, and many are happily unaware of it, I think I would prefer to name it  `relative_gain_array` for clarity. Although it is a bit long, I think it is better than adding a three letter acronym.

baggepinnen:
You're right, it shouldn't be real. Very often you'd like to check `abs` (not real), but it's better to just leave the elements be since it's  easy to manually abs them. It does seem that the UC inverse is recommended, I removed the `pinv` and made a note that it only supports square systems with a link to the reference for singular systems.

baggepinnen:
Actually, the paper was nice enough to provide some code for the UC inverse so it's now implemented

PR #593, opened 2022-01-06
Add headings for ploting examples

ErikQQY:
The plotting examples docs does't look so good🤪

I think separate them with headings would look better

plot


baggepinnen:
Thanks!

PR #596, opened 2022-01-07
make feedback differentiable by avoiding try-catch

baggepinnen:
Zygote.jl does not handle try-catch, so I modified `feedback` slightly to avoid using it, but maintained the useful error messagees.

The comment `# inv seems to be better than lu` made no sense to me since `inv` is literally implemented like this
```julia
Ai = inv!(lu(AA))
```

### DIFF
--- a/src/connections.jl
+++ b/src/connections.jl
@@ -286,28 +286,23 @@ See Zhou, Doyle, Glover (1996) for similar (somewhat less symmetric) formulas.
         D = [s1_D11 + α*s1_D12*s2_D22*s1_D21        α*s1_D12*s2_D21;
                       s2_D12*s1_D21           s2_D11 + α*s2_D12*s1_D22*s2_D21]
     else
-        # inv seems to be better than lu
-        R1 = try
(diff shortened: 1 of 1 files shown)


baggepinnen:
Let's put his one on hold, Zygote doesn't actually support `@warn` either :/

baggepinnen:
Note to self: `hinfnorm` fails with both ForwardDiff and ReverseDiff due to GLA's HessenbergFactorization having different fieldnames. ReverseDiff also falis to use the chain rule for `hinfnorm` due to the `TrackedReal` being a subtype of an outer type. and thus RD is failing to pass "tracker free" variables into `hinfnorm`

```julia
typeof(sys) = ControlSystems.StateSpace{Continuous, ReverseDiff.TrackedReal{Float64, Float64, Nothing}}
ERROR: LoadError: type HessenbergFactorization has no field H

typeof(sys) = ControlSystems.StateSpace{Continuous, ForwardDiff.Dual{ForwardDiff.Tag{typeof(cost), Float64}, Float64, 4}}
ERROR: LoadError: type HessenbergFactorization has no field H
```

PR #604, opened 2022-01-15
Add tfdata function

ErikQQY:
In Matlab, there is ```tfdata``` to get the numerator and denumerator of a transfer function, so I add the same ```tfdata``` to ControlSystems.jl. If it is ok, I will add tests and docs for this method.

### DIFF
--- a/src/ControlSystems.jl
+++ b/src/ControlSystems.jl
@@ -95,7 +95,8 @@ export  LTISystem,
         denpoly,
         iscontinuous,
         isdiscrete,
-        ssdata
+        ssdata,
+        tfdata
 
 
 # QUESTION: are these used? LaTeXStrings, Requires, IterTools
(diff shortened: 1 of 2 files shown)


baggepinnen:
Hello and thanks for the PR. We already have the function `numvec, denvec` which does this. Also, the function you added would give the wrong result for MIMO systems.

ErikQQY:
Oops, just saw the ```numvec``` and ```denvec``` methods, but I notice these return something like:

```julia
julia> C_111 = tf([1, 2], [1, 5]);
julia> numvec(C_111)
1×1 Matrix{Vector{Int64}}:
 [1, 2]
```

I think it would be better to return a ```Vector``` instead of ```Matrix{Vector{Int64}}```, so we can index more easier?🤔

baggepinnen:
The reason they return a matrix is because they support MIMO transfer functions

ErikQQY:
Thanks a lot! I would use ```numvec``` and ```denvec``` then.

PR #605, opened 2022-01-21
add plotphase kwarg to marginplot

baggepinnen:
and fix order for mimo systems


baggepinnen:
@albheim did you experiment further with this? The plot that is broken above works fine locally, and it's broken also on another PR, so I think there's something up with Plots on CI at the moment

albheim:
Tested locally just now, and I get the mag plot but not the mag margin text. So I'm somewhere between the two plots the CI is showing above.

Added another tf (`s/(s^4-s-4)`) in the list used for the margin plot test, and now it only shows the gain margin but not the phase margin.
Changed the above to `2s/(s^4-s-4)` and now both gain and phase margin shows.

Something seems to be a bit strange with the titles, and I see that for both titles you have put parenthesis around them. Not sure how that is supposed to affect it, but could it be from that?

In current version it seems like the ones with my added tf errors, and with only the two from example plots it looks like above with both gain and phase showing.

baggepinnen:
@JuliaControlBot test-plots

PR #622, opened 2022-01-28
Fix some problems for comparisons between tfs

albheim:
The tests added here would have returned incorrectly before, thanks to incorrect indexing into the polynomials.

Fix indexing, add tests for the previous error and fix new test error that was previously hidden by allowing approximate comparison between polynomials of different sizes.

### DIFF
--- a/src/types/SisoTfTypes/SisoRational.jl
+++ b/src/types/SisoTfTypes/SisoRational.jl
@@ -101,9 +101,16 @@ function isapprox(f1::SisoRational, f2::SisoRational; rtol::Real=sqrt(eps()), at
     # Get representation of num/den so index access is correct
     f1num, f1den = numvec(f1), denvec(f1)
     f2num, f2den = numvec(f2), denvec(f2)
-    _prefill_zero(x, n) = [zeros(n - length(x)); x]
-    f1num, f2num = _prefill_zero.((f1num, f2num), maximum(length.((f1num, f2num))))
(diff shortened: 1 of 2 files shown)


src/types/SisoTfTypes/SisoRational.jl
```diff
@@ -89,11 +89,22 @@ function evalfr(f::SisoRational{T}, s::Number) where T
...
 
 # We might want to consider alowing scaled num and den as equal
 function isapprox(f1::SisoRational, f2::SisoRational; rtol::Real=sqrt(eps()), atol::Real=0)
-    isapprox(f1.num * f2.den[1], f2.num * f1.den[1], rtol=rtol, atol=atol) && isapprox(f1.den * f2.den[1], f2.den * f1.den[1], rtol=rtol, atol=atol)
+    # Get representation of num/den so index access is correct
+    f1num, f1den = numvec(f1), denvec(f1)
+    f2num, f2den = numvec(f2), denvec(f2)
+    _prefill_zero(x, n) = [zeros(n - length(x)); x]
+    f1num, f2num = _prefill_zero.((f1num, f2num), maximum(length.((f1num, f2num))))
+    f1den, f2den = _prefill_zero.((f1den, f2den), maximum(length.((f1den, f2den))))
```
albheim, line 106: Don't think this is the best way to do it, here is yet another allocation. But the allocations can, if not fixed, at least be reduced a little by just implementing it a bit better. Mostly thought I would check if anyone had some insights into if the method seems reasonable and would return correct things, to me it feels right.

PR #624, opened 2022-01-28
remove `ny,nu` fields from TransferFunction

### LINKED ISSUE #399: Add constructors with nu, ny, nx for easy rebuilding

baggepinnen:
Closes #399

### DIFF
--- a/src/types/Lti.jl
+++ b/src/types/Lti.jl
@@ -72,7 +72,7 @@ function Base.getproperty(sys::LTISystem, s::Symbol)
 end
 
 Base.propertynames(sys::LTISystem, private::Bool=false) =
-    (fieldnames(typeof(sys))..., (isdiscrete(sys) ? (:Ts,) : ())...)
+    (fieldnames(typeof(sys))..., :nu, :ny, (isdiscrete(sys) ? (:Ts,) : ())...)
 
 
 
--- a/src/types/PartionedStateSpace.jl
+++ b/src/types/PartionedStateSpace.jl
@@ -56,6 +56,10 @@ function getproperty(sys::PartionedStateSpace, d::Symbol)
     end
 end
 
+function Base.propertynames(s::PartionedStateSpace, private::Bool=false)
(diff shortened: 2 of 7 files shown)


albheim (approved):

albheim:
Realised I approved without checking if it passed... But the changes seem reasonable to me so as long as the expected tests pass I think it is all good :)

PR #625, opened 2022-01-28
fix type errors in ss

### LINKED ISSUE #299: Conversion problem for ss(1,1,1,0)/3
```julia
ss(1,1,1,0)/3
```
throws `InexactError: Int64(0.3333333333333333)`

Just a reminder to fix this once work starts on the promotion rules, etc.

baggepinnen:
the division might change the type, causing the `ST` constructor to fail.

Closes #299

### DIFF
--- a/src/types/StateSpace.jl
+++ b/src/types/StateSpace.jl
@@ -338,7 +338,7 @@ function /(n::Number, sys::ST) where ST <: AbstractStateSpace
 end
 
 Base.inv(sys::AbstractStateSpace) = 1/sys
-/(sys::ST, n::Number) where ST <: AbstractStateSpace = ST(sys.A, sys.B, sys.C/n, sys.D/n, sys.timeevol)
+/(sys::ST, n::Number) where ST <: AbstractStateSpace = ss(sys.A, sys.B, sys.C/n, sys.D/n, sys.timeevol)
 
 
 #####################################################################
--- a/src/types/StateSpace.jl
+++ b/src/types/StateSpace.jl
@@ -223,15 +223,16 @@ function isapprox(sys1::ST1, sys2::ST2; kwargs...) where {ST1<:AbstractStateSpac
 end
 
 ## ADDITION ##
-Base.zero(sys::AbstractStateSpace) = ss(zero(sys.D), sys.timeevol)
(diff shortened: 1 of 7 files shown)


src/types/StateSpace.jl
```diff
@@ -338,7 +338,7 @@ function /(n::Number, sys::ST) where ST <: AbstractStateSpace
 end
 
 Base.inv(sys::AbstractStateSpace) = 1/sys
-/(sys::ST, n::Number) where ST <: AbstractStateSpace = ST(sys.A, sys.B, sys.C/n, sys.D/n, sys.timeevol)
+/(sys::ST, n::Number) where ST <: AbstractStateSpace = ss(sys.A, sys.B, sys.C/n, sys.D/n, sys.timeevol)
```
albheim, line 341: Does this work with heterostatespace then?

baggepinnen: nope, hss needs it's own overload. For `Int` it didn't work before either since hss is even stricter with the types, but this actually breaks hss for float types :/, good catch. It's a bit unfortunate that we can't use `ST` here, perhaps something like `constructor(ST)` that strips the type parameters would be nice to have to make it generic

albheim: > perhaps something like constructor(ST) that strips the type parameters

Is this easy to do? Remember I tried to do something similar with maybe c2d for tf, and we wanted to return a tf of same type as in zpk/rational but the number types could change and timeevol would change. I think I ended up just dispatching differently because I only found some internal methods to access that data which didn't seems stable, but might have changed (or I might not have found the best method).

baggepinnen: This
```julia
julia> tG = typeof(G)
TransferFunction{Continuous, ControlSystems.SisoRational{Int64}}

julia> tG.name.wrapper
TransferFunction
```
is probably one of the most used internals of julia since there is no "supported" way of doing it, but it's often very useful to do. The alternative would be that every type of `LTISystem` implements a method like
```
constructor(::StateSpace) = StateSpace
```
(without the type parameters). This also works, but it's a bit unfortunate that it's required. There's an issue about exposing `.name.wrapper` as an API and the consensus appears to be that it should be done, but no one has done it yet.
https://github.com/JuliaLang/julia/issues/35543
I think we could introduce our own `basetype` function until one exists in Base.

albheim (approved):
LGTM

PR #629, opened 2022-01-31
use timeevol types to dispatch matrix equations

baggepinnen:
This PR "soft deprecates" the matrix-equation solver interfaces that distinguish between continuous/discrete time by means of a prefix `c/d` in favor of function that dispatch on the `TimeEvol` type.

```julia
lqr(...)   -> lqr(Continuous, ...) or lqr(dsys, ...)
dlqr(...)  -> lqr(Discrete, ...)   or lqr(csys, ...)
dlyap(...) -> lyap(Discrete, ...)
```
and so on. This avoids the previously common
```julia
if isdiscrete
    ared
else
    arec
end
```
Instead, we may now use
```julia
are(te, ...)
```
where `te` is the `sys.timeevol` field (type or instance, both work).

I say "soft deprecate" since the old methods are still there, their use is just discouraged in the docstring.

Supersedes https://github.com/JuliaControl/ControlSystems.jl/pull/563


docs/src/examples/example.md
```diff
@@ -21,7 +21,7 @@ C       = [1 0]
 sys     = ss(A,B,C,0, Ts)
 Q       = I
 R       = I
-L       = dlqr(A,B,Q,R) # lqr(sys,Q,R) can also be used
+L       = lqr(Discrete,A,B,Q,R) # lqr(sys,Q,R) can also be used
```
albheim, line 24: I haven't really followed the discussion about these changes, but to me it feels slightly strange with `Discrete` as first argument. But I guess you always have to supply an argument, we don't want to make one default?

The other one where you send in a system looks nice, so maybe it makes sense that the system information comes in with the leftmost argument in some type of consistency with that.

Nothing I have strong feelings about, just reacted to it and unsure if there are good alternatives (I don't have any that I think are really better)

src/matrix_comps.jl
```diff
@@ -1,29 +1,33 @@
...
+    are(::Continuous, A, B, Q, R)
 
 Compute 'X', the solution to the continuous-time algebraic Riccati equation,
 defined as A'X + XA - (XB)R^-1(B'X) + Q = 0, where R is non-singular.
 
 Uses `MatrixEquations.arec`.
+This function exists for legacy reasons, users are encouraged to use the interface `are(Continuous, A, B, Q, R)` instead.
 """
-function care(A, B, Q, R)
+function are(::ContinuousType, A::AbstractMatrix, B, Q, R)
```
albheim, line 10: Why type on `A` but not the others?

albheim: I assume there is a reason since I see it is done on a few more spots, just curios and want to understand :)

baggepinnen: If I do not type the `A`, there is a method ambiguity when `A isa Number` between this method and the method that takes `are(t::TimeEvolType, A::Number, ...)`, since `TimeEvolType` is less strict than `ContinuousType`.

docs/src/examples/example.md
baggepinnen: Having one as default is part of what we're trying to get away from, since `lqr(sys)` and `lqr(A, B)` then not mean the same if the system is discrete. The idea here was to use the name `lqr` rather than `lqr/lqrd` and at the same time allow for dispatch on `sys.timeevol` rather than having manual `if isdiscrete(sys)` everywhere.

albheim: Yeah, makes sense.

PR #630, opened 2022-02-07
make numbers and matrices work as systems for `freqresp`

baggepinnen:
this is convenient in many places where functions that take either systems of gain matrices.

### DIFF
--- a/src/freqresp.jl
+++ b/src/freqresp.jl
@@ -8,6 +8,8 @@ function freqresp(sys::LTISystem, w::Real)
     evalfr(sys, s)
 end
 
+freqresp(G::Union{UniformScaling, AbstractMatrix, Number}, w::Real) = G
+
 """
     sys_fr = freqresp(sys, w)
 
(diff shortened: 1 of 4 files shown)


albheim (approved):
Looks good, also noted that the docstring for `freqresp` had `w -> ... iw*im ...` which I assume should be `I*w*im`? Not really related, just thought I could mention if you wanted to fix.

src/freqresp.jl
```diff
@@ -8,12 +8,14 @@ function freqresp(sys::LTISystem, w::Real)
...
 
+freqresp(G::Union{UniformScaling, AbstractMatrix, Number}, w::Real) = G
+
 """
     sys_fr = freqresp(sys, w)
 
 Evaluate the frequency response of a linear system
 
-`w -> C*((iw*im -A)^-1)*B + D`
+`w -> C*((iw*im*I - A)^-1)*B + D`
```
albheim, line 18: I was more thinking of the `w -> ... iw ...` in my comment, I assume this should be `w->...w...`? I added the `I` since I thought maybe that is where the `iw` came from?

albheim: But yeah, not really related (and not very important) so can fix it some other time.

PR #632, opened 2022-02-15
Fix occasional problems with `solve` not including endpoint

baggepinnen:
Due to numerical rounding issues, `OrdinaryDiffEq.solve` sometimes fails to include the last time point and then `lsim` fails due to trying to access the solution out of bounds. This PR extends the time range by half a sample time to make sure the last time point is available.

This PR also adds an option to plot the states when plotting a `SimResult`

### DIFF
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -103,12 +103,15 @@ end
 
 
 # This will be called on plot(lsim(sys, args...))
-@recipe function simresultplot(r::SimResult; plotu=false)
-    ny, nu = r.ny, r.nu
+@recipe function simresultplot(r::SimResult; plotu=false, plotx=false)
+    ny, nu, nx = r.ny, r.nu, r.nx
     t = r.t
     n_series = size(r.y, 3) # step and impulse produce multiple results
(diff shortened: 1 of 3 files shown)


src/plotting.jl
```diff
@@ -103,12 +103,15 @@ end
 
 
 # This will be called on plot(lsim(sys, args...))
-@recipe function simresultplot(r::SimResult; plotu=false)
-    ny, nu = r.ny, r.nu
+@recipe function simresultplot(r::SimResult; plotu=false, plotx=false)
```
albheim, line 106: Are these keywords documented somewhere? Cause now that we have plotrecipes for types (and not bound to specific plotting function) it will be harder to make good documentation? Or can we just bind documentation to the plot function? Seems messy to me.

albheim (approved):
LGTM

src/plotting.jl
baggepinnen: No they can't really be documented anywhere other than in the general docs or perhaps docs of `lsim`. I'll add a note in the lsim docstring

PR #653, opened 2022-02-23
add precompilation

baggepinnen:
I'm marking this as a draft. The current benefit is not really large enough to include, but 
- https://github.com/JuliaLang/julia/pull/43990

may make it more worthwhile. In particular, I'm worried about the extra overhead incurred by developers of the package, hence the experimentation with
```julia
if !(occursin(joinpath(".julia", "dev"), __CONTROLSYSTEMS_SOURCE_DIR__))
    # Only precompile if the package is not checked out for development. Precompilation takes about 6 seconds, and saves about 2-3 seconds on the compilation script. For a user loading the same version of the package multiple times, precompilation makes the experience slightly more snappy. For a developer, paying the precompilation price each time the package has been modified is unlikely to be beneficial.
    include("precompile.jl")
end
```

The user may save about 2 seconds on a simple but representative workflow, but the loading time of the package goes up by about 0.15s. The *precompilation time* of the package will, however, go up with a whopping 4 seconds, so unless there is a bigger benefit of the precompilation, I'm not sure it's worth it.

### DIFF
--- a/generate_precompile.jl
+++ b/generate_precompile.jl
@@ -1,0 +1,12 @@
+G1 = ssrand(1,1,1)
+G2 = tf(G1)
+G3 = ssrand(1,1,1,Ts=1)
+G4 = tf(G3)
+
+u = randn(1,5)
+w = exp10.(LinRange(-2, 2, 20))
+for G in (G1, G2, G3, G4)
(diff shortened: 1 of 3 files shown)


baggepinnen:
Closed by #838

Issue #640, opened 2022-02-18
Multiple systems in same plot has inconsistent display


albheim:
Have two systems H and H2 to compare. Running
```julia
bodeplot([H, H2])
```
I get a nice plot with legend G1 and G2 so I know which is which. Running
```julia
plot(step.([H, H2]))
```
does not give any. Easy to add in the call, but just felt if there are automatic labels for one one there might as well be for the other to be consistent.

baggepinnen:
I'm opposed to automatic labels, because they behave very poorly if you call bodeplot twice. Also, if you plot and set a title, next call to bodeplot will override your title which is quite terrible.

baggepinnen:
The same applies to our custom ticks in bode. Plot twice where the axes differ between the plots and the ticks will be messed up.

albheim:
I'm mostly arguing for consistency, so I'd be equally up for not having it anywhere (or maybe more now that I think about the problems it might cause).

baggepinnen:
I also favor consistency. Every time we've tried to be smart in the plot recipes it has backfired and been prone to breakage with time, so I'd really like to get rid of most of the smartness, like custom ticks, labels and latex stuff.

Issue #603, opened 2022-01-14
nyquistplot strange behaviour


albheim:
Saw that the nyquist plot in the updated [ControlExamplePlots](https://github.com/JuliaControl/ControlExamplePlots.jl/pull/20) looked a bit odd. It writes "u(1) to y(1)" on all four subplots, and the xrange does not seem to scale. Remade locally and can confirm this is the same plot I get.

Had a quick look at the code and to me it seems like neither of these should happen.

baggepinnen:
Plots has become a bit buggy with default keyword arguments in plot recipes lately. It sometimes helps to move the keyword definition slightly within the recipe. I'll try to have a look tomorrow, can see of we can come up with a nice example to open an issue on the plots repo because it's quite annoying

Issue #435, opened 2021-01-18
Problems with getPhaseTicks and getLogTicks


baggepinnen:
These two functions work well when only a single system is plotted. When plotting two systems in the same plot, the ticks are chosen based on the last system only, if the last system spans other (notably fewer) ticks, there will be no ticks covering parts of the plot. An example is below, where the first command produces a nice looking plot, but the second call removes the phase ticks for most of the plot. 

These two functions also cause significant problems with the ticks rendering for some of the backends, they are quite broken for `plotly`.
```julia
julia> P = tf(1,[1.0, 1, 0])
TransferFunction{Continuous, SisoRational{Float64}}
     1.0
-------------
1.0s^2 + 1.0s

Continuous-time transfer function model

julia> bodeplot(P*delay(1), 0.5:0.1:100)

julia> bodeplot!(P, 0.5:0.1:100)
```

PR #667, opened 2022-03-11
deactivate broken features from bodeplot

baggepinnen:
This PR deactivates a lot of the "smart features" in the `bodeplot` since they are broken in too many cases.
- https://github.com/JuliaControl/ControlSystems.jl/issues/435
- https://github.com/JuliaControl/ControlSystems.jl/issues/640
- https://github.com/JuliaControl/ControlSystems.jl/issues/603

The biggest problem is that if `bodeplot` is invoked twice, the second invocation overrides everything that was set during the first invocation. This causes x/y-ticks to be garbage and overrides user set titles and labels. Also, the automatic label `G_1(s)` does not increase the subscript number unless bodeplot is invoked once only, causing multiple systems to get the same label.

I have tried for a while to fix the problems, but it requires making use of a lot of internals and I didn't manage to get a robust solution. If anyone wants to commit to implementing and maintaining a really solid and robust solution that works for all backends, I'd be up for that, otherwise I say we ditch all the broken code.

### DIFF
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -25,81 +25,81 @@ function setPlotScale(str::AbstractString)
     _PlotScale, _PlotScaleFunc, _PlotScaleStr = plotSettings
 end
 
-"""
-Get atributes from xlims or ylims
-default to extrema(wmag) if xlims/ylims not defined or empty
-"""
-function getlims(xylims, plotattributes, wmag)
-    lims = get(plotattributes, xylims, extrema(wmag))
-    if !isa(lims, Tuple{<:Number, <:Number}) # If x/ylims not supplied as empty
-        lims = extrema(wmag)
-    end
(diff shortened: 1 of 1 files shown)


albheim (approved):
I can't say I really like some of the defaults after this change, e.g. y1 and y2 labels in mag resp phase for a single curve, but then there were some problematic default before also. And it is not really that much of a problem to send in some keywords from the user side if one wants something more specific in the label.

I think it seems like a reasonable step to take. If anyone wants to specify a lot of stuff it is probably less chance we have overridden something to make that problematic, and for just quick plots we have all basic information that is needed in a format that is good enough.

baggepinnen:
Yeah, the `y1` default label Plots uses feels rather arbitrary, but then again, it's the same for every kind of plot using Plots :/

PR #669, opened 2022-03-12
Nonlinear systems using LFT representation

baggepinnen:
Add support for simulation of systems with static nonlinear components in a feedback loop. This framework is flexible, and can model things like
- input saturation (available as `saturation`)
- constant offsets (example: non-zero operating points)  (available as `offset`)
- Coulomb friction (e.g., as `nonlinearity(sign)`)
- Duffing oscillators (nonlinear springs)
- anti windup

Currently, the "component library" is very small, but it's trivial to add user-defined functions. Current limitations are
- Only `Continuous`. This can easily be extended to discrete-time but that will come in a separate PR.
- Experimental status. All entry-points are documented as experimental and subject to breakage without respecting semver. No functions are exported either to reflect the experimental status.
- No docs. Docs will be added before the functionality is considered stable.
- Algebraic loops are not handled even in cases where they could be if a nonlinear root solver was used. The added complexity that would be required makes me think it's not worth implementing this.


src/hammerstein_weiner.jl
```diff
@@ -0,0 +1,265 @@
...
+function _bounds_and_features(sys::HammersteinWienerSystem, plot::Val)
+    _bounds_and_features(sys.P.P, plot)
+end
+
+# Againm we have to do something for default vectors, more or less a copy from timeresp.jl
+function _default_dt(sys::HammersteinWienerSystem)
+    _default_dt(sys.P.P)
+end
+
+function Base.step(sys::HammersteinWienerSystem{T}, t::AbstractVector; kwargs...) where T
```
mfalt, line 202: Can we not use a previous implementation here?

mfalt:
Looks good to me after a quick check. But the name change from typo to PartitionedStateSpace might be breaking (although unlikely?).

src/hammerstein_weiner.jl
baggepinnen: Good point. The standard `step` supports stuff that we don't here, like `method=:zoh` but does not make use of the inplace `u` and is actually fundamentally type unstable, so I'm leaning more towards making use of this new implementation in favor of the old one.

PR #688, opened 2022-06-01
Add argument to nyquistplot for critical point

Zinoex:
Considering a Nyquist plot for the return difference F = 1 + L, we should mark the origin as the critical point instead of -1. Similarly, it allows plotting of the determinant of the return difference for a MIMO system, as we necessarily need to test encirclements of the origin. Default remains -1 as the critical point.

### DIFF
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -306,7 +306,7 @@ If `hz=true`, the hover information will be displayed in Hertz, the input freque
 `kwargs` is sent as argument to plot.
 """
 nyquistplot
-@recipe function nyquistplot(p::Nyquistplot; Ms_circles=Float64[], unit_circle=false, hz=false)
+@recipe function nyquistplot(p::Nyquistplot; Ms_circles=Float64[], unit_circle=false, hz=false, critical_origin=false)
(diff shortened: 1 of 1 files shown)


baggepinnen:
Maybe a good option would be to let the argument control the position of the point, so that 
```
critical_point = -1
```
would be the default and 
```
critical_point = 0
```
would mark the origin?

Zinoex:
@albheim For a generalized Nyquist stability criterion, you plot `det(I + L)` where `L` is a matrix of transfer functions. Hence, you cannot modify the problem to encircle -1. Also, the Nyquist stability criterion is derived using the principle of the argument for encirclements around the origin for `F = 1 + L`, thus it may be desirable to plot `F` instead. I am by no means an expert in this area, but considering in particular the MIMO case, it is just a low-hanging fruit of 'nice-to-have'.

baggepinnen:
Cool, thanks! The argument should probably be reflected in the docstring as well, and then we're good to merge

baggepinnen:
Thanks for your contribution, a new release will be out soon :)

PR #695, opened 2022-06-20
improvements to gangoffour and gofplot

baggepinnen:
Now works for MIMO systems etc.

### DIFF
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -784,7 +784,7 @@ function gangoffourplot(P::Union{<:Vector, LTISystem}, C::Vector, args...; minim
         Plots.hline!([1.0], l=(:dash, :black), sp=1, ylims=(1e-3,1.8))
     end
     f2 = bodeplot(D, args...; show=false, title="P/(1+PC)", plotphase=false, kwargs...)
-    Plots.hline!(ones(1, length(D)), l=(:black, :dash), primary=false, sp=(1:length(D))')
+    Plots.hline!(ones(1, ninputs(D[1])*noutputs(D[1])), l=(:black, :dash), primary=false)
     f3 = bodeplot(N, args...; show=false, title="C/(1+PC)", plotphase=false, kwargs...)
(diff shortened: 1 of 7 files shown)


src/plotting.jl
```diff
@@ -759,29 +759,41 @@ pzmap(sys::LTISystem; kwargs...) = pzmap([sys]; kwargs...)
...
-    RecipesBase.plot!(fig, [x-> _PlotScale == "dB" ? 0 : 1 for _ in 1:4], l=(:black, :dash), primary=false)
-    titles = fill("", 1, plotphase ? 8 : 4)
-    # Empty titles on phase
-    titleIdx = plotphase ? [1,2,5,6] : [1,2,3,4]
-    titles[titleIdx] = ["S = 1/(1+PC)", "P/(1+PC)", "C/(1+PC)", "T = PC/(1+PC)"]
-    RecipesBase.plot!(fig, title = titles)
-    return fig
+    plots_id = Base.PkgId(UUID("91a5bcdd-55d7-5caf-9e0b-520d859cae80"), "Plots")
+    haskey(Base.loaded_modules, plots_id) || error("Call using Plots before calling this function")
+    Plots = Base.loaded_modules[plots_id]
```
albheim, line 775: Nice way to solve the implicit dependency. Just thought that one other nice thing with recipes is that they could technically work with more backends, e.g. Makie, IIUC. Maybe not the main focus for us, so I don't think it should hinder this PR.

albheim: Nvm, just tried and seems Makie does not play with the recipes we have so I guess they have their own recipe system? I was pretty sure I had tried this before and it worked, seems like a miss if they aren't cross compatible with the recipes :/

baggepinnen: makie did have some very basic support for Plots recipes a long time ago, but it got stale and never updated.

baggepinnen: If makie did work with Plots recipes, I'd probably use Makie way more often. Now, there's just too much invested in Plots recipes to bother making the change :P

Issue #738, opened 2022-09-06
Split out OrdinaryDiffEq functionality


baggepinnen:
In Julia v1.8, OrdinaryDiffEq takes an enormous amount of time to precompile and load without a system image. The following breakdown indicates that most of the latency when loading CS is related to `OrdinaryDiffEq` and dependencies of `OrdinaryDiffEq` such as `ReverseDiff`. I therefore propose to create a `ControlSystemsBase` where most things, everything that does not significantly contribute to latency, is kept, and `ControlSystemsSimulation` that contains the features that depend on `OrdinaryDiffEq`. The package `ControlSystems` will continue to export all the existing functionality, so the change would be non-breaking and only add the option to load `ControlSystemsBase` for workflows that do not need continuous-time simulation.
```
julia> @time using ControlSystems
 11.893709 seconds (48.08 M allocations: 3.387 GiB, 7.21% gc time, 5.97% compilation time: 65% of which was recompilation)

julia> @time_imports using ControlSystems
     34.1 ms  RecipesBase
      0.5 ms  LaTeXStrings
    114.9 ms  Polynomials
      2.6 ms  DocStringExtensions 66.35% compilation time
      0.1 ms  Reexport
      0.2 ms  Requires
      1.8 ms  ArrayInterfaceCore
      1.4 ms  StaticArraysCore
    477.7 ms  StaticArrays
      9.0 ms  FunctionWrappers
      0.2 ms  MuladdMacro
      4.7 ms  OrderedCollections
      0.1 ms  UnPack
      0.2 ms  Parameters
     11.4 ms  FiniteDiff 60.64% compilation time (13% recompilation)
      3.7 ms  IrrationalConstants
      0.5 ms  DiffRules
      1.6 ms  DiffResults
     13.6 ms  Preferences
      0.1 ms  OpenLibm_jll
      0.2 ms  NaNMath
      0.3 ms  Compat
     46.3 ms  ChainRulesCore
      0.3 ms  ChangesOfVariables
      0.4 ms  InverseFunctions
      0.5 ms  LogExpFunctions
      0.2 ms  JLLWrappers
      0.2 ms  CompilerSupportLibraries_jll
     20.0 ms  OpenSpecFun_jll 98.10% compilation time (94% recompilation)
     10.5 ms  SpecialFunctions
     46.0 ms  MacroTools
      0.2 ms  CommonSubexpressions
     91.8 ms  ForwardDiff
[...]

PR #746, opened 2022-09-08
Split out most functionality into ControlSystemsBase

baggepinnen:
Ref: #738

Timings without any system image:
```julia
julia> @time using ControlSystemsBase
  1.70 seconds
```
```julia
julia> @time using ControlSystems
  21.71 seconds
```
The difference in pre-compilation time is several minutes.

This moves most functionality into the sub-folder `lib/ControlSystemsBase`, which I intend to release in General as a separate package. The documentation is still contained in the same location as before, and will hopefully work and look more or less the same. I edited some of the examples in the docs to only load ControlSystemsBase if applicable to indicate the readers that this is a possibility, I also explained the differences in the readme and in the installation section in the docs. 

All tests fail since ControlSystemsBase is not registered, so I think we'll just have to
1. merge
2. register ControlSystemsBase
3. make sure tests pass (tests for both packages pass locally)
4. Tag new release of ControlSystems

@albheim does this sound reasonable?


albheim:
That is quite a difference in load time, nice job!

One thing I'm thinking about is if the tests should not use the local version of `ControlSystemsBase` (dev it) and thus not need it registered (we should still register though). If not changes in base would require a long chain before actually being tested?
push changes A -> register changes A -> push other changes B that now pull A -> A gets tested

baggepinnen:
Good point, I'm all for speeding up developer workflows! :) 

We need to find a way to guarantee that we indeed register a new version of Base as well if something has changed there. If we test ControlSystems with `dev Base`, we can pass tests that would fail for the user since no new release of Base has been made. 
We could potentially run tests with both `dev` and the latest registered version, but that will lead to very long CI times.  In practice, maybe this is not a huge problem since the amount of functionality that lives in ControlSystems after this PR is rather small, and will likely not be touched nearly as often as the stuff in Base.

baggepinnen:
It's currently set up so that any change to this github repo (including all packages) triggers all tests.
- The tests for ControlSystemsBase are independent of ControlSystems, but any change will test ControlSystems as well
- The tests for ControlSystems use the latest state of ControlSystemsBase (dev)

They do run in parallel, so we might get some speedup from this compared to before. 

So for changes to ControlSystems, we only need to tag a new version of ControlSystems. For changes to ControlSystemsBase, we will notice if the ControlSystems tests fail and will thus have to update that as well. 

I think this should work out well :)

PR #761, opened 2022-10-27
Notify user what vector/matrix size is expected for x0, u

mzaffalon:


mzaffalon:
It looks like the check `u0 isa Number` in https://github.com/JuliaControl/ControlSystems.jl/blob/7b4c1aa1f879226a4bf351cbb2dbd916457e9290/src/timeresp.jl#L29 should also go:

```
julia> sys = ss([1 0; 0 1], [1; 0], [1 0], 0);

julia> lsim(sys, (x,t) -> t≥0, 5)
ERROR: ArgumentError: indexed assignment with a single value to possibly many locations is not supported; perhaps use broadcasting `.=` instead?
Stacktrace:
 [1] setindex_shape_check(::Bool, ::Int64, ::Int64)
   @ Base .\indices.jl:261
 [2] _unsafe_setindex!(::IndexLinear, ::Matrix{Float64}, ::Bool, ::Base.Slice{Base.OneTo{Int64}}, ::Int64)
   @ Base .\multidimensional.jl:939
 [3] _setindex!
   @ .\multidimensional.jl:930 [inlined]
<snip>
```

baggepinnen:
To elaborate on why this is a problem, when having a question about why a particular implementation or line of code looks like it does, it's useful to consult the git history and check the commit that last changed the code in question. Commits that make trivial white-space changes makes this much harder since one now has to look several steps back in the history to reach the actual meaningful change to the code. The tool shown in the screenshot below, that indicates the last commit message
> Accept controller returning scalar

and links to the relevant commit would be broken by such white-space changes
image

mzaffalon:
Sorry about it: this PR was prior to you comment in https://github.com/JuliaControl/ControlSystems.jl/pull/766#issuecomment-1297428741. Please wait for the updated PR.

mzaffalon:
There are two `ControlSystemBase`: the independent package and a copy inside lib, which is not used. Moreover the package hash is weird:

```
(ControlSystems) pkg> st
Project ControlSystems v1.5.1
Status `C:\Users\michele.zaffalon\.julia\dev\ControlSystems\Project.toml`
  [aaaaaaaa] ControlSystemsBase v1.0.1
  [bcd4f6db] DelayDiffEq v5.38.3
  [459566f4] DiffEqCallbacks v2.24.3
 <snip>
```
I guess I should make a PR towards `ControlSystemsBase` too.

baggepinnen:
The source code for ControlSystemsBase is located in the same repo as ControlSystems, under the `/lib` folder. There is no independent repository for ControlSystemsBase. The hash displayed above is only the first few characters, the full hash is much longer and it's correct.

baggepinnen:
There does not have to be a one-to-one mapping between julia packages and gitrepositories, one repo can contain multiple packages, which is the case here :)

mzaffalon:
Thank you. I didn't know that and now I see that there is a different `Project.toml` for `ControlSystemsBase`. And also I didn't know that one could do `dev ControlSystemsBase` inside a devved `ControlSystems.jl`.

Issue #764, opened 2022-10-31
rlocus method


mzaffalon:
The method `rlocus` is currently (v1.5.0) is deprecated and calls `rlocusplot`. Is there a plan to re-introduce `rlocus` that returns the root loci, in the same way as `bode`-`bodeplot` and `nyquist`-`nyquistplot`?
If this is the intention and nobody is doing it, I can take a stab at a PR.

baggepinnen:
There is no such intention, but don't let that discourage you from implementing it in a PR if you would find it useful :)

baggepinnen:
Thanks for your contribution :)

PR #766, opened 2022-10-31
Add rlocus

mzaffalon:


mzaffalon:
Addresses #764.

baggepinnen:
Cool, thanks for this PR!

We should probably add the output arguments to the docstring as well since there's more than one. 

I've noticed that your editor/IDE appears to do some normalization of the code, removing white space etc. on lines unrelated to the PR. This is generally undesirable for pull requests since it adds a lot of noise in the git history, could you perhaps find a way to turn that off for repos that are not your own to avoid this?

mzaffalon:
Sorry about the poor thought-out doc string.

Yes, I have set up Emacs to remove trailing white spaces. I thought white spaces after the code line end was unwanted, but if it is your wish not to do that, I will disable that feature.

mzaffalon:
I see that there are tests for `rlocusplot`: do you think I should add any for `rlocus`, i.e. check against an analytical solution? From a quick check I saw a difference of 0.1-1.

baggepinnen:
Unfortunately, the tests for `rlocusplot` are quite bad, it's hard to test plots properly :/ 
A test for `rlocus` would thus be very welcome :)

PR #768, opened 2022-11-03
Notify user what vector/matrix size is expected for x0, u

mzaffalon:


mzaffalon:
Replaces https://github.com/JuliaControl/ControlSystems.jl/pull/761

Also, when u0 is a scalar, there is the error message reported here: https://github.com/JuliaControl/ControlSystems.jl/pull/761#issuecomment-1293071647.

baggepinnen:
For this method, `u` should always have time in the second dimension, we allow `AbstractVectOrMat` simply to be able to provide the error message for a vector telling the user what size of `u` is expected.

> One last thing: some files do not have a newline after the last line and Emacs works very hard to add one: sorry if such a change escaped my attention.

Adding this newline is much less problematic since it doesn't change any existing code, so don't worry about hem :)

baggepinnen:
Thanks for fixing up the PR, and also for fixing some of the docs :)

mzaffalon:
> For this method, `u` should always have time in the second dimension, we allow `AbstractVectOrMat` simply to be able to provide the error message for a vector telling the user what size of `u` is expected.

I assume you mean for discrete systems: continuous systems need a function that returns a vector with dimention `nu`. I was speaking about continuous systems, in particular https://github.com/JuliaControl/ControlSystems.jl/blob/master/src/timeresp.jl#L29. Would it not be easier to impose `u` be just a vector?

Why is `x0` initialized with `zeros(Bool)`?

baggepinnen:
> Would it not be easier to impose u be just a vector?

If it solves https://github.com/JuliaControl/ControlSystems.jl/issues/769 then sure, sounds like a good idea.

> Why is x0 initialized with zeros(Bool)?

To make sure we do not accidentally promote the state to a larger type than what is used for the system parameters. Nowadays we have `ControlSystemsBase.numeric_type` which could be used to initialize `x0` instead.

PR #784, opened 2022-12-15
Fix links in README file.

pietrop88:
All README links to the documentation pages lead to a 404 page because their paths contain "latest" directory instead of "dev" directory.


baggepinnen:
Thanks! This was changed recently and I'm not sure why tbh, I had totally missed updating the links 😯

PR #792, opened 2023-01-06
RootLocusResult recipe

mzaffalon:
Add struct RootLocusResult and Plot recipe

### DIFF
--- a/lib/ControlSystemsBase/src/types/result_types.jl
+++ b/lib/ControlSystemsBase/src/types/result_types.jl
@@ -63,3 +63,36 @@ function Base.getproperty(r::SimResult, s::Symbol)
         throw(ArgumentError("Unsupported property $s"))
     end
 end
+
+
+"""
+    RootLocusResult{T} <: AbstractResult
+
+Result structure containing the result of the root locus using `rlocus`.
+The structure can be plotted using
+```julia
+result = rlocus(...)
+plot(result; plotu=true, plotx=false)?
+```
+and destructured like
+```julia
+roots, Z, K = result
+```
+
+# Fields:
+- `roots::Tr`
+- `Z::Tz`
+- `K::Tk`
+- `sys::Ts`
+"""
+struct RootLocusResult{Tr, Tz, Tk, Ts} <: AbstractResult # Result of rlocus
+    roots::Tr
+    Z::Tz
+    K::Tk
+    sys::Ts
+end
+
+Base.iterate(r::RootLocusResult)               = (r.roots, Val(:Z))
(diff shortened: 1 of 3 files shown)


mzaffalon:
I am not familiar with `Plots` and its recipes system, so I just copy-pasted the `rlocusplot`'s code. I would love to have somebody show me how to avoid code duplication.

baggepinnen:
Thanks for this PR!

Looking at the changes, it does not like you have duplicated any code from `rlocusplot`? Since the tests pass, you have not broken the old functionality either so I think this looks good as it is :)

mzaffalon:
Yes, no more duplication: I figured out how to do it this morning.

There are two more changes I would like to discuss:

1. can the return type of `getpoles(G, K::Number)` https://github.com/JuliaControl/ControlSystems.jl/blob/master/src/root_locus.jl#L45 be the same as a that of `getpoles(G, K::AbstractVector{T}) where {T<:Number}`, that is a `Matrix` instead of an transposed matrix?
2. should the types of `RootLocusResult` https://github.com/JuliaControl/ControlSystems.jl/pull/792/commits/fe81aa3fed805be31f620a3b4fe613e8a5f2210e#diff-10c6446e6ced8e95a649dc28592df86d16e33c92d7898a7d57d6a29ea5c36018R88 be tighter?

baggepinnen:
> can the return type of getpoles(G, K::Number) 
 ttps://github.com/JuliaControl/ControlSystems.jl/blob/master/src/root_locus.jl#L45 be the same as a that of getpoles(G, K::AbstractVector{T}) where {T<:Number}, that is a Matrix instead of an transposed matrix?

Yeah I guess. It shouldn't really matter for generic code, but this is typically not a performance sensitive function so the `copy` would probably not be problematic. 

> should the types of RootLocusResult https://github.com/JuliaControl/ControlSystems.jl/commit/fe81aa3fed805be31f620a3b4fe613e8a5f2210e#diff-10c6446e6ced8e95a649dc28592df86d16e33c92d7898a7d57d6a29ea5c36018R88 be tighter?

Do they need to be? Users are not expected to create this type themselves so guarding against the wrong field types is not required, and all fields of the struct are fully parameterized so there would be no performance benefit either.

baggepinnen:
Well, you could make that restriction on the fields of the type if you want, I think the only difference would be that the type would print less verbosely :)

mzaffalon:
You are right, there is no much gain.

I think the PR is complete: let me know if you have feedback.

baggepinnen:
Thanks for the PR, a new release will be out shortly :)

PR #794, opened 2023-01-12
Automatically select line color for plotting the unit circle in `pzmap`.

gruberchr:
Instead of setting the line color of the unit circle in `pzmap` explicitely (to `:black`), the foreground color defined in the plot attributes is used to plot the unit circle. This is useful, when using dark themes for instance.

### DIFF
--- a/lib/ControlSystemsBase/src/plotting.jl
+++ b/lib/ControlSystemsBase/src/plotting.jl
@@ -770,7 +770,6 @@ pzmap
                 seriestype := :shape
                 fillalpha := 0
                 Plots.partialcircle(0, 2π, 100)
-                C,S
             end
         end
     end
--- a/lib/ControlSystemsBase/src/plotting.jl
+++ b/lib/ControlSystemsBase/src/plotting.jl
@@ -768,9 +768,9 @@ pzmap
         if isdiscrete(system)
             θ = range(0,stop=2π,length=100)
             S,C = sin.(θ),cos.(θ)
-            seriestype := :path
             @series begin
-                linecolor := :black
+                seriestype := :shape
+                fillalpha := 0
                 C,S
             end
         end
--- a/lib/ControlSystemsBase/src/plotting.jl
+++ b/lib/ControlSystemsBase/src/plotting.jl
@@ -766,11 +766,10 @@ pzmap
         end
 
         if isdiscrete(system)
-            θ = range(0,stop=2π,length=100)
-            S,C = sin.(θ),cos.(θ)
             @series begin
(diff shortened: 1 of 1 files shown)


baggepinnen:
Thanks for the PR!

I don't think the code works, you need to remove `C,S` in the end of the series as well?

lib/ControlSystemsBase/src/plotting.jl
```diff
@@ -766,11 +766,10 @@ pzmap
...
         if isdiscrete(system)
-            θ = range(0,stop=2π,length=100)
-            S,C = sin.(θ),cos.(θ)
-            seriestype := :path
             @series begin
-                linecolor := :black
+                seriestype := :shape
+                fillalpha := 0
+                Plots.partialcircle(0, 2π, 100)
                 C,S
```
baggepinnen, line 773:
```suggestion
```

baggepinnen:
It still will not work, the name `Plots` is not available in the recipe

baggepinnen:
- #795 

fixes it so it works, if you want, you can send me your github-registered email in private and I can add you as co-author so you show up as contributor :)

gruberchr:
> It still will not work, the name Plots is not available in the recipe

I'm just curious. Why is the `Plots.jl` package loaded by UUID in `ControlSystems.jl`?

mzaffalon:
You mean here: https://github.com/JuliaControl/ControlSystems.jl/blob/9f80f34fa3129ce38d1a03e46e8322c9f018a94a/lib/ControlSystemsBase/src/ControlSystemsBase.jl#L223? I don't think it loads it, `ControlSystemsBase` tells the user to load `Plots`.

gruberchr:
I mean this: https://github.com/JuliaControl/ControlSystems.jl/blob/ed0b8ac32482a3805650842cd10d2d717733a081/lib/ControlSystemsBase/src/plotting.jl#L771-L773

I don't know the exact meaning of these code lines. It seemed to me, that `Plots.jl` is loaded by UUID.

baggepinnen:
RecipesBase is a lightweight package that allows package authors to implement plot recipes without having to take a dependency on the much heavier Plots.jl. However, in some recipes, like the recipe touched by this PR, requires some functionality only available in Plots. In these cases, we check if plots is loaded, and if it is, we obtain a handle to the `Plots` module from the `Base.loaded_modules` dict such that we can use the functionality available in Plots. I think that the `haskey` check here is never going to fail, since the recipe cannot be called at all if Plots hasn't been loaded by the user or by another package.

PR #804, opened 2023-02-05
Charpoly computation improved

stepanoslejsek:
I took a task for improving computation of characteristic polynomial as it was mentioned in the #TODO note using the provided article. Testing withing runtest.jl was successful.

### DIFF
--- a/lib/ControlSystemsBase/src/types/conversion.jl
+++ b/lib/ControlSystemsBase/src/types/conversion.jl
@@ -271,7 +271,7 @@ function convert(::Type{TransferFunction{TE1,SisoRational}}, sys::StateSpace{TE2
 end
 
 
-function convert(::Type{TransferFunction{TE,SisoZpk{T,TR}}}, sys::StateSpace) where {TE,T<:Number, TR <: Number}
+function convert(::Type{TransferFunction{TE,SisoZpk{T,TR}}}, sys::AbstractStateSpace) where {TE,T<:Number, TR <: Number}
     matrix = Matrix{SisoZpk{T,TR}}(undef, size(sys))
 
     for i=1:noutputs(sys), j=1:ninputs(sys)
@@ -280,7 +280,8 @@ function convert(::Type{TransferFunction{TE,SisoZpk{T,TR}}}, sys::StateSpace) wh
     end
     TransferFunction{TE,SisoZpk{T,TR}}(matrix, TE(sys.timeevol))
 end
-function convert(::Type{TransferFunction{TE1,SisoZpk}}, sys::StateSpace{TE2,T0}) where {TE1,TE2,T0<:Number}
+function convert(::Type{TransferFunction{TE1,SisoZpk}}, sys::AbstractStateSpace{TE2}) where {TE1,TE2}
+    T0 = numeric_type(sys)
     T = typeof(one(T0)/one(T0))
     convert(TransferFunction{TE1,SisoZpk{T,complex(T)}}, sys)
 end
@@ -298,17 +299,42 @@ function siso_ss_to_zpk(sys, i, j)
     return z, eigvals(A), k
 end
 
+"""
+Implements: "An accurate and efficient algorithm for the computation of thecharacteristic polynomial of a general square matrix."
+"""
+function charpoly(A::AbstractMatrix{T}) where {T}
+    N = size(A, 1)
+    poly_factors = vec(ones(T, N+1))
+    if N <= 0
+        return poly_factors
+    end
 
-# TODO: Could perhaps be made more accurate. See: An accurate and efficient
-# algorithm for the computation of the # characteristic polynomial of a general square matrix.
-function charpoly(A::AbstractMatrix{T}) where T
-    Λ::Vector{T} = eigvalsnosort(A)
-    return prod(roots2poly_factors(Λ))::Polynomial{T,:x} # Compute the polynomial factors directly?
+    t = zeros(T, N, N) # Preallocation
+    P, H = hessenberg(A)
+    for j=N:-1:1
+        for i=1:j
+            for k=N-j:-1:1
+                t[k+1, i] = H[i, j]*t[k, j+1]-H[j+1, j]*t[k, i]
+            end
+            t[1, i] = H[i, j]
+        end
+        for k=1:N-j
+            t[k, j] += t[k, j+1]
+        end
+    end
+    i = N-1:-1:0
+    @. poly_factors[2:N+1, 1] = (-1)^(N-i)*t[N-i, 1]
+    return poly_factors
 end
-function charpoly(A::AbstractMatrix{<:Real})
-    Λ = eigvalsnosort(A)
-    return prod(roots2real_poly_factors(Λ))
-end
+
+# function charpoly(A::AbstractMatrix{T}) where T
+#     Λ::Vector{T} = eigvalsnosort(A)
+#     return prod(roots2poly_factors(Λ))::Polynomial{T,:x} # Compute the polynomial factors directly?
+# end
+# function charpoly(A::AbstractMatrix{<:Real})
+#     Λ = eigvalsnosort(A)
+#     return prod(roots2real_poly_factors(Λ))
+# end
 
 
 # function charpoly(A)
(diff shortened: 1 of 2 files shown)


baggepinnen:
Hello and thanks for your contribution :)

Would you mind submitting a version of the PR without any automatic formatting changes applied?

stepanoslejsek:
> 

Yeah for sure, I apologize for that formatting stuff.

baggepinnen:
No worries, thanks for updating! It seems like github has been thrown off a bit by some of your changes, it shows the entire file as affected by the diff. I think the PR needs a squash, are you familiar with how to perform one?

stepanoslejsek:
> the PR needs a

No, I'm not familiar with PR squashing, but I'm looking forward to learn about it.

baggepinnen:
hehe, it's quite simple, but it can be come messy at times. It boils down to combining several different commits so that they appear as a single commit. It can be done from the command-line interface, but I usually use a tool for the purpose, in my case, I use "git lens" in vscode.

stepanoslejsek:
Thank you for the explanation, I'm going to squash the PR then. Hopefully it won't mess up.

baggepinnen:
Hmm, your squash worked (nice ;) but there's still a whole-file diff. Base on this article
https://stackoverflow.com/questions/19593909/git-diff-sees-whole-file-as-changed-when-its-not
it seems like your editor has changed the whole file, are you by any chance using windows?

baggepinnen:
Yeah, running this command `git diff --ignore-space-at-eol master` shows that your editor has inserted a bunch of invisible characters at the end of each line (all the `^M`s). 

image

I'm not a windows person so I'm not sure how to fix that :/ I suggest googling how to turn that behavior off in your editor

stepanoslejsek:
Oh I see now, I'm using linux, let me fix those invisible characters and squash it once again.

stepanoslejsek:
It's still there, not exactly sure why, I'll try something else to deal with it.

mzaffalon:
@stepanoslejsek What editor are you using? I found that Emacs + julia-mode + magit usually work, although I have never tried on Linux.

stepanoslejsek:
I am using neovim, but I found the solution. It'll be without those invisible characters. I checked it with cat -A command.

lib/ControlSystemsBase/src/types/conversion.jl
```diff
@@ -298,18 +299,40 @@ function siso_ss_to_zpk(sys, i, j)
     return z, eigvals(A), k
 end
 
+function charpoly(A::AbstractMatrix{T}) where {T}
+    N = size(A, 1)
+    poly_factors::Vector{T} = vec(ones(T, N+1, 1))
```
baggepinnen, line 304:
```suggestion
    poly_factors = vec(ones(T, N+1))
```
the type assertion is only beneficial if type inference fails

lib/ControlSystemsBase/src/types/conversion.jl
```diff
@@ -298,18 +299,40 @@ function siso_ss_to_zpk(sys, i, j)
...
-# TODO: Could perhaps be made more accurate. See: An accurate and efficient
-# algorithm for the computation of the # characteristic polynomial of a general square matrix.
-function charpoly(A::AbstractMatrix{T}) where T
-    Λ::Vector{T} = eigvalsnosort(A)
-    return prod(roots2poly_factors(Λ))::Polynomial{T,:x} # Compute the polynomial factors directly?
-end
-function charpoly(A::AbstractMatrix{<:Real})
-    Λ = eigvalsnosort(A)
-    return prod(roots2real_poly_factors(Λ))
+    t::Matrix{T} = zeros(N, N) # Preallocation
```
baggepinnen, line 309:
```suggestion
    t = zeros(T, N, N) # Preallocation
```
the element type of a zero matrix is specified like this

lib/ControlSystemsBase/src/types/conversion.jl
```diff
@@ -298,18 +299,40 @@ function siso_ss_to_zpk(sys, i, j)
...
+    P::Matrix{T}, H::Matrix{T} = hessenberg(A)
```
baggepinnen, line 310:
```suggestion
    P, H = hessenberg(A)
```

lib/ControlSystemsBase/src/types/conversion.jl
```diff
@@ -298,18 +299,40 @@ function siso_ss_to_zpk(sys, i, j)
...
+    @simd for j=N:-1:1
+        @simd for i=1:j
+            @simd for k=N-j:-1:1
+                t[k+1, i] = H[i, j]*t[k, j+1]-H[j+1, j]*t[k, i]
+            end
+            t[1, i] = H[i, j]
+        end
+        @simd for k=1:N-j
```
baggepinnen, line 318:
```suggestion
    for j=N:-1:1
        for i=1:j
            for k=N-j:-1:1
                t[k+1, i] = H[i, j]*t[k, j+1]-H[j+1, j]*t[k, i]
            end
            t[1, i] = H[i, j]
        end
        for k=1:N-j
```
`@simd` is unlikely to do anything here since you are not using `@inbounds` annotations

lib/ControlSystemsBase/src/types/conversion.jl
```diff
@@ -298,18 +299,40 @@ function siso_ss_to_zpk(sys, i, j)
...
+            t[k, j] += t[k, j+1]
+        end
+    end
+    i = N-1:-1:0
+    poly_factors[2:N+1, 1] = (-1).^(N.-i).*t[N.-i, 1]
```
baggepinnen, line 323:
```suggestion
    @. poly_factors[2:N+1, 1] = (-1)^(N-i)*t[N-i, 1]
```

lib/ControlSystemsBase/src/types/conversion.jl
```diff
@@ -298,18 +299,40 @@ function siso_ss_to_zpk(sys, i, j)
     return z, eigvals(A), k
 end
 
+function charpoly(A::AbstractMatrix{T}) where {T}
```
baggepinnen, line 302:
```suggestion
"""
Implements: "An accurate and efficient algorithm for the computation of the characteristic polynomial of a general square matrix."
"""
function charpoly(A::AbstractMatrix{T}) where {T}
```

baggepinnen (commented):
Overall this looks good, thanks for this PR!

What I'm missing is some form of evaluation whether or not this implementation is actually faster and more accurate than the existing one. 

Benchmarking the speed should be straightforward using BenchmarkTools.jl, and benchmarking the accuracy could potentially be done by comparing to the result computed using `BigFloat`

stepanoslejsek:
I did a benchmarking on BigFloat matrix using the old and new method and here is the result of benchmarking using @btime
image

baggepinnen:
Wonderful, that's a solid 10x improvement :) I think you might have misunderstood slightly what I meant with using BigFloat though, BigFloats are for comparing *accuracy*, while performance benchmarks are likely better run using the standard `Float64`. When you compute using BigFloats, you get an extremely accurate answer which can be considered "the truth", the old and new algorithms are then computed in `Float64` and compared to the `BigFloat` result to see which one is closer to the truth.

stepanoslejsek:
I see, sorry for misunderstanding, I took a random 10x10 Float64 matrix and use the old and new algorithm and pass trough a BigFloat representation and Float64 representation and then took a L2 norm on each coefficient and here is the result.

image

baggepinnen:
That looks amazing :) Thanks for your nice contribution!

baggepinnen:
May I ask how you performed the `BigFloat` computations? When I try to use any floating-point type that is not recognized by BLAS I get
```
julia> charpoly(big.(A))
ERROR: MethodError: no method matching hessenberg!(::Matrix{BigFloat})
```

stepanoslejsek:
> May I ask how you performed the `BigFloat` computations? When I try to use any floating-point type that is not recognized by BLAS I get
> 
> ```
> julia> charpoly(big.(A))
> ERROR: MethodError: no method matching hessenberg!(::Matrix{BigFloat})
> ```

I ran into same problem with only `using ControlSystemsBase`, however I found out that in order to work, it is necessary to write `using OrdinaryDiffEq`, which is somewhat weird. I tested it with only `using ControlSystems` and then call `ControlSystemsBase.charpoly(big.(A))`.

baggepinnen:
Thanks for the hint, OrdinaryDiffEq loads GenericSchur.jl which contains an implementation of `hessenberg` that works for BigFloat. I tried with GenericLinearAlgebra which has much poorer compatibility than GenericSchur, but now I know what to use instead :)

PR #811, opened 2023-02-22
reintroduce old charpoly

baggepinnen:
to handle things for which `hessenberg` fails


baggepinnen:
A better solution is to load GenericSchur.jl (avoid GenericLinearAlgebra.jl) instead

PR #849, opened 2023-05-25
make schur factorization differentiable

baggepinnen:
Waiting for a new release that incorporates
- https://github.com/gdalle/ImplicitDifferentiation.jl/pull/51


gdalle:
Can you make it work without that release? I'm not yet sure how to make it work in reverse mode

PR #864, opened 2023-08-25
truncate T in baltrunc

baggepinnen:


hexaeder:
This should have been released as part of `ControlSystemsBase` 2.0 (and ControlSystems 2.0 for that matter) since it is a breaking change... I was actually using the full balancing transformation T for further analysis.

Just a minor gripe though, thanks for building all the ControlSystems infrastructure in the first place :)

PR #866, opened 2023-09-04
add functions to check controllability/observability using the PH test

baggepinnen:
See discussion in #865

### DIFF
--- a/lib/ControlSystemsBase/src/ControlSystemsBase.jl
+++ b/lib/ControlSystemsBase/src/ControlSystemsBase.jl
@@ -33,6 +33,8 @@ export  LTISystem,
         grampd,
         ctrb,
         obsv,
+        controllability,
+        observability,
         place,
         place_knvd,
         # Model Simplification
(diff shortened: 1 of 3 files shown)


franckgaga:
good job @baggepinnen! I'll use them in my package :)

andreasvarga:
I would move the check for square matrix before eigenvalue computation.

andreasvarga:
In a strict sense, the function `controllabiliy` checks the reachability of a LTI system. While for continuous-time systems the two notions are equivalent, for discrete-time systems there is a subtle difference between them. In general reachability implies controllability, but the reverse is not true. The controllability property is defined as the ability of a system to drive any state into the origin, while reachability is the ability to drive any state into another arbitrary state, in finite time. For example, discrete-time systems with B = 0 and A nillpotent are controllable! In MATLAB this aspect is handled incorrectly, so there is a chance to address it in a rigourous manner in Julia! The function `observability` checks the right property!

baggepinnen:
Thanks for the comment! Is your proposed change to rename the implemented function to `reachability`, or to change the implementation to distinguish between controllability and reachability for discrete-time systems? If the latter, do you know of a reference that can be used to guide the implementation?

andreasvarga:
A pertinent discussion of this aspect is, for example, in the book of Kailath (1980), starting with page 95, where also the dual concepts observability and constructibility are discussed. See also the picture on page 100, which illustrates the equivalence of these notions in discrete case if A is nonsingular. I would suggest to keep the function `controllability` as it is now, but add a remark, that it concerns with “controllability from the origin (also called reachability)”.

Issue #867, opened 2023-09-07
Rendering of state space equations in documentation


albheim:
See e.g. [here](https://juliacontrol.github.io/ControlSystems.jl/stable/lib/synthesis/#ControlSystemsBase.add_input), for me the equations for x' and y render on the same line. 
![image](https://github.com/JuliaControl/ControlSystems.jl/assets/3112674/7b7c06c4-b4ab-488f-8aed-47c7e9acc862)

There are multiple other places that seem to have the same problem.

baggepinnen:
There's unfortunately a bit of a conflict between having nice rendering in the docs (latex) and in the REPL (no latex) :( 
In some places, we use code formatting together with unicode to have a bit of a compromise, but for sufficiently complicated things that's just not really feasible. I don't know if something like https://github.com/SimonDanisch/UnicodeFun.jl could be useful to have equations display nice in both places? I've never seen anyone solving this problem before

albheim:
Yeah, I don't have a good solution. Maybe some REPL tool that detects markdown math-blocks and just "cleans" them somehow. But I don't really mind that much, I just noted it while looking around and thought I would report it quickly.

PR #868, opened 2023-09-07
fix rendering of some equations systems

baggepinnen:
addresses #867


albheim:
Haha, might have overstated the "multiple other places". Saw two and assumed there might be more... :sweat_smile:

PR #344, opened 2020-10-05
Improve documentation for tf(), step, stepplot, impulse and impulseplot

KronosTheLate:
In this PR I have changed the following
* changed the documentation for "creating systems" to include an example of using ```s=tf("s")``` syntax
* included the julia> text consistently in the examples under "creating systems"
* removed the #output comment under "creating systems" as I found it un-nessecary, and I don't like the fact that it is not copy-pasted from the REPL, which I think examples should be when possible for maximum clarity
* Clearly stated in the documentation for step, stepplot, impulse and impulseplot what they do
* Fixed a typo in the impulseplot documentation where it said "step" instead of "impulse"

PR #881, opened 2023-10-10
Incorporate changes from #344

KronosTheLate:
I finally made a branch in the correct way, and incorporated the changed made in https://github.com/JuliaControl/ControlSystems.jl/pull/344/. I left out the changes for the linestyles, as I honestly do not know their effect anymore, and the code-base has changed significantly.

Porting 344 was done by copy-pasting and incoperating to the best of my ability. The original changes were however made long ago, so this PR definitively needs some external checking.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "workbench.colorTheme": "Julia (Monokai Vibrant)"
+}
--- a/docs/src/man/creating_systems.md
+++ b/docs/src/man/creating_systems.md
@@ -59,7 +59,7 @@ s^3 + 4*s^2 + 5*s + 2
 Continuous-time transfer function model
 ```
 
-## Using the `zpk()` function - Pole-Zero-Gain Representation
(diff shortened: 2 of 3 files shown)


KronosTheLate:
> not modify the section headers?

I changed them for a reason, but fair enough. Done ^_^

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "workbench.colorTheme": "Julia (Monokai Vibrant)"
```
albheim, line 2: Was this really supposed to be added?

baggepinnen: oops, no certainly not :)

KronosTheLate: I am sorry, no. VSCode keeps adding this file in every folder I open with VSCode, which has made its way into more than one git commit. I will make a PR to remove it.

KronosTheLate: Deleted the folder in https://github.com/JuliaControl/ControlSystems.jl/pull/882

PR #882, opened 2023-10-11
Delete vscode folder

KronosTheLate:


baggepinnen:
the folder is already deleted :)

Issue #897, opened 2023-10-24
`pid` with zero proportional gain


mzaffalon:
`pid` fails when the proportional gain is 0. This is because the coefficients are [converted to the standard form](https://github.com/JuliaControl/ControlSystems.jl/blob/master/lib/ControlSystemsBase/src/pid_design.jl#L51) which assumes a non-zero proportional gain. 

```
julia> pid(0, 1; form=:parallel) # expected 1/s
TransferFunction{Continuous, ControlSystemsBase.SisoRational{Float64}}
NaNs
----
1.0
```

albheim:
Hmm, that's not good. 

I think using the standard form as the "default" that was used in conversion was something I just picked, didn't really think it through. Probably would have been better to have the parallel form for this since the parameters are not tied together and thus won't (?) lead to these problems.
On the whole that code is not too pretty, and maybe adding another special case would be the simplest for now.

baggepinnen:
Yeah, using parallel form by default is probably better in general

mzaffalon:
> Anyone want to have a look? Otherwise I could probably have a go during the week.

Let me have a go.

PR #900, opened 2023-11-11
Convert pid parameters to :parallel

mzaffalon:
Using standard fails when KP=0.

### DIFF
--- a/lib/ControlSystemsBase/src/pid_design.jl
+++ b/lib/ControlSystemsBase/src/pid_design.jl
@@ -48,43 +48,41 @@ end
 @deprecate pid(; kp=0, ki=0, kd=0, series = false) pid(kp, ki, kd; form=series ? :series : :parallel)
 
 function pid_tf(param_p, param_i, param_d=zero(typeof(param_p)); form=:standard,  Tf=nothing)
-    Kp, Ti, Td = convert_pidparams_to_standard(param_p, param_i, param_d, form)
-    ia = Ti != Inf && Ti != 0 # integral action, 0 would result in division by zero, but typically indicates that the user wants no integral action
+    Kp, Ki, Kd = convert_pidparams_to_parallel(param_p, param_i, param_d, form)
     if isnothing(Tf)
-        if ia
-            return tf([Kp * Td, Kp, Kp / Ti], [1, 0])
+        if Ki != 0
+            return tf([Kd, Kp, Ki], [1, 0])
         else
-            return tf([Kp * Td, Kp], [1])
+            return tf([Kd, Kp], [1])
         end
     else
-        if ia
-            return tf([Kp * Td, Kp, Kp / Ti], [Tf^2/2, Tf, 1, 0])
+        if Ki != 0
+            return tf([Kd, Kp, Ki], [Tf^2/2, Tf, 1, 0])
         else
-            return tf([Kp * Td, Kp], [Tf^2/2, Tf, 1])
+            return tf([Kd, Kp], [Tf^2/2, Tf, 1])
         end
     end
 end
 
 function pid_ss(param_p, param_i, param_d=zero(typeof(param_p)); form=:standard,  Tf=nothing)
-    Kp, Ti, Td = convert_pidparams_to_standard(param_p, param_i, param_d, form)
+    Kp, Ki, Kd = convert_pidparams_to_parallel(param_p, param_i, param_d, form)
     TE = Continuous()
-    ia = Ti != Inf && Ti != 0 # integral action, 0 would result in division by zero, but typically indicates that the user wants no integral action
     if !isnothing(Tf)
-        if ia
+        if Ki != 0
             A = [0 1 0; 0 0 1; 0 -2/Tf^2 -2/Tf]
             B = [0; 0; 1]
-            C = 2 * Kp / Tf^2 * [1/Ti 1 Td]
+            C = 2 / Tf^2 * [Ki Kp Kd]
         else
             A = [0 1; -2/Tf^2 -2/Tf]
             B = [0; 1]
-            C = 2 * Kp / Tf^2 * [1 Td]
+            C = 2 / Tf^2 * [Kp Kd]
         end
         D = 0
-    elseif Td == 0
-        if ia
+    elseif Kd == 0
+        if Ki != 0
             A = 0
             B = 1
-            C = Kp / Ti # Ti == 0 would result in division by zero, but typically indicates that the user wants no integral action
+            C = Ki # Ti == 0 would result in division by zero, but typically indicates that the user wants no integral action
             D = Kp
         else
             return ss([Kp])
@@ -98,7 +96,7 @@ end
 """
     pidplots(P, args...; params_p, params_i, params_d=0, form=:standard, ω=0, grid=false, kwargs...)
 
-Plots interesting figures related to closing the loop around process `P` with a PID controller supplied in `params`
+Display the relevant plots related to closing the loop around process `P` with a PID controller supplied in `params`
 on one of the following forms:
 * `:standard` - `Kp*(1 + 1/(Ti*s) + Td*s)` 
 * `:series` - `Kc*(1 + 1/(τi*s))*(τd*s + 1)`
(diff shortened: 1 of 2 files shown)


mzaffalon:
Addresses #897.

mzaffalon:
Two questions:

1. how can I increase code coverage?
2. what is the purpose of `@.` in, say, https://github.com/JuliaControl/ControlSystems.jl/blob/master/lib/ControlSystemsBase/src/pid_design.jl#L559

baggepinnen:
1. By adding tests that causes the added lines of code to be executed.
2. image

lib/ControlSystemsBase/src/pid_design.jl
```diff
@@ -511,26 +509,26 @@ function loopshapingPID(P0, ω; Mt = 1.3, ϕt=75, form::Symbol = :standard, dopl
 end
 
 """
-    Kp, Ti, Td = convert_pidparams_to_standard(param_p, param_i, param_d, form)
+    Kp, Ti, Td = convert_pidparams_to_parallel(param_p, param_i, param_d, form)
```
baggepinnen, line 512: It would be better to keep the old function as it was and just add the new function. We don't want to unnecessarily break any downstream package that might have use this function.

albheim: On the phone so haven't checked, but iirc we neither have this in online docs or export it, so I probably wouldn't count it as api.
I think the thing that is exported is the `convert_pidparams_from_to`, though I could be wrong.
But it is also easy to leave it as an extra method if you feel more comfortable with that.

baggepinnen: I'd like to keep it for a few reasons
- I make use of it in some downstream packages
- It would be nice to keep the old version around for comparison for a while in case we encounter any strange results with the new method
- Julia currently (as of v1.9, but [soon improved by the new `public` keyword](https://github.com/JuliaLang/julia/blob/master/NEWS.md#new-language-features)) lacks a good way to indicate what is and isn't public facing API, and whether or not something is included in the docs is a policy that is awkward to work with from a user perspective. I thus wouldn't be surprised if other people have found this function and made use of it. It does have a nice docstring after all, and we haven't explicitly documented any policy for what is and isn't public API in ControlSystems.

mzaffalon:
>     2. image

So it has no use on this specific line.

baggepinnen:
It did have a use, it supported parameter arrays as inputs
```julia
julia> ControlSystemsBase.convert_pidparams_from_standard(rand(3), rand(3), rand(3), :parallel)
([0.17773398598165757, 0.32161078829707446, 0.722649466067516], [0.2515777728255106, 0.36780970052897355, 0.7964910544026514], [0.09807961789854291, 0.2953669775189507, 0.042697386412817936])
```
with the refactoring currently applied in this PR, that functionality was broken. One could argue whether that functionality was needed or not, but since it was there there is no need to change the implementation for the small simplification to the code

mzaffalon:
Thank you. I hope I didn't forget any.

lib/ControlSystemsBase/src/pid_design.jl
```diff
@@ -536,36 +534,94 @@ function convert_pidparams_to_standard(param_p, param_i, param_d, form::Symbol)
...
     if form === :standard
-        return Kp, Ti, Td
+        return @. (Kp, Ti, Td)
     elseif form === :series
-        return @. (
-            (Ti - sqrt(Ti * (Ti - 4 * Td))) / 2 * Kp / Ti,
-            (Ti - sqrt(Ti * (Ti - 4 * Td))) / 2,
-            (Ti + sqrt(Ti * (Ti - 4 * Td))) / 2,
-        )
+        Δ = Ti * (Ti - 4 * Td)
```
baggepinnen, line 577: These lines are broken for parameter arrays

lib/ControlSystemsBase/src/pid_design.jl
```diff
@@ -536,36 +534,94 @@ function convert_pidparams_to_standard(param_p, param_i, param_d, form::Symbol)
...
 The `form` can be chosen as one of the following
-* `:standard` - ``K_p(1 + 1/(T_i s) + T_ds)``
+* `:standard` - ``K_p(1 + 1/(T_i s) + T_d s)``
 * `:series` - ``K_c(1 + 1/(τ_i s))(τ_d s + 1)``
 * `:parallel` - ``K_p + K_i/s + K_d s``
 """
 function convert_pidparams_from_standard(Kp, Ti, Td, form::Symbol)
     if form === :standard
-        return Kp, Ti, Td
+        return @. (Kp, Ti, Td)
```
baggepinnen, line 575:
```suggestion
        return Kp, Ti, Td
```
this `@.` didn't do anything

lib/ControlSystemsBase/src/pid_design.jl
```diff
@@ -536,36 +534,94 @@ function convert_pidparams_to_standard(param_p, param_i, param_d, form::Symbol)
...
+Convert parameters from form `:parallel` to form `to_form`.
+
+The `form` can be chosen as one of the following
+* `:standard` - ``K_p(1 + 1/(T_i s) + T_d s)``
+* `:series` - ``K_c(1 + 1/(τ_i s))(τ_d s + 1)``
+* `:parallel` - ``K_p + K_i/s + K_d s``
+"""
+function convert_pidparams_from_parallel(Kp, Ki, Kd, to::Symbol)
+    if to === :parallel
+        return @. (Kp, Ki, Kd)
```
baggepinnen, line 603:
```suggestion
        return Kp, Ki, Kd
```

lib/ControlSystemsBase/src/pid_design.jl
```diff
@@ -536,36 +534,94 @@ function convert_pidparams_to_standard(param_p, param_i, param_d, form::Symbol)
...
+    elseif to === :series
+        Ki == 0 && return @. Kp * (1, 0, Kd)
+        Δ = Kp^2-4Ki*Kd
+        Δ < 0 &&
+            throw(DomainError("The condition Kp^2 ≥ 4Ki*Kd is not satisfied: the PID parameters cannot be converted to series form"))
+        sqrtΔ = sqrt(Δ)
+        return @. ((Kp - sqrtΔ)/2, (Kp - sqrtΔ)/(2Ki), (Kp + sqrtΔ)/(2Ki))
```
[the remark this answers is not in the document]
mzaffalon: Would you rather have Julia throw a `DomainError` here? In general to me it makes more sense to convert single pids at the time.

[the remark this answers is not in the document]
baggepinnen: You could check
```
any(<(0), Δ)
```

> In general to me it makes more sense to convert single pids at the time.

I agree, that would probably be a nicer implementation. I performed some tests locally, and it does not seem like most PID-related functions actually handle array parameters, and I cannot remember why it was supported in the first place. Maybe it's safe to assume that this functionality isn't used anywhere and not bother supporting it?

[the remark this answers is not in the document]
baggepinnen: Ah, I see now that `stabregionPID` does pass in array parameters

lib/ControlSystemsBase/test/test_pid_design.jl
```diff
@@ -72,13 +84,13 @@ C, Kp, Ti = placePI(P, 2, 0.7; form=:standard)
...
 
 # Test internal functions convert_pidparams*
-params = (2, 3, 0.5)
-parallel_params = ControlSystemsBase.convert_pidparams_from_standard(params..., :parallel)
-@test parallel_params == (2, 2/3, 1)
-@test ControlSystemsBase.convert_pidparams_to_standard(parallel_params..., :parallel) == params
-series_params = ControlSystemsBase.convert_pidparams_from_standard(params..., :series)
-@test series_params == ((3-sqrt(3))/3, (3-sqrt(3))/2, (3+sqrt(3))/2)
-@test ControlSystemsBase.convert_pidparams_to_standard(series_params..., :series) == params
+params = (4, 3, 0.5)
```
baggepinnen, line 87:
```suggestion
# Standard
params = (2, 3, 0.5)
parallel_params = ControlSystemsBase.convert_pidparams_from_standard(params..., :parallel)
@test parallel_params == (2, 2/3, 1)
@test ControlSystemsBase.convert_pidparams_to_standard(parallel_params..., :parallel) == params
series_params = ControlSystemsBase.convert_pidparams_from_standard(params..., :series)
@test series_params == ((3-sqrt(3))/3, (3-sqrt(3))/2, (3+sqrt(3))/2)
@test ControlSystemsBase.convert_pidparams_to_standard(series_params..., :series) == params

# Parallel
params = (4, 3, 0.5)
```

we can keep the tests for the old method

baggepinnen:
Very nice, thanks for your hard work :)

mzaffalon:
My pleasure and thank you for your patience.

PR #913, opened 2024-01-11
Allow bode and bodeplot to use a range of frequencies and a freq to calculate the freq. vector

VaiTon:
- [x] Added docs
- [x] Added tests

### DIFF
--- a/lib/ControlSystemsBase/src/freqresp.jl
+++ b/lib/ControlSystemsBase/src/freqresp.jl
@@ -296,13 +296,27 @@ function (sys::TransferFunction)(z_or_omegas::AbstractVector, map_to_unit_circle
 end
 
 """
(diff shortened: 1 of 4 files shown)


baggepinnen:
Hello and thank you for your PR! A few general comments to start off with
- Please consider submitting PRs without running a formatter. on the whole file. The formatter makes lots of changes that are unrelated to the PR which makes it hard to see what you have actually changed in the PR, and I also generally do not like the changes the formatter does, hence it's not being used on this repo.
- Please consider opening an issue before submitting a PR so that we can discuss whether or not the idea is worthwhile pursuing. In this case, we already have a PR that attempts to do something similar, but it takes a much wider approach applying to all frequency-response functions rather than just Bode: https://github.com/JuliaControl/ControlSystems.jl/pull/532

PR #929, opened 2024-07-30
Add "asymptotic" keyword to `kalman` docstring

franckgaga:
I think it would be a bit clearer like that. I met many people that learned the classical time-varying form at school, and does not even know that the asymptotic form exists.


baggepinnen:
it's the dual to the _infinite-horizon_ LQR problem :)

PR #963, opened 2025-01-29
Extend docstring

CasBex:
The docstring of `ssdata` lead me to believe that the matrices A, B, C, D were copies of those of sys. I suggest to change the documentation as in this PR to avoid confusion.


baggepinnen:
Nowadays, you can also do
```julia
(;A, B, C, D) = sys
```

CasBex:
It's indeed not the most rocket sciencey function in this library, but it took two hours of debugging, so I hope to spare others the same fate.

Issue #975, opened 2025-02-23
Update `test.yml` with new github actions


franckgaga:
I tagged you somewhere with the same comment, but I'm raising an issue here for posterity.

Might be worth it to update the `test.yml` file in `ControlSystem.jl`. See https://discourse.julialang.org/t/with-julia-actions-setup-julia-you-can-now-specify-lts-and-pre-named-versions/

The interesting parts would be to update to `julia-actions/setup-julia@v2`, `actions/checkout@v4` and `julia-actions/cache@v2`. The v2 of cache speeds up the testing but requires this addition in the yml file:

```yml
    runs-on: ${{ matrix.os }} # new part just after this line:
    permissions: # needed for julia-actions/cache delete old caches that it has created
      actions: write
      contents: read
```

With v2 of `setup-julia` you can execute the tests on:

```yml
      matrix:
        version:
          - 'lts' # long-term support release
          - '1'   # latest stable 1.x release
          - 'pre' # latest stable prerelease
          # - 'nightly' # commented since noisy + 'pre' allows testing upcoming versions
```

IMO it covers large enough to see major issues for Julia version incompatibilities.

PR #978, opened 2025-03-30
changed: bump `ForwardDiff` compat in `ControlSystems` + `ControlSystemsBase` also

franckgaga:
more complete than #977

Some tests do not pass locally, but I think it was already the case before. (edit: just tested locally and yes, they were failing also before)

BTW, I can also try to do #975 myself.


franckgaga:
Weird It does not pass locally on my computer but it passes here. Oh well, good news XP

baggepinnen:
Did you verify that the tests actually ran with the new version? OrdinaryDiffEq might not yet be compatible so the tests might have run with an old version

franckgaga:
Forgot to answer but yes you are right the test are executed with v0.10 instead of v1.0. I'm starting to understand why the CompatHelper forces the upper bound of compat entries.

But at the same time, I find hard to fully understand dependency issues. If everyone (as in every package) is waiting for everyone to update the ForwardDiff compat entries in their `Project.toml` because of this issue, it feels like nobody will update it. Except if we are "lucky" enough and everything is done in the correct order, for exemple:

1. A dependancy of `OrdinaryDiffEq.jl` update it's codebase to be compatible with ForwardDiff v1.0 and update the compat entry
2. `OrdinaryDiffEq.jl` update it's codebase to be compatible with ForwardDiff v1.0 and update the compat entry
3. `ControlSystemsBase.jl` update it's codebase to be compatible with ForwardDiff v1.0 and update the compat entry
4. `ModelPredictiveControl.jl` update it's codebase to be compatible with ForwardDiff v1.0 and update the compat entry (I decided to update it anyway, FYI)

And realistically, there is probably many steps before step 1. Am I misunderstanding something if I feel like "waiting for the other to update in the correct order" is almost a Mission: Impossible?

franckgaga:
I will close this PR since now you have two separate CompatHelper PRs. BTW, OrdinaryDiffEq.jl now supports ForwardDiff v1.0, so you may try to re-run the jobs in #977 and it may passes now.

baggepinnen:
> BTW, OrdinaryDiffEq.jl now supports ForwardDiff v1.0

Where did you get that from? It does not look like it here

https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/Project.toml#L89

franckgaga:
Hum weird, my mistake. I look at the same place yesterday and I saw `"0.10, 1"`. I'm hallucinating like a LLM 😆

baggepinnen:
You might have looked another package with a similar name, there are _very_ many :P

oscardssmith:
what's the way forward here? do you want me to run tests locally with deved OrdinaryDiffEq to force ForwardDiff@1? Alternatively, can this be merged as is?

franckgaga:
I think it would be to wait that `OrdinaryDiffEq.jl` code and `ForwardDiff` compat entry is updated. It is planned in the short term, as mentionned by Chris [here](https://discourse.julialang.org/t/understanding-dependancy-issues-and-compathelper/127676/2?u=franckgaga).

oscardssmith:
Yeah I'm the person in the process of doing so :) The tricky part is that a lot of libraries (e.g. MTK) rely on ControlSystemsBase so not releasing a new version of it makes it more effort to update the rest of the SciML stack.

franckgaga:
Oh I see. It corroborate my theory above. It's even worse here since there is a circular dependency if I understand it right, so a correct order does not exist here.

I think the only viable alternative is to `dev` everything, modify compat entries and test locally yes. What do you think @baggepinnen ?

oscardssmith:
> I think the only viable alternative is to dev everything, modify compat entries and test locally yes

yeah. That's been my life for the past couple weeks

baggepinnen:
I don't mind merging if someone has tested locally

oscardssmith:
running the tests locally, there was 1 failure (in ControlSystemsBase):
```
test_conversion: Test Failed at /home/oscardssmith/.julia/dev/ControlSystems.jl/lib/ControlSystemsBase/test/test_conversion.jl:221
  Expression: (hinfnorm(mp))[1] < 1.0e-10
   Evaluated: Inf < 1.0e-10
```
Any idea whether this is real?

franckgaga:
I had the same results in my side when I tested it

oscardssmith:
are tests clean on main?

franckgaga:
Yes they were on my side, at least.

PR #980, opened 2025-04-11
Bump ImplicitDifferentiation bound + run CompatHelper on subpackages

gdalle:
I just realized your ControlSystemsBase was stuck on an old version of ImplicitDifferentiation (0.4 when we're at 0.7), so I checked and your CompatHelper workflow doesn't see subdirectories. This PR should fix it, and CompatHelper should open new PRs itself soon to upgrade ImplicitDifferentiation

See also:

- https://juliaregistries.github.io/CompatHelper.jl/dev/options/#Packages-in-subdirectories


baggepinnen:
Thanks for the PR! It seems the extension needs an update to be compatible with the latest version of DI. Do you recognize this error https://github.com/JuliaControl/ControlSystems.jl/actions/runs/14397751088/job/40385323705?pr=980#step:5:61 ? It seems some internal function has gotten a different type signature?

gdalle:
Oh, that's probably because the new version of ImplicitDifferentiation only accepts vectors instead of general arrays. It's easier to code that way, but if it causes real problems downstream I may reconsider it

baggepinnen:
What's the intended workflow for handling matrices? `vec -> reshape`?

gdalle:
> What's the intended workflow for handling matrices? vec -> reshape?

Yeah exactly, either I do it myself in ImplicitDifferentiation or every user does it. Inside ID it's easier to work with vectors because we need to solve a linear system, which is more naturally formulated with matrices. But you're not the first user for whom this is a problem, so I'll see what I can do

gdalle:
https://github.com/JuliaDecisionFocusedLearning/ImplicitDifferentiation.jl/pull/167

gdalle:
@baggepinnen can you approve workflows? The last commit picks the version of ImplicitDifferentiation where arrays are supported again

gdalle:
Oh right, I see what's happening here. Your first `ImplicitFunction` did work with a byproduct `z` so I didn't bother to check the rest, but not all of them do actually. In the new version of ImplicitDifferentiation, byproducts are mandatory. Fixing it rn

gdalle:
All good now I think

baggepinnen:
Thanks for all your fixes :)

gdalle:
Turns out the ImplicitDifferentiation ones were not needed after all😅 but it was a good excuse.
As a side note, since you're mostly working with scalars, this code would probably benefit tremendously from StaticArrays

PR #1002, opened 2025-07-01
test: update github actions CI and run tests on `LTS` and `pre`

### LINKED ISSUE #975: Update `test.yml` with new github actions

### LINKED ISSUE #987: Different behavior of MIMO system construction on Julia 1.12
Hello!

I don't know if it's intended (probably not) but on Julia v1.12, this code:

```julia
sys = [tf(1) tf(2); tf(3) tf(4)]
```

produces:

```julia
2×2 Matrix{TransferFunction{Continuous, ControlSystemsBase.SisoRational{Int64}}}:

...
```

whereas on Julia v1.11 and below the result is:

```julia
TransferFunction{Continuous, ControlSystemsBase.SisoRational{Int64}}

...
```

This is also brings up #975, since a CI test run on the new `pre` channel would have probably catch this.

franckgaga:
Trying to:
- closes #975 (it theoretically improves the performance of testing on CI with caching of testing dependencies)
- closes #987 (the part about testing on `pre`: it can be useful for preemptive testing and is less noisy than `nightly`)

Since `pre` is not intended for normal julia users (i.e. non-package developers), I don't think it should impact the CI badge.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
     permissions: # needed for julia-actions/cache to delete old caches
       actions: write
       contents: read
-    continue-on-error: ${{ matrix.version == 'pre' }} # the CI badge will still pass if 'pre' fails
+    continue-on-error: ${{ matrix.version == 'pre' || matrix.version == 'nightly' }} # the CI badge will still pass if 'pre' or 'nightly' fails
     strategy:
       fail-fast: false
       matrix:
@@ -53,7 +53,6 @@ jobs:
       - uses: julia-actions/julia-runtest@v1
         env:
           GROUP: ${{ matrix.group }}
-        # continue-on-error: ${{ matrix.version == 'nightly' }} # Allow nightly to fail and workflow still count as completed
       - uses: julia-actions/julia-processcoverage@v1
         with:
           directories: src,lib/ControlSystemsBase/src,lib/ControlSystemsBase/ext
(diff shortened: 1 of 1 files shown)


franckgaga:
@baggepinnen See above, I think it's a great improvement to be a bit less "extinguishing the fire" in the maintenance of the package XP

It seems the the failling tests on `pre` are legit bugs, do you agree ?

franckgaga:
So I'm pretty sure that it's a bug in Julia 1.12, not in this package.

See https://github.com/JuliaLang/julia/issues/58866

